PR #1, opened 2018-04-20
UpdateDOIs conversion to Java

cookersjs:
- UpdateDOIs converted to Java from Perl

- It is quicker due to it querying specific fields in the 'Pathways' database, rather then iterating over everything

- Some of the 'behind-the-scenes' stuff still needs to be implemented (such as actually recording InstanceEdits), but the basic functionality works. There are lines in the code for testing on your local db, if you have the mysql container running


update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -0,0 +1,40 @@
...
+    MySQLAdaptor gkCentralDBA = null;
+    try {
+      String user = "root";
+      String password = "root";
+      String host = "localhost";
+      String databaseTest = "test_reactome_64";
+      String databaseGk = "gk_central";
+      int port = 3306;
+      // Set up db connections.
+      testReactomeDBA = new MySQLAdaptor(host, databaseTest, user, password, port);
```
SolomonShorser-OICR: All of the values passed into the MySQLAdaptor constructor should come from a properties file. Look to release-qa for some simple examples of how this can be done.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+      NewDOIChecker newDOIChecker = new NewDOIChecker();
+      newDOIChecker.setTestReactomeAdaptor(testReactomeDBA);
+      newDOIChecker.setGkCentralAdaptor(gkCentralDBA);
+  //// Insert your own author id!
+      // long authorId = <YourOwnId>;
+      GKInstance instanceEdit = newDOIChecker.createInstanceEdit(authorId, "org.reactome.release.updateDOIs.UpdateDOIs");
+      newDOIChecker.findNewDOIs(instanceEdit);
+    // Useful to report information back, such as number of changes?
+      System.out.println( "UpdateDOIs Complete" );
```
SolomonShorser-OICR: We should try to avoid using System.out.println, if we can. AddLinks already uses log4j, I can show you how to set this program up to use log4j.

update-dois/src/test/java/org/reactome/release/updateDOIs/AppTest.java
```diff
@@ -0,0 +1,38 @@
+package org.reactome.release.updateDOIs;
+
+import junit.framework.Test;
+import junit.framework.TestCase;
+import junit.framework.TestSuite;
+
+/**
+ * Unit test for simple App.
+ */
+public class AppTest
```
SolomonShorser-OICR: This would be a good time to learn about writing unit tests with JUnit! There are some good, recent examples in the release-qa project. Also, this class should probably be renamed to something like "UpdateDOITest".

update-dois/pom.xml
```diff
@@ -0,0 +1,86 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.reactome.release.updateDOIs</groupId>
+  <artifactId>update-dois</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
```
SolomonShorser-OICR: Leave the version number as "0.0.1" (I don't think we need to bother with SNAPSHOTs right now) until the first release.

update-dois/src/main/java/org/reactome/release/updateDOIs/NewDOIChecker.java
```diff
@@ -0,0 +1,138 @@
...
+          // String updatedDoi = "needs DOI";
+          String dbId = doi.getAttributeValue("DB_ID").toString();
+
+          doi.setAttributeValue("doi", updatedDoi);
+          this.dbaTestReactome.updateInstanceAttribute(doi, "doi");
+
+          // Grabs instance from gk central based on DB_ID taken from test_reactome, and updated doi field
+          gkdois = this.dbaGkCentral.fetchInstanceByAttribute("Pathway", "DB_ID", "=", dbId);
+          if (!gkdois.isEmpty()) {
+            for (GKInstance gkdoi : gkdois) {
```
SolomonShorser-OICR: Maybe there should be some message written to the application log here, stating that an update was performed? That way when it runs we'll know what it did.

update-dois/.classpath
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SolomonShorser-OICR: This file is only valid for Eclipse and if it contains any absolute paths, it will only be useful for *you*. This file should be removed from the repo and then ignored in .gitignore

update-dois/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SolomonShorser-OICR: This file is only valid for Eclipse and if it contains any absolute paths, it will only be useful for *you*. This file should be removed from the repo and then ignored in .gitignore

update-dois/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
SolomonShorser-OICR: This file is only valid for Eclipse and if it contains any absolute paths, it will only be useful for *you*. This file should be removed from the repo and then ignored in .gitignore

update-dois/bin/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SolomonShorser-OICR: This file is only valid for Eclipse and if it contains any absolute paths, it will only be useful for *you*. This file should be removed from the repo and then ignored in .gitignore

update-dois/bin/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.reactome.server</groupId>
+      <artifactId>reactome-core</artifactId>
+      <version>1.1.2</version>
+    </dependency>
+    <dependency>
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <version>1.2.17</version>
```
SolomonShorser-OICR: Log4J 2 has been out for a while, and the current version is 2.11.0. It would be best to use a more current version. Upgrading *probably* won't require and major code changes (maybe some imports will have to be changed), but you *might* need to change the log4j config a little.

update-dois/bin/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>3.8.1</version>
```
SolomonShorser-OICR: Junit's currently at 5.x I think. The 3.x series is old.

update-dois/src/main/java/org/reactome/release/updateDOIs/NewDOIChecker.java
cookersjs: Logging now happens after both databases have been updated

update-dois/.classpath
cookersjs: done

update-dois/.project
cookersjs: done

update-dois/.settings/org.eclipse.jdt.core.prefs
cookersjs: done

update-dois/bin/.project
cookersjs: done

cookersjs:
Testing has been completed in the basic sense, but testing the 'setAttributeValue' and 'updateInstanceAttribute' still needs to be done. As well, I believe there was discussion about capturing log messages, which is something I will need to look into for testing. I am going to create a Project that keeps track of this stuff, in the spirit of a Sprint

SolomonShorser-OICR:
The junit and log4j dependencies are still old, but that can be updated later.

PR #2, opened 2018-05-14
Update dois

cookersjs:
Java Update DOIs now works more similarly to it's Perl equivalent. Two notable changes:
- It queries the DBs directly for DOI's to update, rather than iterating through the entire table
- Instance Edits now work based on Author ID, and will throw errors for any invalid IDs given (Different IDs for different DBs)

### DIFF
--- a/update-dois/src/main/java/org/reactome/release/updateDOIs/findNewDOIsAndUpdate.java
+++ b/update-dois/src/main/java/org/reactome/release/updateDOIs/findNewDOIsAndUpdate.java
@@ -33,7 +33,6 @@ public class findNewDOIsAndUpdate {
     this.dbaGkCentral = adaptor;
   }
 
-  @SuppressWarnings({"unchecked","rawtypes"})
   public void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pathToReport) {
 
     Collection<GKInstance> dois;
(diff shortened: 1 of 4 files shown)


update-dois/src/main/java/org/reactome/release/updateDOIs/findNewDOIsAndUpdate.java
```diff
@@ -29,42 +33,89 @@ public void setGkCentralAdaptor(MySQLAdaptor adaptor) {
...
+    HashMap reportContents = this.getReportContents(pathToReport);
+    int reportHits = 0;
+    int fetchHits = 0;
+    ArrayList updatedDOIs = new ArrayList();
     try {
-     dois = this.dbaTestReactome.fetchInstanceByAttribute("Pathway", "doi", "NOT REGEXP", "^10.3180");
-//     dois = this.dbaTestReactome.fetchInstanceByAttribute("Pathway", "DB_ID", "REGEXP", "8939211|9006115|9018678|9033241");
+
+      dois = this.dbaTestReactome.fetchInstanceByAttribute("Pathway", "doi", "NOT REGEXP", "^10.3180");
+      dois = this.dbaTestReactome.fetchInstanceByAttribute("Pathway", "DB_ID", "REGEXP", "8939211|9006115|9018678|9033241");
```
SolomonShorser-OICR: `dois` gets a value assigned twice. I think the second assignment will overwrite the results of the first one, so why is the first one done?

cookersjs: Second one should have been commented out or removed, fixed

update-dois/src/main/java/org/reactome/release/updateDOIs/findNewDOIsAndUpdate.java
```diff
@@ -73,20 +124,44 @@ public void findNewDOIsAndUpdate(long authorId) {
     }
   }
 
+  // Parses input report and places each line's contents in HashMap
+  @SuppressWarnings("rawtypes")
+  public HashMap getReportContents(String pathToReport) {
+
+	HashMap reportContents = new HashMap();
```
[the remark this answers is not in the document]
cookersjs: Fixed

PR #3, opened 2018-05-15
Feature/ch ebi update

SolomonShorser-OICR:


chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -0,0 +1,487 @@
...
+		GKInstance instanceEdit = null;
+		if (!testMode)
+		{
+			instanceEdit = createInstanceEdit(this.adaptor, this.personID, this.getClass().getCanonicalName());
+		}
+		
+		for (Long moleculeDBID : entityMap.keySet())
+		{
+			// One transaction per molecule - is this too many? If this runs too slow, maybe switch to one transaction per
+			// program execution.
```
cookersjs: Guanming did recommend trying to minimize DB interactions, is there any specific benefit to having one transaction per molecule vs doing a bulk update?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -0,0 +1,487 @@
...
+			if (!testMode && adaptor.supportsTransactions())
+			{
+				adaptor.startTransaction();
+			}
+			
+			GKInstance molecule = adaptor.fetchInstance(moleculeDBID);
+			Entity entity = entityMap.get(moleculeDBID);
+
+			// Not sure why the old Perl code stripped
+			// out "CHEBI:", but I'll do it here for consistency.
```
cookersjs: Related to how it's stored in the DB, as an integer?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -0,0 +1,487 @@
...
+							nameFound = true;
+							// Remove the ChEBI name at position i
+							names.remove(i);
+							// Add the ChEBI name back at position 0.
+							names.add(0, chebiName);
+						}
+						i++;
+					}
+					// If we went through the whole array and didn't find the ChEBI name,
+					// we must add it, at the begining of the list of names.
```
cookersjs: Is this block new?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -0,0 +1,487 @@
...
+					if (!nameFound)
+					{
+						names.add(0, chebiName);
+					}
```
cookersjs: is this block new?

cookersjs:
To the best of my understanding, it seems to match the Perl implementation. Has it been checked for comparable updates?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: Transactions are supposed to be atomic - if it succeeds, the whole thing succeeds. If it fails, the whole thing fails. In this case, I considered that if one transaction fails (for some reason), I didn't want ALL of the data update transactions to fail. 
Also, this program only takes a couple of minutes to run so I don't think the performance overhead is too bad. I could change it to one large transaction. If that fails, all data updates would need to be redone. In a program of this scale, it probably doesn't matter which way it's done. I could do timing tests later (if i have time) to see if multiple transactions is significantly slower or not.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: This is to check that if the name from the ChEBI service hasn't been found in the list of existing names by the time all existing names have been checked, then add the name to the beginning of the list.

Similar functionality exists in the Perl code but I think this code is easier to understand (even if it has more lines)

SolomonShorser-OICR:
Checking for comparable results is a work-in-progress ;)

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
cookersjs: Can't argue with that logic -- better to fail a tiny bit for a good reason then to fail everything due to a tiny reason. Also, the module is down to minutes? Now thats performance improvement!

SolomonShorser-OICR: On my workstation, a few minutes at most.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -0,0 +1,499 @@
...
+			if (!testMode)
+			{
+				addInstanceEditToExistingModifieds(instanceEdit, refEntity);
+				
+				adaptor.updateInstanceAttribute(refEntity, ReactomeJavaConstants.name);
+			}
+		}
+	}
+
+	private void addInstanceEditToExistingModifieds(GKInstance instanceEdit, GKInstance instance) throws InvalidAttributeException, Exception, InvalidAttributeValueException
```
SolomonShorser-OICR: @cookersjs You might want to check your code to ensure you're adding a new InstanceEdit to the existing list, rather than overwriting the existing list of "modified" InstanceEdits. I'd forgotten this "gotcha!" until a few minutes ago.

SolomonShorser-OICR: I'd forgotten this potential problem, but I'd already dealt with in AddLinks: https://github.com/reactome/add-links/blob/de443d43852a9307c387daa9d3d1abbd5c36c4b1/src/main/java/org/reactome/addlinks/db/ReferenceCreator.java#L289-L292

cookersjs: Thanks for the heads up! I copied the AddLinks Instance Edits, and it correctly pushed to the array

SolomonShorser-OICR:
I ran this and the Perl version on copies of the same database, and 75 showed mismatching formulas. When I check CHEBI's website, it looks like the Perl code is not updating some (or all) of the formulas correctly.

cookersjs:
Meaning that the Perl code had been working incorrectly for likely quite a while?

SolomonShorser-OICR:
Possibly.

PR #4, opened 2018-06-04
Transactional support added for Gk Central

cookersjs:
Also updated formatting of code

Point of interest: lines 124-126 in findNewDOIsAndUpdate.java. I have a rollback there, depending on if transactional support exists. Is it necessary?


SolomonShorser-OICR:
Have you tested that the logging now goes to its own file instead of the file that the Perl script captures? Also, upgrade to log4j2 as I don't think log4j 1.x is still supported.

cookersjs:
I didn't test it, I'll do that and do the upgrade and then push those changes for review

SolomonShorser-OICR:
Ok :+1:

cookersjs:
@SolomonShorser-OICR if you could look at the new runtime tests I've added for UpdateDOIs and see if anything jumps out that would be great. You should be able to look at just the 'ReportTests.java' file for this. Commenting describes the point of each test

update-dois/src/main/java/org/reactome/release/updateDOIs/ReportTests.java
```diff
@@ -0,0 +1,92 @@
...
+				int resolved = 0;
+				for (String key : expectedUpdatedDOIs.keySet()) 
+				{
+					if (expectedUpdatedDOIs.get(key).get("stableId").equals(missedStableId)) 
+					{
+						if (!expectedUpdatedDOIs.get(key).get("stableIdVersion").equals(missedStableIdVersion)) 
+						{
+							logger.warn("[" + key + "] StableID 'version' in DB different from expected: [DB] " + missedDoi + "* ~ [Expected] " + key + "*");
+							resolved++;
+							continue;
```
SolomonShorser-OICR: Is "continue" necessary here?

SolomonShorser-OICR: memo: add comment explaining the "continue"

cookersjs:
@SolomonShorser-OICR I've finished all updates I planned for release 66. Basic functionality remains the same, but I've added those new tests (which you've reviewed) and re-structured a bit so that the logical flow of things is cleaner.

PR #6, opened 2018-07-13
Feature/go update

SolomonShorser-OICR:
GO Update functionality. The original Perl GO Update code is here:  https://github.com/reactome/Release/tree/85545efd58e42b6f8c6cf326fe9933491d2edf37/scripts/release/go_update and here: https://github.com/reactome/Release/blob/85545efd58e42b6f8c6cf326fe9933491d2edf37/modules/GKB/Release/Steps/GoUpdate.pm


chebi-update/src/main/java/org/reactome/release/chebiupdate/Main.java
```diff
@@ -27,36 +21,8 @@ public static void main(String[] args) throws SQLException, Exception
...
-		String dbHost = props.getProperty("db.host", "localhost");
-		String dbUser = props.getProperty("db.user");
-		String dbPassword = props.getProperty("db.password");
-		String dbName = props.getProperty("db.name");
-		int dbPort = new Integer(props.getProperty("db.port", "3306"));
-		
-		MySQLAdaptor adaptor = new MySQLAdaptor(dbHost, dbName, dbUser, dbPassword, dbPort);
-		return adaptor;
+		ChebiUpdateStep chebiUpdateStep = new ChebiUpdateStep();
+		chebiUpdateStep.executeStep(props);
```
cookersjs: For my own curiousity, why is it better to name a variable for the class rather than just call the classes function directly: example: `ChebiUpdateStep.executeStep(props);`

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * This class reports on duplicate GO Accessions.
+ * @author sshorser
+ *
+ */
+public class DuplicateReporter
+{
+	private static String duplicateAccessionQuery = "select accession, count(db_id) as instance_count\n" + 
+			"from (\n" +
```
cookersjs: This 'from' doesn't connect to anything I can see?

go-update/src/main/java/org/reactome/release/goupdate/GoUpdateStep.java
```diff
@@ -0,0 +1,130 @@
...
+			long personID = new Long(props.getProperty("person.id"));
+			
+			String pathToGOFile = props.getProperty("pathToGOFile","src/main/resources/gene_ontology_ext.obo");
+			String pathToEC2GOFile = props.getProperty("pathToEC2GOFile","src/main/resources/ec2go");
+			
+			// Load the files.
+			List<String> goLines = Files.readAllLines(Paths.get(pathToGOFile));
+			List<String> ec2GoLines = Files.readAllLines(Paths.get(pathToEC2GOFile));
+
+			reportOnDuplicateAccessions(adaptor);
```
cookersjs: Since identifying duplicates is a new function for GO update, maybe a small explanation of how it accomplishes this would be useful?

chebi-update/src/main/java/org/reactome/release/chebiupdate/Main.java
SolomonShorser-OICR: `executeStep` is not a static method. An instance of `ReleaseStep` is required.
As for the name: yes, I could have just named that variable *step*, or *chebiStep*... but I went with *chebiUpdateStep*, I guess to make it very clear what kind of step it is.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
SolomonShorser-OICR: It is selecting from the subquery (which is a union of three different tables), which is between parentheses.

go-update/src/main/java/org/reactome/release/goupdate/GoUpdateStep.java
SolomonShorser-OICR: I added a note in the docs for the DuplicateReporter class.

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
```diff
@@ -0,0 +1,232 @@
...
+							{
+								goTerms.put(goID, new HashMap<String,Object>());
+								currentGOID = goID;
+							}
+							else
+							{
+								// If a GO ID appears a second time, it will not be added to the hash, and the user will 
+								// get a message asking them to verify if the file is really OK.
+								// Maybe throw a RuntimeException for this? It really should never happen.
+								logger.error("GO ID {} has appeared more than once in the input! This is highly unexpected. "
```
cookersjs: Is it rare enough and bad enough that it warrants the runtime exception? Or could it be flagged and filtered out

SolomonShorser-OICR: It never actually happened in my testing with real data. I don't think the old code handled it all. Currently, the code will just log an error message. That code was just a thought about throwing a `RuntimeException` if it happens. But maybe I should actually add that in.
The reason? If a GO Id actually did appear twice in the file, and the content for each occurrence was different, how would you decide which one to keep and which to discard?

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -0,0 +1,545 @@
...
+						{
+							mismatchCount++;
+							categoryMismatchStringBuilder.append("Category mismatch! GO ID: ").append(goID).append(" Category in DB: ").append(goInst.getSchemClass().getName()).append(" category in GO file: ").append(currentCategory).append("\n");
+							instancesForDeletion.add(goInst);
+						}
+					}
+				}
+			}
+			else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.PENDING_OBSOLETION) && goTermsFromFile.get(goID).get(GoUpdateConstants.PENDING_OBSOLETION).equals(true))
+			{
```
cookersjs: Regarding the obsoletion of terms, is there anyway to update old ones or is this part of something like a 'total restart' to gene ontology, meaning there isn't a 1:1 way of connecting obsolete terms to new ones?

SolomonShorser-OICR: In the old Perl code, obsolete GO terms were logged and deleted. Anything that referred to those terms was modified to refer to the GO terms that were suggested for the obsolete term, in fields such as "alternate", "replaced_by", or "consider". A couple of weeks ago, Peter said that we should not be removing them, but giving the list of obsolete terms to curators so the curators can make a decision on how they are replaced. It sounds like this process of replacing obsolete GO terms is not always simple or straightforward and that doing it automatically is maybe not the best idea.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
cookersjs: Ah of course, BEDMAS

SolomonShorser-OICR: :+1:  :)

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
cookersjs: Yea I suppose in that case you would want to make a note of it and potentially remove both. I guess we'll cross that bridge when it appears?

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
cookersjs: Makes sense, good to know

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
SolomonShorser-OICR: Right now, it would actually keep the first one it finds and all others would trigger an error message. That seems arbitrary, so I think I *will* add that RuntimeException.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -0,0 +1,366 @@
...
+	
+	/**
+	 * Updates the relationships between GO terms in the database.
+	 * @param allGoInstances - Map of all GO instances in the database.
+	 * @param goProps - The properties to update with.
+	 * @param relationshipKey - The key to use to look up the values  in goProps.
+	 * @param reactomeRelationshipName - The name of the relationship, can be one of "is_a", "has_part", "part_of", "component_of", "regulates", "positively_regulates", "negatively_regulates".
+	 */
+	public void updateRelationship(Map<String, List<GKInstance>> allGoInstances, Map<String, Object> goProps, String relationshipKey, String reactomeRelationshipName)
+	{
```
cookersjs: This function is needed because of the changed status of GO terms, or because of how things in the data model have changed (such as regulations), or both, or neither?

SolomonShorser-OICR: Based on what I read in the Perl code, the relationship fields for a GO term may change, so the GO terms in the database need to be updated to keep in sync with the changes from GO.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -0,0 +1,188 @@
...
+
+import org.gk.model.GKInstance;
+import org.gk.model.InstanceUtilities;
+import org.gk.model.ReactomeJavaConstants;
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.schema.SchemaAttribute;
+import org.reactome.util.compare.DBObjectComparer;
+
+public class JavaVsPerlComparison {
+
```
cookersjs: Will this test be needed moving forward? I assume after the transitionary release where we first go from Perl to Java that this test will be 'pending obsoletion'

SolomonShorser-OICR: Yeah, after a few releases that use this code, after it's trusted, then this test code won't really be needed anymore.

cookersjs:
All done -- Considering my questions/comments were mostly about the goings-ons of the script, I think that means things make sense. I only skimmed the testing portions, but everything else seems good!

PR #10, opened 2018-10-18
Feature/ch ebi update

SolomonShorser-OICR:
@jweiser @cookersjs Looking for general feedback and comments on this.

This is a Re-implementation of the ChEBI Update code and algorithm. It is *not* the same as the Perl algorithm.

### DIFF
--- a/chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
+++ b/chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
@@ -41,6 +41,7 @@ public class ChebiUpdater
 	private static final Logger refMolIdentChangeLog = LogManager.getLogger("molIdentChangeLog");
 	private static final Logger refEntChangeLog = LogManager.getLogger("refEntChangeLog");
 	private static final Logger duplicatesLog = LogManager.getLogger("duplicatesLog");
+	private static final Logger failedChebiLookupsLog = LogManager.getLogger("failedChebiLookupsLog");
 	private boolean testMode = true;
 	private MySQLAdaptor adaptor;
 	private ChebiWebServiceClient chebiClient = new ChebiWebServiceClient();
@@ -49,6 +50,7 @@ public class ChebiUpdater
 	private StringBuilder duplicatesSB = new StringBuilder();
 	private Map<GKInstance, List<String>> referenceEntityChanges = new HashMap<GKInstance, List<String>>();
 	private long personID;
+	private boolean useCache;
 	// A Comparator object that will compare GKInstances, assuming that they are of the "Person" type, with a surname and firstname.
 	private Comparator<GKInstance> personComparator = new Comparator<GKInstance>()
 		{
@@ -73,11 +75,12 @@ public class ChebiUpdater
 			}
 		};
 	
-	public ChebiUpdater(MySQLAdaptor adaptor, boolean testMode, long personID)
+	public ChebiUpdater(MySQLAdaptor adaptor, boolean testMode, long personID, boolean useCache)
 	{
 		this.adaptor = adaptor;
 		this.testMode = testMode;
 		this.personID = personID;
+		this.useCache = useCache;
 	}
 
 	/**
@@ -106,7 +109,7 @@ public class ChebiUpdater
 		// uk.ac.ebi.chebi.webapps.chebiWS.model.Entity from ChEBI.
 		Map<Long, Entity> entityMap = Collections.synchronizedMap(new HashMap<Long, Entity>());
 		// A list of the ReferenceMolecules where we could nto get info from ChEBI.
-		List<GKInstance> failedEntitiesList = Collections.synchronizedList(new ArrayList<GKInstance>());
+		Map<GKInstance, String> failedEntitiesList = Collections.synchronizedMap(new HashMap<GKInstance, String>());
 
 		logger.info("{} ChEBI ReferenceMolecules to check...", refMolecules.size());
 
@@ -115,9 +118,10 @@ public class ChebiUpdater
 		logger.info("Number of entities we were able to retrieve information about: {}", entityMap.size());
 		logger.info("Number of entities we were NOT able to retrieve information about: {}", failedEntitiesList.size());
 
-		for (GKInstance molecule : failedEntitiesList)
+		failedChebiLookupsLog.info("# ReferenceMolecule\tReason");
+		for (GKInstance molecule : failedEntitiesList.keySet())
 		{
-			logger.info("Could not get info from ChEBI for: {}", molecule.toString());
+			failedChebiLookupsLog.info("{}\t{}", molecule.toString(), failedEntitiesList.get(molecule));
 		}
 		
 		// print headers for log files
@@ -455,22 +459,29 @@ public class ChebiUpdater
 	 * @param failedEntitiesList - A list of ReferenceMolecules for which no information was returned by ChEBI. Will be updated by this method.
 	 * @throws IOException 
 	 */
-	private void retrieveUpdatesFromChebi(Collection<GKInstance> refMolecules, Map<Long, Entity> entityMap, List<GKInstance> failedEntitiesList) throws IOException
+	private void retrieveUpdatesFromChebi(Collection<GKInstance> refMolecules, Map<Long, Entity> entityMap, Map<GKInstance, String> failedEntitiesList) throws IOException
 	{
-		// TODO: Add a config flag that can be used to ignore the cached file (for when you want only the *freshest* results)
 		Map<String,List<String>> chebiCache = Collections.synchronizedMap(new HashMap<String, List<String>>());
-		// if the cache exists, load it.
-		if (Files.exists(Paths.get("chebi-cache")))
+		if (this.useCache)
 		{
-			Files.readAllLines(Paths.get("chebi-cache")).parallelStream().forEach( line -> {
-				String[] parts = line.split("\t");
-				String chebiID = parts[0];
-				String name = parts[1];
-				String formula = parts.length > 2 ? parts[2] : "";
-				chebiCache.put(chebiID, Arrays.asList(name, formula));
-			});
+			logger.info("useCache is TRUE - chebi-cache file will be read, and populated. Identifiers not in the cache will be queried from ChEBI.");
+			// if the cache exists, load it.
+			if (Files.exists(Paths.get("chebi-cache")))
+			{
+				Files.readAllLines(Paths.get("chebi-cache")).parallelStream().forEach( line -> {
+					String[] parts = line.split("\t");
+					String chebiID = parts[0];
+					String name = parts[1];
+					String formula = parts.length > 2 ? parts[2] : "";
+					chebiCache.put(chebiID, Arrays.asList(name, formula));
+				});
+			}
+			logger.debug("{} entries in the chebi-cache", chebiCache.size());
 		}
-		logger.debug("{} entries in the chebi-cache", chebiCache.size());
+		else
+		{
+			logger.info("useCache is FALSE - chebi-cache will NOT be read. ChEBI will be queried for ALL identifiers.");
+		}
 		FileWriter fileWriter = new FileWriter("chebi-cache", true);
 		// BufferedWriter is supposed to be thread-safe.
 		BufferedWriter bw = new BufferedWriter(fileWriter);
@@ -484,19 +495,26 @@ public class ChebiUpdater
 				identifier = (String) molecule.getAttributeValue("identifier");
 				if (identifier != null && !identifier.trim().equals(""))
 				{
-					// only query web service if the data is not in the chebi-cache
+					// only query web service if the data is not in the chebi-cache - NOTE: chebiCache will always be empty if this.useCache == false
 					if (!chebiCache.containsKey("CHEBI:"+identifier))
 					{
+						if (this.useCache && chebiCache.size() > 0)
+						{
+							logger.trace("Cache miss for CHEBI:{}", identifier);
+						}
 						Entity entity = this.chebiClient.getCompleteEntity(identifier);
 						if (entity != null)
 						{
 							entityMap.put(molecule.getDBID(), entity);
-							bw.write(entity.getChebiId()+"\t"+entity.getChebiAsciiName()+"\t"+ (entity.getFormulae().size() > 0 ? entity.getFormulae().get(0).getData() : "") + "\n");
-							bw.flush();
+							if (this.useCache)
+							{
+								bw.write(entity.getChebiId()+"\t"+entity.getChebiAsciiName()+"\t"+ (entity.getFormulae().size() > 0 ? entity.getFormulae().get(0).getData() : "") + "\n");
+								bw.flush();
+							}
 						}
 						else
 						{
-							failedEntitiesList.add(molecule);
+							failedEntitiesList.put(molecule, "ChEBI WebService response was NULL.");
 						}
 					}
 					else
@@ -513,6 +531,7 @@ public class ChebiUpdater
 				else
 				{
 					logger.error("ERROR: Instance \"{}\" has an empty/null identifier. This should not be allowed.", molecule.toString());
+					failedEntitiesList.put(molecule, molecule.toString() + " has an empty/NULL identifier.");
 				}
 				int i = counter.getAndIncrement();
 				if (i % 250 == 0)
@@ -526,11 +545,13 @@ public class ChebiUpdater
 				if (e.getMessage().contains("invalid ChEBI identifier"))
 				{
 					logger.error("ERROR: ChEBI Identifier \"{}\" is not formatted correctly.", identifier);
+					failedEntitiesList.put(molecule, "ChEBI Identifier \""+identifier+"\" is not formatted correctly.");
 				}
 				// Log this identifier, but don't fail.
 				else if (e.getMessage().contains("the entity in question is deleted, obsolete, or not yet released"))
 				{
 					logger.error("ERROR: ChEBI Identifier \"{}\" is deleted, obsolete, or not yet released.", identifier);
+					failedEntitiesList.put(molecule, "ChEBI Identifier \""+identifier+"\" is deleted, obsolete, or not yet released.");
 				}
 				else
 				{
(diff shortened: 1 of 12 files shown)


cookersjs:
I'll review this first thing once downloadDirectory work is wrapping up

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -1,53 +1,87 @@
...
+		{
+			@Override
+			public int compare(GKInstance o1, GKInstance o2)
+			{
+				try
+				{
+					int surnameCompare = ((String)o1.getAttributeValue(ReactomeJavaConstants.surname)).compareTo((String)o2.getAttributeValue(ReactomeJavaConstants.surname));
+					if (surnameCompare == 0)
+					{
+						return ((String)o1.getAttributeValue(ReactomeJavaConstants.firstname)).compareTo((String)o2.getAttributeValue(ReactomeJavaConstants.firstname));
```
jweiser: Extracting to a private method shared for lines 63 and 66 might make this clearer

SolomonShorser-OICR: Yeah, there are a few ways this could be made more readable, I guess. I'll look into it.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -1,53 +1,93 @@
...
 	private StringBuilder formulaUpdateSB = new StringBuilder();
 	private StringBuilder formulaFillSB = new StringBuilder();
-	private StringBuilder nameSB = new StringBuilder();
 	private StringBuilder duplicatesSB = new StringBuilder();
-
+	private Map<GKInstance, List<String>> referenceEntityChanges = new HashMap<GKInstance, List<String>>();
 	private long personID;
+	private boolean useCache;
+	// A Comparator object that will compare GKInstances, assuming that they are of the "Person" type, with a surname and firstname.
+	private Comparator<GKInstance> personComparator = new Comparator<GKInstance>()
```
jweiser: Maybe move the initialization of the personComparator to its own method so the constructor is immediately after the field declarations

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +150,107 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
 		}
+		if (!testMode)
+		{
+			adaptor.commit();
+		}
+		else
+		{
+			adaptor.rollback();
+		}
+
```
jweiser: Might be more succinct as ternary - !testmode ? adaptor.commit() : adaptor.rollback();

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +150,107 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
+		
+		refEntChangeLog.info("# DB_ID\tCreator\tAffected ReferenceEntity\tNew ChEBI Name\tUpdated list of all names");
+		// Print the referenceEntities that have changes, sorted by who created them.
+		for (GKInstance creator : this.referenceEntityChanges.keySet().stream().sorted(this.personComparator).collect(Collectors.toList()))
+		{
+			for (String message : this.referenceEntityChanges.get(creator))
+			{
+				refEntChangeLog.info("{}\t{}",creator.toString(), message);
+			}
+		}
```
jweiser: Can extract this as a private method called something like 'reportReferenceEntityChanges' similar to how there is 'reportIfMoleculeIdentifierChanged'

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: That would be more succinct, but I prefer the clarity of what is written. I expect a ternary operator to be used in an expression that returns, such as:

```
int x = someNumber > 0 ? processPositive(someNumber) : processNegative(someNumber);
```

A ternary that does not produce a value is odd-looking to me. I don't think this is commonly done in Java, and I don't recall seeing this practice.

SolomonShorser-OICR: UPDATE: I just tried this and Java does not seem to allow this:
```
0 < 1 ? System.exit(0) : System.exit(1);
```
Error: "The left-hand side of an assignment must be a variable"

SolomonShorser-OICR: ...I guess that's why I've never seen it.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
jweiser: That's true.  Personally, I just find it easier to read methods doing multiple things when the related sections of behaviour are encapsulated

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +152,112 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
 			String chebiName = entity.getChebiAsciiName();
 			List<DataItem> chebiFormulae = entity.getFormulae();
 
-			updateReferenceEntities(molecule, chebiName, instanceEdit);
 
-			// Now, check to see if we need to update the ReferenceMolecule itself.
 			String moleculeIdentifier = (String) molecule.getAttributeValue(ReactomeJavaConstants.identifier);
 			String moleculeName = (String) molecule.getAttributeValuesList(ReactomeJavaConstants.name).get(0);
 			String moleculeFormulae = (String) molecule.getAttributeValue(ReactomeJavaConstants.formula);
```
jweiser: Could these three Strings be moved into the private methods that use them?  That way you could also get rid of them as parameters and use the 'molecule' parameter to get their values inside the methods.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +152,112 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
 
 			String prefix = "ReferenceMolecule (DB ID: " + molecule.getDBID() + " / ChEBI ID: " + moleculeIdentifier + ") has changes: ";
```
jweiser: Prefix could also become a simple private method using molecule as a parameter.  Then prefix could be removed as a parameter from other methods and just called for inside them.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +152,112 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
 	 * @throws InvalidAttributeException
 	 * @throws InvalidAttributeValueException
 	 * @throws Exception
 	 */
-	private void updateMoleculeFormula(GKInstance molecule, List<DataItem> chebiFormulae, String moleculeFormulae, String prefix) throws InvalidAttributeException, InvalidAttributeValueException, Exception
+	private boolean updateMoleculeFormula(GKInstance molecule, List<DataItem> chebiFormulae, String moleculeFormulae, String prefix) throws InvalidAttributeException, InvalidAttributeValueException, Exception
 	{
+		boolean updated = false;
 		if (!chebiFormulae.isEmpty())
 		{
```
jweiser: You could invert this to use a guard clause so the method logic doesn't need to be in an if statement:

if (chebiFormula.isEmpty()) {
    return false;
}
...rest of the method

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +152,112 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
 			String firstFormula = chebiFormulae.get(0).getData();
-			if (firstFormula != null)
+			if (firstFormula != null && !firstFormula.trim().equals(""))
 			{
 				if (moleculeFormulae == null)
 				{
 					this.formulaFillSB.append(prefix).append("New Formula: ").append(firstFormula).append("\n");
+					updated = true;
```
jweiser: moving each .append to a new line might make this more readable

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -115,76 +152,112 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
 				}
 				else if (!firstFormula.equals(moleculeFormulae))
 				{
 					this.formulaUpdateSB.append(prefix).append(" Old Formula: ").append(moleculeFormulae).append(" ; ").append("New Formula: ").append(firstFormula).append("\n");
```
jweiser: moving .append calls would be helpful here too

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: I find that programming style confusing as the exit point is not obvious. I admit I do that sometimes but I try to avoid it when possible.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -114,196 +152,263 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
+				//}
+			}
+			else
+			{
+				// Only print a warning if we're going from non-NULL formula to NULL formula.
+				if (moleculeFormulae != null && !moleculeFormulae.trim().equals(""))
 				{
-					adaptor.updateInstanceAttribute(molecule, ReactomeJavaConstants.formula);
+					logger.warn("Got empty/NULL formula for {}, old formula was: {}", molecule.toString(), moleculeFormulae);
 				}
```
jweiser: I can't remember -- did Peter confirm that it was okay to set the formula to null when we have a value in our database and ChEBI gives us an empty value?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -114,196 +152,263 @@ public void updateChebiReferenceMolecules() throws SQLException, Exception
...
+			List<GKInstance> oldIdentifierReferrers = (List<GKInstance>) molecule.getReferers(ReactomeJavaConstants.referenceEntity);
+			if (oldIdentifierReferrers != null && !oldIdentifierReferrers.isEmpty())
 			{
-				adaptor.updateInstanceAttribute(molecule, ReactomeJavaConstants.identifier);
+				for (GKInstance referrer : oldIdentifierReferrers)
+				{
+					oldIdentifierReferrersString += referrer.getDBID().toString() + "|";
+				}
+			}
+
```
jweiser: You could do the generation of the oldIdentifierReferrersString like this:
			// Need to get list of DB_IDs of referrers for *old* Identifier and also for *new* Identifier.
			String oldIdentifierReferrersString = 
				((List<GKInstance>) molecule.getReferers(ReactomeJavaConstants.referenceEntity))
				.stream()
				.map(referrer -> referrer.getDBID().toString())
				.collect(Collectors.joining("|"));

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -383,9 +496,36 @@ public void checkForDuplicates() throws SQLException, Exception
...
+					chebiCache.put(oldChebiID, Arrays.asList(newChebiID, name, formula));
+				});
+			}
+			logger.debug("{} entries in the chebi-cache", chebiCache.size());
+		}
+		else
+		{
+			logger.info("useCache is FALSE - chebi-cache will NOT be read. ChEBI will be queried for ALL identifiers.");
+		}
+		FileWriter fileWriter = new FileWriter("chebi-cache", true);
```
jweiser: This could be extracted to a method to populate the chebiCache

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -383,9 +496,36 @@ public void checkForDuplicates() throws SQLException, Exception
...
+					String formula = parts.length > 2 ? parts[3] : "";
+					chebiCache.put(oldChebiID, Arrays.asList(newChebiID, name, formula));
+				});
+			}
+			logger.debug("{} entries in the chebi-cache", chebiCache.size());
+		}
+		else
+		{
+			logger.info("useCache is FALSE - chebi-cache will NOT be read. ChEBI will be queried for ALL identifiers.");
+		}
```
jweiser: This could be extracted to a method to populate the chebiCache

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -395,39 +535,75 @@ private void retrieveUpdatesFromChebi(Collection<GKInstance> refMolecules, Map<L
...
+				// "invalid ChEBI identifier" probably shouldn't break execution but should be logged for further investigation.
+				if (e.getMessage().contains("invalid ChEBI identifier"))
 				{
-					logger.error("WebService error: {} ; Identifier that triggered it: {}; Molecule DB ID: {}", e.getMessage(), identifier, molecule.getDBID());
+					logger.error("ERROR: ChEBI Identifier \"{}\" is not formatted correctly.", identifier);
+					failedEntitiesList.put(molecule, "ChEBI Identifier \""+identifier+"\" is not formatted correctly.");
 				}
-				catch (Exception e1)
+				// Log this identifier, but don't fail.
+				else if (e.getMessage().contains("the entity in question is deleted, obsolete, or not yet released"))
```
jweiser: Maybe using .matches with keywords and "OR" in a regex would be better in case ChEBI changes their error message slightly

SolomonShorser-OICR: Do you mean something like
```
...matches(".*the |entity |in |question |is |deleted....*");
```
Anticipating possible future changes to messages seems iffy. If ChEBI sends a `ChebiWebServiceFault_Exception` that has a different message, the application will still log it.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: Peter had no comment. Lisa said:
> I think the formula additions/changes don't need to be reported.

And Bijay said:

> Information about formulae I don't think we use anywhere so those first 3 files can be done silently without curator involvement.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: It could be, but I am not sure much value would be gained from that in this case.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
jweiser: Okay, sounds good!

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
jweiser: It would just shorten the method a bit.  I found it a bit difficult to keep in my head what was going on for over 100 lines.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
jweiser: Could it look for just for 'deleted', 'obsolete', or 'not yet released'?  My concern was that if the error message doesn't match exactly, the else clause runs and a RuntimException is triggered.

SolomonShorser-OICR: There could be other error messages that contain those words but don't mean the same thing as the message I'm checking for now. In this case I'd rather let the program fail on a completely new error message, than let it continue running when it finds a completely new message that might need to be handled in a new way.

chebi-update/src/test/java/org/reactome/release/test/chebiupdate/ChebiUpdaterTest.java
```diff
@@ -70,92 +98,215 @@
...
+			when(molecule4.getSchemClass()).thenReturn(refMolSchemaClass);
+			when(molecule4.getDBID()).thenReturn(444L);
+			when(adaptor.fetchInstance(444L)).thenReturn(molecule4);
+			
+			when(refEnt1.getAttributeValuesList("name")).thenReturn(new ArrayList<String>(Arrays.asList("OTHERNAME", "NAME1", "BLAH")));
+			when(refEnt1.getSchemClass()).thenReturn(simpleEntitySchemaClass);
+			when(refEnt2.getAttributeValuesList("name")).thenReturn(new ArrayList<String>(Arrays.asList("TEST", "NAME2")));
+			when(refEnt2.getSchemClass()).thenReturn(simpleEntitySchemaClass);
+			when(refEnt3.getAttributeValuesList("name")).thenReturn(new ArrayList<String>(Arrays.asList("NAME-2", "ASDFQ@#$FASFASDF","NOPE")));
+			when(refEnt1.getSchemClass()).thenReturn(chemicalDrugSchemaClass);
```
jweiser: Should this be refEnt3 returning chemicalDrugSchemaClass?

chebi-update/src/test/java/org/reactome/release/test/chebiupdate/ChebiUpdaterTest.java
```diff
@@ -70,92 +98,215 @@
...
-		
 		when(chebiEntity1.getChebiAsciiName()).thenReturn("NAME1");
-		when(chebiEntity2.getChebiAsciiName()).thenReturn("NAME-2");
+		// This mock ChEBI entity will return a different ChEBI ID! It's ID is now the same as molecule2.
+		when(chebiEntity1.getChebiId()).thenReturn("CHEBI:332211");
 		
-		when(chebiEntity1.getChebiId()).thenReturn("CHEBI:112233");
-		when(chebiEntity2.getChebiId()).thenReturn("CHEBI:332211-2");
+		when(chebiEntity2.getChebiAsciiName()).thenReturn("NAME-2");
+		when(chebiEntity2.getChebiId()).thenReturn("CHEBI:332211");
```
jweiser: Just wanted to double check -- should chebiEntity2 and chebiEntity1 above return the same values for a call to getChebiId()?

SolomonShorser-OICR: That diff looks... weird. It's comparing line 92 to line 235. The method in which line 235 currently exists was not even in that file prior to this PR, so I think github got confused.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
jweiser: Okay, that makes sense

chebi-update/src/test/java/org/reactome/release/test/chebiupdate/ChebiUpdaterTest.java
jweiser: Okay, looks like there's no problem then after going over it in person!

PR #12, opened 2018-11-07
Feature/orthoinference

cookersjs:
OrthoInference, in all its glory. Comments on sustainable coding practices, reactome java api usage/improvement suggestions and readability of code are what this needs the most. Obviously those are just requests, so if theres anything at all that can be improved I'd like to hear it. Thanks in advance


orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+	private static GKInstance instanceEdit;
+	private static GKInstance speciesInst;
+	private static HashMap<GKInstance,GKInstance> manualEventToNonHumanSource = new HashMap<GKInstance,GKInstance>();
+	private static ArrayList<GKInstance> manualHumanEvents = new ArrayList<GKInstance>();
+
+	@SuppressWarnings("unchecked")
+	public static void eventInferrer(Properties props, String pathToConfig, String pathToSpeciesConfig) throws Exception
+	{
+		// Set up DB adaptor using config.properties file
+		Object speciesToInferFromLong = "Homo sapiens";
```
jweiser: Could this be a String?

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+		SkipTests.setAdaptor(dbAdaptor);
+		GenerateInstance.setAdaptor(dbAdaptor);
+		OrthologousEntity.setAdaptor(dbAdaptor);
+		InferEWAS.setAdaptor(dbAdaptor);
+		UpdateHumanEvents.setAdaptor(dbAdaptor);
+		
+		SkipTests.getSkipList("normal_event_skip_list.txt");
+		
+		// For now an array of species names are used. I will likely change it so that the wrapper calls each organism individually during release.
+		ArrayList<String> speciesList = new ArrayList<String>(Arrays.asList("pfal", "spom", "scer", "ddis", "cele", "sscr", "btau", "cfam", "mmus", "rnor", "ggal", "tgut", "xtro", "drer", "dmel", "atha", "osat"));
```
jweiser: Declaration better as List<String> speciesList

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: Yes not sure why I used the Object type here

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: I changed it to this:
`List<String> speciesList = new ArrayList<String>(Arrays.asList(...`

Is there a reason List is better here?

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+ *
+ */
+
+// To Reviewers: I recommend following the code sequentially, rather then looking at each class separately.
+public class InferEvents
+{
+	static MySQLAdaptor dbAdaptor = null;
+	private static GKInstance instanceEdit;
+	private static GKInstance speciesInst;
+	private static HashMap<GKInstance,GKInstance> manualEventToNonHumanSource = new HashMap<GKInstance,GKInstance>();
```
jweiser: All Collections (Maps, Lists, Sets, etc.) are better to be declared with the interface.  It might be good to do a search of all files for the declarations and change them to use the interfaces rather than implementations like HashMap and ArrayList.

cookersjs: I'm not sure I understand the concept of Interfaces still, to be honest

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
jweiser: It's not strictly necessary, but it's Java convention to try to stick to interfaces when you can.  It allows you to swap in a better implementation if one comes along.  It's a minor thing, but the most common practice that I've seen.

cookersjs: Ohh so an Interface is just the generic/super version of objects?

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
jweiser: The main idea is that an interface only specifies what methods a class will have, but not how they will do things.  This way any code that implements the interface is agreeing to a contract that it will provide the methods and as a client, you and other developers know that you can call a function and get back a value (or not if the return is void).  This lets you know that you can do things with code the uses the interface but if the algorithm doing them changes under the hood, it won't require the code that uses it to change because the interaction is still the same from outside.  Using an interface as the declaring type also only permits you to use methods specified by the interface whereas the implementation allows you to use things that may be specific only to that class.  If you use those kinds of methods, it ties you to using that specific implementation and doesn't let you switch if you decide something else might be better or more appropriate later.

All that said, it may be no problem to use a specific implementation but if there's not a reason that you need to (like a method specific only to that implementation), sticking with the interface is likely better.

Some good discussion about it here:
https://stackoverflow.com/questions/9852831/polymorphism-why-use-list-list-new-arraylist-instead-of-arraylist-list-n

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+		UpdateHumanEvents.setAdaptor(dbAdaptor);
+		
+		SkipTests.getSkipList("normal_event_skip_list.txt");
+		
+		// For now an array of species names are used. I will likely change it so that the wrapper calls each organism individually during release.
+		ArrayList<String> speciesList = new ArrayList<String>(Arrays.asList("pfal", "spom", "scer", "ddis", "cele", "sscr", "btau", "cfam", "mmus", "rnor", "ggal", "tgut", "xtro", "drer", "dmel", "atha", "osat"));
+
+		JSONParser parser = new JSONParser();
+		Object obj = parser.parse(new FileReader(pathToSpeciesConfig));
+		JSONObject jsonObject = (JSONObject) obj;
```
jweiser: The case can to JSONObject can be done on the parse line:
`JSONObject jsonObject = (JSONObject) parser.parse(new FileReader(pathToSpeciesConfig));`

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+			InferReaction.setEligibleFilename(eligibleFilename);
+			InferReaction.setInferredFilename(inferredFilename);
+
+			// Set static variables (DB/Species Instances, mapping files) that will be repeatedly used
+			InferEvents.setInstanceEdits();
+			try {
+				HashMap<String,String[]> homologueMappings = InferEvents.readHomologueMappingFile(species, "hsap");
+				ProteinCount.setHomologueMappingFile(homologueMappings);
+				InferEWAS.setHomologueMappingFile(homologueMappings);
+				homologueMappings = new HashMap<String,String[]>();
```
jweiser: I was wondering why the homologueMappings variable is set to an empty HashMap here?

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,306 @@
...
+
+			JSONObject altRefDb = (JSONObject) speciesObject.get("alt_refdb");
+
+			// Creates two files that a) list reactions that are eligible for inference and b) those that are successfully inferred
+			String eligibleFilename = "eligible_" + species	+ "_75.txt";
+			String inferredFilename = "inferred_" + species + "_75.txt";
+			PrintWriter eligibleFile = new PrintWriter(eligibleFilename);
+			PrintWriter inferredFile = new PrintWriter(inferredFilename);
+			eligibleFile.close();
+			inferredFile.close();
```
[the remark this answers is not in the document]
cookersjs: Yea this is something I feel like I must be doing wrong. Done this way, it actually creates the (empty) file. Later on, lines are appended to the files

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: Hmm that might be an artifact, or, more likely, it could have been an attempt at memory management (so still an artifact). In the summer there was a stretch of time where the memory usage of the script wasn't great, and I likely did this to null the mappings once they've been set. 

As for why I decided to do that by creating an empty HashMap instead of just using null, I do not know haha

orthoinference/src/main/java/org/reactome/orthoinference/GenerateInstance.java
```diff
@@ -0,0 +1,200 @@
...
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.schema.GKSchemaAttribute;
+import org.gk.schema.GKSchemaClass;
+import org.gk.schema.InvalidAttributeException;
+import org.gk.schema.InvalidAttributeValueException;
+import org.gk.schema.SchemaClass;
+
+
+// GenerateInstance is meant to act as a catch-all for functions that are instance-oriented, such as creating, mocking, or identical-checking.
+public class GenerateInstance {
```
SolomonShorser-OICR: Maybe rename this to "InstanceUtilities" - many methods in here don't seem to have anything to do with "generating" anything.

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,354 @@
...
+
+import org.gk.model.GKInstance;
+import org.gk.model.ReactomeJavaConstants;
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.schema.GKSchemaClass;
+import org.gk.schema.InvalidAttributeException;
+import org.gk.schema.InvalidAttributeValueException;
+import org.gk.schema.SchemaClass;
+
+public class InferEWAS {
```
SolomonShorser-OICR: Naming convention: classes are usually named as _nouns_. So this would be "EWASInferrer".

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,354 @@
...
+	private static GKInstance speciesInst;
+	private static HashMap<String, String[]> homologueMappings = new HashMap<String,String[]>();
+	private static HashMap<String, ArrayList<String>> ensgMappings = new HashMap<String,ArrayList<String>>();
+	private static HashMap<String, GKInstance> seenRGP = new HashMap<String,GKInstance>();
+	private static HashMap<String,GKInstance> ewasIdenticals = new HashMap<String,GKInstance>();
+	private static HashMap<String,GKInstance> residueIdenticals = new HashMap<String,GKInstance>();
+
+    // Creates an array of inferred EWAS instances from the homologue mappings file (hsap_species_mapping.txt).
+	@SuppressWarnings("unchecked")
+	public static ArrayList<GKInstance> ewasInferrer(GKInstance ewasInst) throws InvalidAttributeException, Exception
```
SolomonShorser-OICR: Naming convention: methods are usually named as _verbs_. So this would be "inferEwas".

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.List;
+
+import org.gk.model.ClassAttributeFollowingInstruction;
+import org.gk.model.GKInstance;
+import org.gk.model.InstanceUtilities;
+import org.gk.model.ReactomeJavaConstants;
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.schema.InvalidAttributeException;
+
+public class SkipTests {
```
SolomonShorser-OICR: Is this code for testing? If so, it should probably be moved from src/main/java/... to src/test/java/...

orthoinference/src/main/java/org/reactome/orthoinference/SpeciesCheck.java
```diff
@@ -0,0 +1,65 @@
...
+	// Unless its an 'OtherEntity' (which will return false), the function will check the instance or iterate on it's
+	// sub-instances until it finds an existing 'species' attribute, or else it will return false.
+	public static boolean hasSpecies(GKInstance entityInst) throws InvalidAttributeException, Exception
+	{
+		if (entityInst.getSchemClass().isa(ReactomeJavaConstants.OtherEntity))
+		{
+			return false;
+		} else if (entityInst.getSchemClass().isa(ReactomeJavaConstants.EntitySet)) // || entityInst.getSchemClass().isa(ReactomeJavaConstants.Polymer) || entityInst.getSchemClass().isa(ReactomeJavaConstants.EntitySet))
+		{
+			for (Object member : entityInst.getAttributeValuesList(ReactomeJavaConstants.hasMember))
```
SolomonShorser-OICR: `member` is probably always going to be a `GKInstance`, so you shouldn't need to use `Object` here.

orthoinference/src/main/java/org/reactome/orthoinference/GenerateInstance.java
cookersjs: Yea thats what I was thinking actually, it started out as an 'instance-specific' class, but gradually became more of a utility class

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
cookersjs: They are tests that determine if an instance should have an inference run it or not, effectively testing if they should be skipped

SolomonShorser-OICR: Ok, never mind.

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
```diff
@@ -0,0 +1,219 @@
...
+	private static MySQLAdaptor dba;
+//	private static ArrayList<String> skipList = new ArrayList<String>();
+	static HashSet<String> skipList = new HashSet<String>();
+	
+	public static void setAdaptor(MySQLAdaptor dbAdaptor)
+	{
+		dba = dbAdaptor;
+	}
+	
+	// Skiplist was traditionally provided in a file, but since it's currently just 3 instances, I've just hard-coded them here.
```
SolomonShorser-OICR: ...but you _will_ restore the ability to read them from a file, right?

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
```diff
@@ -0,0 +1,219 @@
...
+				@SuppressWarnings("unchecked")
+				Collection<GKInstance> followedInstances = InstanceUtilities.followInstanceAttributes(pathwayInst, classesToFollow, outClasses);
+				
+				for (GKInstance entity : followedInstances)
+				{
+					skipList.add(entity.getDBID().toString());
+				}
+			}
+		}
+		// Generates new skiplist file
```
SolomonShorser-OICR: The code below looks like it's _reading_ a file, not generating one.

SolomonShorser-OICR: I'm also confused: a comment earlier on says the instances to be skipped will be hard-coded, but then it seems to read more instances to skip from a file?

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
cookersjs: Yes I'll code that in

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
cookersjs: Sounds like comments from different stages of me working on it. The skip list portion of the code was completed in the summer and has worked comparatively to the Perl version of the code since then. I'll clean up the comments

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,364 @@
...
+		ensgDbInst.addAttributeValue(ReactomeJavaConstants.name, "ENSEMBL"); 
+		ensgDbInst.addAttributeValue(ReactomeJavaConstants.name, ensgSpeciesDb);
+		ensgDbInst.addAttributeValue(ReactomeJavaConstants.url, toSpeciesReferenceDbUrl);
+		ensgDbInst.addAttributeValue(ReactomeJavaConstants.accessUrl, toSpeciesEnsgAccessUrl);
+		ensgDbInst.setAttributeValue(ReactomeJavaConstants._displayName, "ENSEMBL");
+		ensgDbInst = GenerateInstance.checkForIdenticalInstances(ensgDbInst);
+	}
+	
+	// Create instance pertaining to any alternative reference DB for the species
+	public static void createAlternateReferenceDBInst(String toSpeciesLong, String alternateDbName, String toSpeciesAlternateDbUrl, String toSpeciesAlternateAccessUrl) throws InvalidAttributeException, InvalidAttributeValueException, Exception
```
jweiser: Can the first parameter be removed -- it doesn't seem to still be used.

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+			InferEWAS.readENSGMappingFile(species);
+			InferEWAS.createUniprotDbInst();
+			InferEWAS.createEnsemblProteinDbInst(speciesName, refDbUrl, refDbProteinUrl);
+			InferEWAS.createEnsemblGeneDBInst(speciesName, refDbUrl, refDbGeneUrl);
+			if (altRefDb != null)
+			{
+				JSONArray altRefDbNames = (JSONArray) altRefDb.get("dbname");
+				String altRefDbUrl = (String) altRefDb.get("url");
+				String altRefDbAccess = (String) altRefDb.get("access");
+				InferEWAS.createAlternateReferenceDBInst(speciesName, (String) altRefDbNames.get(0), altRefDbUrl, altRefDbAccess);
```
jweiser: This method might be better with a single parameter of altRefDb rather than decomposing it.

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+			JSONObject speciesObject = (JSONObject) jsonObject.get(species);
+			JSONArray speciesNames = (JSONArray) speciesObject.get("name");
+			String speciesName = (String) speciesNames.get(0);
+			System.out.println("Beginning orthoinference of " + speciesName + ".");
+			JSONObject refDb = (JSONObject) speciesObject.get("refdb");
+			String refDbUrl = (String) refDb.get("url");
+			String refDbProteinUrl = (String) refDb.get("access");
+			String refDbGeneUrl = (String) refDb.get("ensg_access");
+
+			JSONObject altRefDb = (JSONObject) speciesObject.get("alt_refdb");
```
jweiser: Maybe move this down to line 122 so it is declared just before it is used.

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,364 @@
...
+	public static void setHomologueMappingFile(HashMap<String, String[]> homologueMappingsCopy) throws IOException
+	{
+		homologueMappings = homologueMappingsCopy;
+	}
+	
+	// Read the species-specific ENSG gene-protein mappings, and create a Hashmap with the contents
+	public static void readENSGMappingFile(String toSpecies) throws IOException
+	{
+		String mappingFileName = toSpecies + "_gene_protein_mapping.txt";
+		String mappingFilePath = "src/main/resources/orthopairs/" + mappingFileName;
```
SolomonShorser-OICR: Will this file _always_ be included in src/main/resources? Or would it be better to make this configurable? I don't know enough about this to know if it's OK to hard-code this path like this, or if it should be more flexible/configurable...

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+public class InferEvents
+{
+	static MySQLAdaptor dbAdaptor = null;
+	private static GKInstance instanceEdit;
+	private static GKInstance speciesInst;
+	private static HashMap<GKInstance,GKInstance> manualEventToNonHumanSource = new HashMap<GKInstance,GKInstance>();
+	private static ArrayList<GKInstance> manualHumanEvents = new ArrayList<GKInstance>();
+
+	@SuppressWarnings("unchecked")
+	public static void eventInferrer(Properties props, String pathToConfig, String pathToSpeciesConfig) throws Exception
```
jweiser: pathToConfig doesn't look like it is used -- can it be removed?

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+	{
+		// Set up DB adaptor using config.properties file
+		String speciesToInferFromLong = "Homo sapiens";
```
jweiser: This is only used on line 147.  Can this be a inlined there or made a configuration property?

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
cookersjs: So it will be configurable for release, since the Orthopairs files will be under /orthopairs/{releaseNumber} -- just temporarily hard-coded in my workspace

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: Yep, InferEvents used to be the Main class for this script and when I moved the 'Main' to its own class, I kept the arguments the same

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: Yea there are a few artifacts related to the 'FromSpecies', and I think it started out due to me mimicing the Perl script, which can do inferences 'from' other species aside from humans. I can remove them, or keep them there, depending on if we want the same functionality for this script

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
SolomonShorser-OICR: Ok... Just remember that the final slice is supposed to happen on Wednesday, and OrthoInference will happen shortly after (Thursday or Friday, probably). I hope you'll have enough time to code and test the changes necessary to get it working on the reactomerelease!

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+		InferEWAS.setAdaptor(dbAdaptor);
+		UpdateHumanEvents.setAdaptor(dbAdaptor);
+		
+		SkipTests.getSkipList("normal_event_skip_list.txt");
+		
+		// For now an array of species names are used. I will likely change it so that the wrapper calls each organism individually during release.
+		List<String> speciesList = new ArrayList<String>(Arrays.asList("pfal", "spom", "scer", "ddis", "cele", "sscr", "btau", "cfam", "mmus", "rnor", "ggal", "tgut", "xtro", "drer", "dmel", "atha", "osat"));
+
+		JSONParser parser = new JSONParser();
+		Object obj = parser.parse(new FileReader(pathToSpeciesConfig));
```
jweiser: Not a priority before release, but it might be cleaner to create a Java class to represent a speciesConfig that handles all the JSON parsing and provides getters that InferEvents can call for the info it needs.

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+			GenerateInstance.setSpeciesInst(speciesInst);
+			InferEvents.setSummationInst();
+			InferEvents.setEvidenceTypeInst();
+			OrthologousEntity.setComplexSummationInst();
+
+/**
+ *  Start of ReactionlikeEvent inference. Retrieves all human ReactionlikeEvents, and attempts to infer each for the species.
+ */
+		// Gets DB instance of source species (human)
+		Collection<GKInstance> sourceSpeciesInst = (Collection<GKInstance>) dbAdaptor.fetchInstanceByAttribute("Species", "name", "=", speciesToInferFromLong);
```
jweiser: An extra indent is needed from this point

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+						}
+					}
+					for (GKInstance inferredFromInst : (Collection<GKInstance>) reactionInst.getAttributeValuesList(ReactomeJavaConstants.inferredFrom))
+					{
+						GKInstance reactionSpeciesInst = (GKInstance) inferredFromInst.getAttributeValue(ReactomeJavaConstants.species);
+						if (reactionSpeciesInst.getDBID() == speciesInst.getDBID() && inferredFromInst.getAttributeValue(ReactomeJavaConstants.isChimeric) == null)
+						{
+							previouslyInferredInstances.add(inferredFromInst);
+						}
+					}
```
jweiser: This loop and the previous loop can be extracted into a common method passing in either `ReactomeJavaConstants.orthologousEvent` or `ReactomeJavaConstants.inferredFrom`.

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+		InferEvents.outputReport(species);
+		InferEvents.resetVariables();
+//		System.gc();
+		System.out.println("Finished orthoinference of " + speciesName + ".");
+		}
+	}
+
+	public static void outputReport(String species) throws IOException
+	{
+		int[] counts = InferReaction.getCounts();
```
jweiser: It's probably clearer to separate this into two getters rather than returning an array.
`InferReaction.getEligibleCount()` and `InferReaction.getInferredCount()`

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+		int percent = 100*counts[1]/counts[0];
```
jweiser: The getters could actually be called inline here without a need for a temp variable

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -0,0 +1,307 @@
...
+		int port = Integer.valueOf(props.getProperty("port"));
+		
+		dbAdaptor = new MySQLAdaptor(host, database, username, password, port);
+		// Statically store the adaptor variable in each class
+		InferReaction.setAdaptor(dbAdaptor);
+		SkipTests.setAdaptor(dbAdaptor);
+		GenerateInstance.setAdaptor(dbAdaptor);
+		OrthologousEntity.setAdaptor(dbAdaptor);
+		InferEWAS.setAdaptor(dbAdaptor);
+		UpdateHumanEvents.setAdaptor(dbAdaptor);
```
jweiser: Maybe these could all be set together in one method call like with 'resetVariables'

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
```diff
@@ -0,0 +1,219 @@
...
+//	private static ArrayList<String> skipList = new ArrayList<String>();
+	static HashSet<String> skipList = new HashSet<String>();
+	
+	public static void setAdaptor(MySQLAdaptor dbAdaptor)
+	{
+		dba = dbAdaptor;
+	}
+	
+	// Skiplist was traditionally provided in a file, but since it's currently just 3 instances, I've just hard-coded them here.
+	public static void getSkipList(String skipListFilename) throws NumberFormatException, Exception
```
jweiser: NumberFormatException isn't needed since Exception is more general and will cover it

jweiser: The same is true with other specific exception declarations in methods below

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
cookersjs: Yep that would clean this part of the code a good bit

orthoinference/src/main/java/org/reactome/orthoinference/SkipTests.java
SolomonShorser-OICR: Throwing more specific exceptions allows the caller to catch them and handle them, if necessary. The caller might be able to deal with a NumberFormatException, but not the others.

cookersjs: These were populated automatically by Eclipse -- it would say something akin to '{Exception name} needed and would do it

jweiser: True, that way is better from a documentation perspective and the other is clearer for what the compiler needs.  Whichever makes more sense and is clearer in the code is fine with me.

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,384 @@
...
+	}
+	
+    // These are setup functions called at the beginning of the 'inferEvent' script 
+	
+	public static void setAdaptor(MySQLAdaptor dbAdaptor)
+	{
+		InferEWAS.dba = dbAdaptor;
+	}
+	// Sets the HashMap of species-specific homologue-identifier mappings
+	public static void setHomologueMappingFile(HashMap<String, String[]> homologueMappingsCopy) throws IOException
```
jweiser: This might be clearer if named `setHomologueMapping`

orthoinference/src/main/java/org/reactome/orthoinference/InferEWAS.java
```diff
@@ -0,0 +1,384 @@
...
+						ArrayList<GKInstance> inferredReferenceDNAInstances = InferEWAS.createReferenceDNASequence(homologueId);
+						infReferenceGeneProduct.addAttributeValue(ReactomeJavaConstants.referenceGene, inferredReferenceDNAInstances);
+						
+						infReferenceGeneProduct.addAttributeValue(ReactomeJavaConstants.species, speciesInst);
+						infReferenceGeneProduct.setAttributeValue(ReactomeJavaConstants._displayName, "UniProt:" + homologueId);
+						infReferenceGeneProduct = InstanceUtilities.checkForIdenticalInstances(infReferenceGeneProduct);
+						seenRGP.put(homologueId, infReferenceGeneProduct);
+					} else {
+						infReferenceGeneProduct = seenRGP.get(homologueId);
+					}
```
jweiser: Obtaining the inferred reference gene product makes sense to be extracted as its own private method

orthoinference/src/main/java/org/reactome/orthoinference/CandidateSetsConvertedToDefinedSetsSpeciesFix.java
```diff
@@ -0,0 +1,78 @@
...
+ * from a CandidateSet to DefinedSet instance. At the location of the code in question (OrthologousEntity.createInfEntitySet in Java Orthoinference),
+ * the new instance is produced by creating a new GKInstance from a DefinedSet SchemaClass, rather than using the InstanceUtilities.newInferredGKInstance
+ * method that would normally be used (using that method would create a new CandidateSet instance, based on the source instance).
+ * 
+ * The mistake was that the 'species' and 'created' slots that are populated in 'newInferredGKInstance' were not populated when the new DefinedSet was created.
+ * This code fixes that mistake (Justin Cook November 2018).
+ * 
+ */
+
+public class CandidateSetsConvertedToDefinedSetsSpeciesFix {
```
SolomonShorser-OICR: I think it might be better if one-time data fix code is maintained in a separate directory or project within the repository. The way it is right now, any future builds of the OrthoInference application will include this code, and it is completely unnecessary for them to do so. As an example, the KEGGIdentifierFix was moved to a separate project in the AddLinks repository (https://github.com/reactome/add-links/tree/develop/KeggIdentifierFixer). I probably could have also moved that into a "datafix" subdirectory to distinguish it from the main project, and I probably _will_ do that once there is more than a couple of data fix projects.

cookersjs: Fair enough, I wanted it to be visible for transparency's sake but I'll move it to something akin to a 'datafix' directory

SolomonShorser-OICR: Visibility is a very good thing, so thank you for doing that! :)

We just need to keep the code organized a little better than was done in the past. I tried to clean up the current Release repo by also adding a "fixes" directory for one-time fixer scripts (https://github.com/reactome/Release/tree/repository_cleanup/scripts/fixes) but by the time I finish the work on that cleanup effort, that whole repository might be mostly useless.

orthoinference/src/main/java/org/reactome/orthoinference/InferEvents.java
```diff
@@ -262,50 +239,65 @@ public static void resetVariables() {
...
+		evidenceTypeInst.addAttributeValue(_displayName, evidenceTypeText);
 		evidenceTypeInst = InstanceUtilities.checkForIdenticalInstances(evidenceTypeInst);
-		InferReaction.setEvidenceTypeInst(evidenceTypeInst);
-		UpdateHumanEvents.setEvidenceTypeInst(evidenceTypeInst);
+		InferReaction.setEvidenceTypeInstance(evidenceTypeInst);
+		UpdateHumanEvents.setEvidenceTypeInstance(evidenceTypeInst);
+	}
+	
+	private static void setInstanceEdits() throws Exception {
+		instanceEditInst = InstanceEditUtils.createInstanceEdit(dbAdaptor, 8948690, "org.reactome.orthoinference");
```
SolomonShorser-OICR: What is the significance of this magic number: 8948690?

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -0,0 +1,385 @@
...
+	public static void setHomologueMappingFile(Map<String, String[]> homologueMappingsCopy) throws IOException
+	{
+		homologueMappings = homologueMappingsCopy;
+	}
+	
+	// Read the species-specific ENSG gene-protein mappings, and create a Hashmap with the contents
+	public static void readENSGMappingFile(String toSpecies) throws IOException
+	{
+		String mappingFileName = toSpecies + "_gene_protein_mapping.txt";
+		String mappingFilePath = "src/main=/resources/orthopairs/" + mappingFileName;
```
SolomonShorser-OICR: Is that "=" a typo?

SolomonShorser-OICR: Also, will the files ALWAYS be on that path?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -0,0 +1,315 @@
...
+		JSONArray speciesNames = (JSONArray) speciesObject.get("name");
+		String speciesName = (String) speciesNames.get(0);
+		System.out.println("Beginning orthoinference of " + speciesName + ".");
+		JSONObject refDb = (JSONObject) speciesObject.get("refdb");
+		String refDbUrl = (String) refDb.get("url");
+		String refDbProteinUrl = (String) refDb.get("access");
+		String refDbGeneUrl = (String) refDb.get("ensg_access");
+		
+		// Creates two files that a) list reactions that are eligible for inference and b) those that are successfully inferred
+		String eligibleFilename = "eligible_" + species	+ "_75.txt";
```
SolomonShorser-OICR: I'm curious about the significance of "75" in the filename...

SolomonShorser-OICR:
Looks good, almost ready to merge (but I think that "=" in the path needs to be fixed, and maybe run the code once more to make sure these changes haven't broken anything), and also resolve the merge conflict on .gitignore.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: In the Perl version the threshold was 75 for inference to be considered valid. It was a variable that could be modified (though it never was). It could be removed and it remained there due to my Perl results mimicry

cookersjs:
Yep I've fixed the path so that its configged out now. Currently running the code and I've added logging as well

cookersjs:
README now has instructions for running Orthoinference and provides some explanation to the process as well

PR #18, opened 2018-12-10
Feature/update stable ids

cookersjs:
Rewrite for UpdateStableIds. This excludes the 'restore_database.pl' script since that backs up the database via a mysqldump command, which we're going to incorporate into the eventual wrapper.

This script is different from the Perl version of UpdateStableIds in that it takes all instances from Event and PhysicalEntity classes, rather than specifically from Pathways, EWAS', Complexes, etc. The Java API was taking sub-classes of instances (for example GenomeEncodedEntity had ~760 instances in Perl but 30000+ in Java since 'fetchInstancesByClass' method includes all child instances, which was all EWAS' in this case). Since UpdateStableIds only updates instances under the Event and PhysicalEntity nodes, this change will also be able to properly update any new class types that qualify (such as Drugs, which weren't getting updated in the Perl version)


updateStableIds/src/main/java/org/reactome/release/update_stable_ids/DatabaseUtils.java
```diff
@@ -0,0 +1,10 @@
+package org.reactome.release.update_stable_ids;
+
+public class DatabaseUtils {
```
SolomonShorser-OICR: It doesn't look like this class is doing anything useful. Please remove it or make it more useful.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/Main.java
```diff
@@ -0,0 +1,66 @@
...
+	private static final Logger logger = LogManager.getLogger();
+	
+    public static void main( String[] args ) throws Exception
+    {
+    	logger.info("Beginning UpdateStableIds step...");
+    	// This utility was meant to serve the same function as 'restore_database.pl'. 
+    	// For now, we don't want the Java to do the DB archiving, so this won't be used. 
+//       DatabaseUtils.archiveUsedDatabases("test_slice_67", "gk_central");
+       
+       String pathToConfig = "src/main/resources/config.properties";
```
SolomonShorser-OICR: Users should be able to specify the path to the properties file from the command line. It's Ok to have a value like this as a _default_, but it should be possible for users to override this, if necessary.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,170 @@
...
+		GKInstance sliceIE = InstanceEditUtils.createInstanceEdit(dbaSlice, personId, creatorName);
+		GKInstance gkCentralIE = InstanceEditUtils.createInstanceEdit(dbaGkCentral, personId, creatorName);
+
+		// At time of writing (December 2018), test_slice is a non-transactional database. This check has been put in place as a safety net in case that changes.
+		if (dbaSlice.supportsTransactions()) {
+			dbaSlice.startTransaction();
+		}
+		dbaGkCentral.startTransaction();
+		
+		//TODO:Should legacy regulation instances be updated?
```
SolomonShorser-OICR: Should this question be answered/resolved before submitting a pull-request, or is this a longer-term issue?

cookersjs: This one is longer-term, I put it in the code so that it wasn't forgotten

updateStableIds/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.reactome.release</groupId>
+  <artifactId>update-stable-ids</artifactId>
+  <version>0.0.1-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>update-stable-ids</name>
+  <url>http://maven.apache.org</url>
```
SolomonShorser-OICR: This is not the URL for your project. You might want to remove this element.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,170 @@
...
+		//TODO:Ensure that re-run handling is implemented (backing up DB)
```
SolomonShorser-OICR: I think this comment needs more explaining...?

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,170 @@
...
+				}
+			}
+		}
+
+		if (dbaSlice.supportsTransactions()) {
+			dbaSlice.commit();
+		}
+		logger.info("Commiting all changes in " + dbaGkCentral.getDBName());
+		dbaGkCentral.commit();
+		logger.info(incrementCount + " Stable Identifiers were updated");
```
SolomonShorser-OICR: Would it also be useful to gather on the number of Stable Identifiers that were checked but did not need updating? Not sure if users would find that informative. Might not be necessary.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
cookersjs: Actually it creates a 'test_slice_XX_snapshot', and it gets used only in the following release. Basically it will look at all instances in the current test_slice, and compare the number of 'Modified' attributes it has with the instance from the snapshot

jweiser:
Minor style point -- could the indentation be changed from 8 spaces to 4 spaces?  I find it's easier to read and helps reduce line width from getting too long.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,170 @@
...
+import org.gk.model.GKInstance;
+import org.gk.model.ReactomeJavaConstants;
+
+import static org.gk.model.ReactomeJavaConstants.*;
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.persistence.TransactionsNotSupportedException;
+import org.gk.schema.GKSchemaClass;
+import org.gk.schema.InvalidAttributeException;
+import org.reactome.release.common.database.InstanceEditUtils;
+
```
jweiser: Can some of the unused imports be removed?

cookersjs: Yea, I think that Eclipse tries to automatically remove them if they're not used, but its less-than-great at it

cookersjs: Gotcha, I'll try to make that a habit before committing anything in the future

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,163 @@
...
+		List<GKInstance> sliceInstances = new ArrayList<GKInstance>();
+		sliceInstances.addAll(eventInstances);
+		sliceInstances.addAll(physicalEntityInstances);
+
+		int incrementCount = 0;
+		logger.info("Total instances to check: " + sliceInstances.size());
+		for (GKInstance sliceInstance : sliceInstances) {
+			logger.info("Checking " + sliceInstance);
+			GKInstance gkCentralInstance = (GKInstance) dbaGkCentral.fetchInstance(sliceInstance.getDBID());
+			GKInstance prevSliceInstance = (GKInstance) dbaPrevSlice.fetchInstance(sliceInstance.getDBID());
```
jweiser: The fetchInstance method looks like it returns a GKInstance object, so no need to cast for these two lines.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
```diff
@@ -0,0 +1,163 @@
...
+						incrementStableIdentifier(sliceInstance, dbaSlice, sliceIE);
+						incrementStableIdentifier(gkCentralInstance, dbaGkCentral, gkCentralIE);
+						incrementCount++;
+					}
+				} else {
+					if (sliceInstanceModified.size() == prevSliceInstanceModified.size()) {
+//						System.out.println("No change between releases");
+					} else {
+						logger.warn("Unexpected: Number of 'Modified' instances in [" + sliceInstance.getDBID() + "] is fewer than in previous release");
+					}
```
jweiser: The if statement is empty, so it might be best to make line 73 an else if block like this:
```
else if (sliceInstanceModified.size() < prevSliceInstanceModified.size()) {
    logger.warn("...");
}
``

**Comment by jweiser** [comment]
Added a few comments, but overall it looks good to me!

File: updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java [hunk 10]
```diff
@@ -0,0 +1,163 @@
...
+				}
+				
+				// Instances that have been updated already during the current release will have their 'releaseStatus' attribute equal to 'UPDATED'.
+				// This will make sure that StableIDs are only updated once per release.
+				try {
+					if (isUpdated(sliceInstance, prevSliceInstance, dbaPrevSlice)) {
+						String releaseStatusString = (String) sliceInstance.getAttributeValue(releaseStatus);
+						String updated = "UPDATED";
+						if (releaseStatusString == null || !releaseStatusString.contentEquals(updated)) {
```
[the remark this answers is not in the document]
jweiser: Hmm, that's weird.  The equals method on String should check for value equality.  Could it have been a case issue?  You could try 'equalsIgnoreCase'.

I just haven't seen that method used before.  I think it is intended if you are comparing to another type of object representing a character sequence that isn't a String object.

https://stackoverflow.com/questions/6476600/difference-between-stringequals-and-stringcontentequals-methods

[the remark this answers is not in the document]
SolomonShorser-OICR: contentEquals: https://stackoverflow.com/a/6476612

It looks like `contentEquals` will compare the outputs of `toString` functions, for *any* object that responds to `toString`. `equals` will also compare types. So I think you can use `contentEquals` to compare a StringBuilder to a StringBuffer, which would be shorter and easier to read than `stringBuilder.toString().equals(stringBuffer.toString())`.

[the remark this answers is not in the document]
jweiser: Yup, that's my understanding too.  I think here `equals` or `equalsIgnoreCase` should work, though, because both objects are Strings.

updateStableIds/src/main/java/org/reactome/release/update_stable_ids/UpdateStableIds.java
jweiser: It seems like for now, we are not concerned about stable identifiers for regulation instances.  The idea of having more classes receive stable identifiers was mentioned, though (e.g. catalyst activities and possibly regulation instances again).  So, this is something to watch but nothing to do right now.

updateStableIds/pom.xml
```diff
@@ -57,4 +57,21 @@
...
   </repositories>
   
+  <build>
+	  <plugins>
+	    <plugin>
+	      <artifactId>maven-assembly-plugin</artifactId>
+	      <configuration>
+	        <archive>
+	          <manifest>
+	            <mainClass>org.reactome.release.update_stable_ids.Main</mainClass>
```
SolomonShorser-OICR: The Java convention is that package names do not contain underscores, unless they are a replacement for a hyphen/dash in a domain name. So, the package `update_stable_ids` would normally be written as `updatestableids` or `updateStableIds`.

https://stackoverflow.com/a/3179277 - with links to other references on package name conventions.

jweiser:
The files from the .idea made me think.  I usually have the whole folder in .gitignore, but it looks like there is debate about committing these files to share IDE configurations.

https://stackoverflow.com/questions/11968531/what-to-gitignore-from-the-idea-folder

We should probably discuss this a bit, but it might be better to leave them out of our projects since we have maven to build things and we are using different IDEs within Reactome.

What do you think?

SolomonShorser-OICR:
IDEs can produce tonnes of extra files and junk. Sometimes they reference specific versions of specific plugins that not all other users might have or want. Personally, I don't think IDE files have any business in the repository, unless they are actually _necessary_ to run the project. I guess if we were using some very special plugin that required a very specific configuration, it would be OK (or necessary!) to check the IDE files in to source control. But we are running simple Java applications that can be built/run from the command line, so I don't think IDE files should be in source control. If you _do_ want to share your IDE config with other developers, I'd recommend a separate directory in the project, named "IDE_CONFIG" and then have subdirectories for various IDEs (IntelliJ, Eclipse, atom, etc...). But if they are going to go in to source control, it should be done in a such a way that other developers won't start using them _by accident_. Because it can be extremely frustrating/annoying when a workspace magically configures itself, or _tries_ to and breaks in incomprehensible ways.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -94,9 +95,11 @@ public static void stableIdUpdater(MySQLAdaptor dbaSlice, MySQLAdaptor dbaPrevSl
 							logger.info("StableIdentifer has already been updated during this release");
 						}
 					}
-				} catch (Exception e) {
+				} catch (StackOverflowError e) {
```
SolomonShorser-OICR: StackOverflowError seems oddly specific. I'm just curious: what is triggering it?

cookersjs: It shouldn't happen in reality, but it came up in testing due to a recursive call

updateStableIds/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+	  <dependency>
+		  <groupId>junit</groupId>
+		  <artifactId>junit</artifactId>
+		  <version>4.12</version>
+		  <scope>test</scope>
+	  </dependency>
+	  <dependency>
+		  <groupId>org.powermock</groupId>
+		  <artifactId>powermock-api-mockito</artifactId>
+		  <version>1.6.5</version>
```
SolomonShorser-OICR: You can move "1.6.5" to a property. Add `<powermockito.version>1.6.5</powermockito.version>` to the `<properties>...</properties>` list, and then you can refer to it on line 52 and 58 as `${powermockito.version}`. Not necessary, but makes version management easier. It's more of an issue when you need to refer to the same version number string for a large list of dependencies.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.*;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.gk.model.GKInstance;
+import org.gk.persistence.MySQLAdaptor;
+import org.gk.schema.InvalidAttributeException;
+import org.reactome.release.common.database.InstanceEditUtils;
+
+public class UpdateStableIds {
```
SolomonShorser-OICR: The class name should probably be changed to `StableIDUpdater` and the method `stableIdUpdater` should probably be changed to `updateStableIDs` to conform to more common conventions.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+		GKInstance sliceIE = InstanceEditUtils.createInstanceEdit(dbaSlice, personId, creatorName);
+		GKInstance gkCentralIE = InstanceEditUtils.createInstanceEdit(dbaGkCentral, personId, creatorName);
+
+		// At time of writing (December 2018), test_slice is a non-transactional database. This check has been put in place as a safety net in case that changes.
+		if (dbaSlice.supportsTransactions()) {
+			dbaSlice.startTransaction();
+		}
+		dbaGkCentral.startTransaction();
+
+		//TODO: Perl wrapper will create a 'snapshot' of the previous slice -- once the wrapper is retired this needs to be done
```
SolomonShorser-OICR: Is this still a "TODO" or is it done?

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+		
+		// These two Lists were originally used to determine what classes' instances should be updated.
+		// Since all 'accepted' classes turned out to encompass all PhysicalEntity and Event classes, and all
+		// 'excluded' classes were Regulation classes, the filter was not needed anymore.
+//		final List<String> acceptedClassesWithStableIdentifiers = Arrays.asList("Pathway","SimpleEntity","OtherEntity","DefinedSet","Complex","EntityWithAccessionedSequence","GenomeEncodedEntity","Reaction","BlackBoxEvent","CandidateSet","Polymer","Depolymerisation","Polymerisation","Drug","FailedReaction","EntitySet");
+//		List<String> excludedClassesWithStableIdentifiers = Arrays.asList("PositiveGeneExpressionRegulation","PositiveRegulation","NegativeRegulation","Requirement","NegativeGeneExpressionRegulation");
+//		ResultSet classesWithStableIdentifiers = dbaGkCentral.executeQuery("SELECT DISTINCT _class FROM DatabaseObject WHERE StableIdentifier IS NOT NULL", null);
```
SolomonShorser-OICR: Can the above 3 lines be removed?

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: Right, I forgot you pointed that out

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: Its still a TODO, since we're still using Perl wrappers for script calling. This is a database step, so for now I've decided to let Perl handle it

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
SolomonShorser-OICR: Remember to use `git mv` for file renaming - you'll probably need to do that first, if you want to preserve file history in git.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: yes they can be

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
SolomonShorser-OICR: Ok.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+		}
+		logger.info("Commiting all changes in " + dbaGkCentral.getDBName());
+		dbaGkCentral.commit();
+		logger.info(incrementedCount + " Stable Identifiers were updated");
+		logger.info(notIncrementedCount + "were not updated");
+		logger.info("UpdateStableIdentifiers step has finished");
+		System.out.println(incrementedCount);
+	}
+
+	// Increments the identifierVersion attribute and updates the StableIdentifier displayName accordingly. TODO: IT
```
SolomonShorser-OICR: The "IT" in "TODO: IT" - Is that for "Integration Test"?

updateStableIds/src/test/java/org/reactome/release/updateStableIds/UpdateStableIdsTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Mockito.when(mockInstance.getSchemClass().isa("Event")).thenReturn(true);
+
+        Mockito.when(mockInstance.getAttributeValuesList("reviewed")).thenReturn(sliceList);
+        Mockito.when(mockInstance2.getAttributeValuesList("reviewed")).thenReturn(sliceList2);
+
+        UpdateStableIds.stableIdUpdater(mockAdaptor, mockAdaptor2, mockAdaptor3, 12345L);
+    }
+
+    @Test
+    public void stableIdUpdaterModifiedListTest() throws Exception {
```
SolomonShorser-OICR: What is the difference between this test method and the other test method?

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+			logger.info("Checking " + sliceInstance);
+			GKInstance gkCentralInstance = dbaGkCentral.fetchInstance(sliceInstance.getDBID());
+			GKInstance prevSliceInstance = dbaPrevSlice.fetchInstance(sliceInstance.getDBID());
+			// Check if instance is new and that it exists on gkCentral (they could be deleted)
+			if (prevSliceInstance != null && gkCentralInstance != null) {
+				
+				// Compare number of 'Modified' instances between slices
+				Collection<GKInstance> sliceInstanceModified = sliceInstance.getAttributeValuesList(modified);
+				Collection<GKInstance> prevSliceInstanceModified = prevSliceInstance.getAttributeValuesList(modified);
+				if (sliceInstanceModified.size() > prevSliceInstanceModified.size()) {
```
SolomonShorser-OICR: Is it _certain_ that `sliceInstanceModified` & `prevSliceInstanceModified` are non-null at this point? If an instance has no `modified`s, will an empty list be returned or null? Is this something that needs to be checked? Or will this code only run for things that have `modified`s? I would have thought that a stableIdentifier that was brand-new in the previous release might _not_ have anything in its `modified` attribute - but maybe it's OK if `getAttributeValuesList` returns an empty list instead of NULL.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: yes ahah, not very clear is it

updateStableIds/pom.xml
SolomonShorser-OICR: (not necessary for this release, just thought you'd like to know that it's an option)

updateStableIds/src/test/java/org/reactome/release/updateStableIds/UpdateStableIdsTest.java
cookersjs: The list returned from `instance.getAttributeValuesList("modified")` is the same size in the second test

SolomonShorser-OICR: Ah. "...ModifiedList..." did not indicate that to me. I would try to include small comments explaining these differences, or rename the method to `stableIdUpdaterTestWithSameSizedLists`... which is long and unwieldly, though clear...

cookersjs: I think we'd made a decision earlier on that long, unwieldy but clear was preferable, I'll fix it

updateStableIds/runUpdateStableIds.sh
```diff
@@ -0,0 +1,9 @@
+## Update repo
```
jweiser: Could this be added to the beginning of the bash script to allow it to be called from any directory:

```
#!/bin/bash
cd $DIR
```

SolomonShorser-OICR: Is it expected that this script will be invoked from _any_ other directories?

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+					// Make sure StableIdentifier instance exists
+					if (sliceInstance.getAttributeValue(stableIdentifier) != null && gkCentralInstance.getAttributeValue(stableIdentifier) != null) {
+						logger.info("\tIncrementing " + sliceInstance.getAttributeValue(stableIdentifier));
+						incrementStableIdentifier(sliceInstance, dbaSlice, sliceIE);
+						incrementStableIdentifier(gkCentralInstance, dbaGkCentral, gkCentralIE);
+						incrementedCount++;
+					} else if (sliceInstance.getAttributeValue(stableIdentifier) == null){
+						logger.warn(sliceInstance + ": could not locate StableIdentifier instance");
+					} else {
+						logger.warn(prevSliceInstance + ": Instance from previous slice did not have StableIdentifier instance");
```
jweiser: Can this be an info statement instead of a warn statement since this case can occur legitimately when a new instance exists that didn't exist in the previous release?

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
```diff
@@ -0,0 +1,171 @@
...
+					}
+				} else {
+					if (sliceInstanceModified.size() < prevSliceInstanceModified.size()) {
+						logger.warn("Unexpected: Number of 'Modified' instances in [" + sliceInstance.getDBID() + "] is fewer than in previous release");
```
jweiser: This should probably be an error since this case should never occur.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: Yes I can change it, I think at time of writing I didn't realize this wasn't a negative thing

updateStableIds/src/test/java/org/reactome/release/updateStableIds/UpdateStableIdsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Mockito.when(mockInstance.getAttributeValue("identifierVersion")).thenReturn("1");
+
+        Mockito.when(mockInstance.getSchemClass()).thenReturn(mockSchemaClass);
+        Mockito.when(mockInstance.getSchemClass().isa("Event")).thenReturn(true);
+
+        Mockito.when(mockInstance.getAttributeValuesList("reviewed")).thenReturn(sliceList);
+        Mockito.when(mockInstance2.getAttributeValuesList("reviewed")).thenReturn(sliceList2);
+
+        UpdateStableIds.stableIdUpdater(mockAdaptor, mockAdaptor2, mockAdaptor3, 12345L);
```
jweiser: Where they are used to mock one specific thing, could the mock instances have more descriptive names like mockSliceAdaptor, mockPrevSliceAdaptor, mockGKCentralAdaptor, etc. for context of what they are replacing?

updateStableIds/runUpdateStableIds.sh
jweiser: I'm not sure yet, but I think it should have the ability to be invoked from elsewhere since it is the entrypoint to the entire program for the end-user.  We may need to invoke from another working directory as part of the pipeline and it's only a small addition to the script.

updateStableIds/src/main/java/org/reactome/release/updateStableIds/UpdateStableIds.java
cookersjs: It returns an empty list, which I think preserves what the code is attempting to do

updateStableIds/runUpdateStableIds.sh
cookersjs: Done

PR #19, opened 2018-12-12
Feature/go update

SolomonShorser-OICR:
Improvements to
 - Duplicate reporting and cleaning.
 - deleting GO terms.


go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,183 @@
...
+	@Override
+	public void executeStep(Properties props) throws Exception
+	{
+		MySQLAdaptor adaptor = DuplicateCleaner.getMySQLAdaptorFromProperties(props);
+		int totalDuplicatesCount = 0;
+		int instancesWithSignificantReferrers = 0;
+		
+		DuplicateReporter dupeReporter = new DuplicateReporter(adaptor);
+		Set<Long> dbIDsToDelete = new HashSet<Long>();
+		Set<Long> dbIDsToKeep = new HashSet<Long>();
```
jweiser: It looks like this is added to but not used -- can it be removed?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,183 @@
...
+					{
+						logger.debug("DB ID {} for accession {} has no significant referrers, and will be deleted.", dbId, accession);
+						dbIDsWithNoReferrers.add(dbId);
+					}
+				}
+			}
+			// update list of things to delete with IDs that have no referrers.
+			dbIDsToDelete.addAll(dbIDsWithNoReferrers);
+			// If no DB IDs have significant referrers, we should delete ALL DB_IDs for the accession, except for the newest one.
+			if (dbIDsWithNoReferrers.size() == refCounts.keySet().size())
```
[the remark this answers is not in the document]
SolomonShorser-OICR: `refCounts` is initialized with a call to `DuplicateReporter`'s `getReferrerCountForAccession` method, which should always return a map, even an empty map.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,183 @@
...
+							{
+								for (GKInstance ref : refs)
+								{
+									if (!Arrays.asList(ReactomeJavaConstants.GO_BiologicalProcess, ReactomeJavaConstants.GO_MolecularFunction, ReactomeJavaConstants.GO_CellularComponent, ReactomeJavaConstants.Compartment).contains(ref.getSchemClass().getName()))
+									{
+										logger.info("\t\tvia "+refAtt.getName()+":\t"+ref.toString());
+									}
+								}
+							}
+						}
```
jweiser: It took me some time to understand this block -- could it be extracted to a method called `logSignificantReferrers(inst)`?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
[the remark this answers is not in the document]
jweiser: Ah okay, then the if statement on line 73 `if (refCounts != null && refCounts.size() > 0)` could be changed to just `if(!refCounts.isEmpty())`?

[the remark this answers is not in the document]
SolomonShorser-OICR: Probably.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+		DuplicateReporter dupeReporter = new DuplicateReporter(adaptor);
+		Set<Long> dbIDsToDelete = new HashSet<Long>();
+
+		Map<String, Integer> duplicates = dupeReporter.getDuplicateAccessions();
+		
+		// loop through all duplicated accessions.
+		for (String accession : duplicates.keySet())
+		{
+			totalDuplicatesCount = duplicates.keySet().size();
+			logger.info("Accession " + accession +" is duplicated " + duplicates.get(accession) + " times.");
```
jweiser: Most places for logging you are using {} for variable insertion which I really like for readability.  Could this be done here too and in all logging statements for consistency?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+								localDate = LocalDateTime.parse((CharSequence) modifiedInstance.getAttributeValue(ReactomeJavaConstants.dateTime), formatter);
+							}
+
+							if (newestInstance == null)
+							{
+								newestInstance = instance;
+								newestDate = localDate;
+								
+							}
+							if (localDate != null && newestDate != null && localDate.compareTo(newestDate) > 0)
```
jweiser: localDate and newestData will always be non-null according to IntelliJ

SolomonShorser-OICR: Both are explicitly initialized as `null`. `localDate` _might_ be safe, but I think there are situations the first time through the loop where `newestDate` might be null, so I'm keeping these checks.

jweiser: `newestDate` will always be set because the first time through the loop, `newestInstance` is null and it will be set to the `localDate` object.  In turn, `localDate` will also always be set because it is in both the if and else clause checking if `modifiedInstance` isn't null.

It's safe to remove the null checks for both.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+					@SuppressWarnings("unchecked")
+					Collection<GKInstance> instances = (Collection<GKInstance>) adaptor.fetchInstanceByAttribute(reactomeClass, ReactomeJavaConstants.accession, "=", accession);
+					// For each instance for this accession, try to find the newest one so that it won't be deleted.
+					if (instances != null && instances.size() > 0)
+					{
+						for (GKInstance instance : instances)
+						{
+							dbIDsToDelete.add(instance.getDBID());
+							LocalDateTime localDate = null;
+							modifiedInstance = (GKInstance) instance.getAttributeValue(ReactomeJavaConstants.modified);
```
jweiser: Declaration for modifiedInstance can be made here to reduce scope

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+							
+							if (modifiedInstance != null)
+							{
+								localDate = LocalDateTime.parse((CharSequence) modifiedInstance.getAttributeValue(ReactomeJavaConstants.dateTime), formatter);
+							}
+							else
+							{
+								modifiedInstance = (GKInstance) instance.getAttributeValue(ReactomeJavaConstants.created);
+								localDate = LocalDateTime.parse((CharSequence) modifiedInstance.getAttributeValue(ReactomeJavaConstants.dateTime), formatter);
+							}
```
jweiser: The assignment for localDate is the same in the if and the else blocks, so it can be moved outside the condition check.
```
if (modifiedInstance == null) {
    modifiedInstance = (GKInstance) instance.getAttributeValue(ReactomeJavaConstants.created);
}
localDate = LocalDateTime.parse((CharSequence) modifiedInstance.getAttributeValue(ReactomeJavaConstants.dateTime), formatter);

File: go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java [hunk 8]
```diff
@@ -0,0 +1,182 @@
...
+
+							if (newestInstance == null)
```
jweiser: Since they set the same variables and the same values, this check can be combined with next:
```
if (newestInstance == null || localDate.compareTo(newestDate) > 0) {
	newestInstance = instance;
	newestDate = localDate;
}
```

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+				GKInstance newestInstance = null;
+				LocalDateTime newestDate = null;
+				DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.S");
+				// Need to try all classes since we don't know which class a GO Accession might be.
+				for (String reactomeClass : Arrays.asList(ReactomeJavaConstants.GO_BiologicalProcess, ReactomeJavaConstants.GO_MolecularFunction, ReactomeJavaConstants.GO_CellularComponent))
+				{
+					GKInstance modifiedInstance = null;
+					
+					@SuppressWarnings("unchecked")
+					Collection<GKInstance> instances = (Collection<GKInstance>) adaptor.fetchInstanceByAttribute(reactomeClass, ReactomeJavaConstants.accession, "=", accession);
```
jweiser: The for loop to look at the different GO classes could be done in a method so the detail of grabbing GO instances is extracted away and the nesting is reduced.  Something like:

```
private List<GKInstance> getGOInstancesByAccession(MySQLAdaptor adaptor, String accession) throws Exception {
	List<GKInstance> GOInstances = new ArrayList<>();
	for (String reactomeClass : Arrays.asList(
		ReactomeJavaConstants.GO_BiologicalProcess,
		ReactomeJavaConstants.GO_MolecularFunction,
		ReactomeJavaConstants.GO_CellularComponent)
	) {

	    Collection<GKInstance> instances = (Collection<GKInstance>) adaptor.fetchInstanceByAttribute(reactomeClass, ReactomeJavaConstants.accession, "=", accession);
            if (GOInstances != null) {	    
                GOInstances.addAll(instances);
            }
	}
	return GOInstances;
}
```

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,182 @@
...
+		
+		DuplicateReporter dupeReporter = new DuplicateReporter(adaptor);
+		Set<Long> dbIDsToDelete = new HashSet<Long>();
+
+		Map<String, Integer> duplicates = dupeReporter.getDuplicateAccessions();
+		
+		// loop through all duplicated accessions.
+		for (String accession : duplicates.keySet())
+		{
+			totalDuplicatesCount = duplicates.keySet().size();
```
jweiser: Should this be `+=` instead of `=` since the count is used after the for loop is finished?

jweiser: Or could the `totalDuplicatesCount` variable be removed and the logger statement using it just call `duplicates.keySet().size()` directly?

jweiser:
For consistency, you can leave out the type parameter when creating concrete lists, sets, and maps as of Java 7 and just use the diamond operator.

https://docs.oracle.com/javase/7/docs/technotes/guides/language/type-inference-generic-instance-creation.html

It helps make things more readable and is less typing :D!

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -64,10 +66,11 @@ public DuplicateReporter(MySQLAdaptor adaptor)
 	/**
 	 * Gets the number of referrers for each instance of a duplicated accession.
 	 * @param accession - The accession to look up.
+	 * @param classesToIgnore - A list of class names to ignore, when looking for referrers. Any referrer whose Schema Class is in classes to ignore will not be considered as a referrer, and will not be added to the count.
 	 * @return The DB_IDs of the duplicated accession mapping to the number of referrers of each one.
 	 * @throws Exception 
 	 */
-	public Map<Long, Integer> getReferrerCountForAccession(String accession) throws Exception
+	public Map<Long, Integer> getReferrerCountForAccession(String accession, String ... classesToIgnore) throws Exception
```
jweiser: If you decide to use the `getGOInstancesByAccession` method in DuplicateCleaner, you could re-use it here and do something like this (with the `getGOInstancesByAccession` living whereever is appropriate):
```
public Map<Long, Integer> getReferrerCountForAccession(String accession, String ... classesToIgnore) throws Exception
{
	List<GKInstance> instances = DuplicateCleaner.getGOInstancesByAccession(this.adaptor, accession);

	return instances
		.stream()
		.collect(
			toMap(
				GKInstance::getDBID,
				i -> getReferrerCountforInstance(i, classesToIgnore)
			)
		);
}
```

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -81,23 +84,29 @@ public DuplicateReporter(MySQLAdaptor adaptor)
...
+					Long dbid = i.getDBID();
+					int refCount = getReferrerCountforInstance(i, classesToIgnore);
+					referrerCounts.put(dbid,refCount);
 				}
 			}
 		}
 		return referrerCounts;
 	}
 	
+	public int getReferrerCountforInstance(GKInstance instance, String ... classesToIgnore) throws Exception
```
jweiser: It might be better to make this method private.  Also, if you like the use of the stream in `getReferrerCountForAccession` above it would be best to handle the exception in the body rather than throw it.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
SolomonShorser-OICR: I think it's better where it is, but I'll updated the comments in this are.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+ */
+public class DuplicateCleaner extends ReleaseStep
+{
+	private static final Logger logger = LogManager.getLogger();
+	private MySQLAdaptor adaptor;
+	private static final String[] goClasses = {ReactomeJavaConstants.GO_BiologicalProcess, ReactomeJavaConstants.GO_MolecularFunction, ReactomeJavaConstants.GO_CellularComponent, ReactomeJavaConstants.Compartment};
+
+	public static void main(String[] args)
+	{
+		String pathToResources = "src/main/resources/go-update.properties";
```
cookersjs: Should this be configed?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+		LocalDateTime newestDate = null;
+		for (GKInstance instance : instances)
+		{
+			LocalDateTime localDate = null;
+			GKInstance modifiedInstance = null;
+			modifiedInstance = (GKInstance) instance.getAttributeValue(ReactomeJavaConstants.modified);
+			
+			if (modifiedInstance != null)
+			{
+				localDate = LocalDateTime.parse((CharSequence) modifiedInstance.getAttributeValue(ReactomeJavaConstants.dateTime), formatter);
```
cookersjs: Do we know that the dateTime attribute is standardized in Reactome? What will happen if the formatter finds an unexpected dateTime format?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+			}
+			else
+			{
+				modifiedInstance = (GKInstance) instance.getAttributeValue(ReactomeJavaConstants.created);
```
cookersjs: I think naming the returned 'created' attribute 'modifiedInstance' is confusing, since there is the 'modified' attribute

cookersjs: Nevermind, I see that you are just repurposing the above 'modifiedInstance' variable

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -318,38 +343,33 @@ public void deleteGoInstance(Map<String, Map<String,Object>> goTerms, Map<String
 	
 	private void pointAllReferrersToOtherInstance(GKInstance replacementGOTerm) throws Exception
 	{
-//		Collection<GKInstance> referrers = GoTermInstanceModifier.getReferrersForGoTerm(this.goInstance);
-//		for (GKInstance referrer : referrers)
-//		{
+		@SuppressWarnings("unchecked")
+		Collection<GKSchemaAttribute> attributes = (Collection<GKSchemaAttribute>) this.goInstance.getSchemClass().getReferers();
```
cookersjs: It's not intuitive to me that the referers of the goInstance would be 'attributes'. Doesn't 'getReferers' return 'instances' that refer to the current instance?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
SolomonShorser-OICR: If the user species a path to the properties file on the command line, that is used. This line here sets a default value that only gets used when no argument is passed in.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -318,38 +343,33 @@ public void deleteGoInstance(Map<String, Map<String,Object>> goTerms, Map<String
...
-						referrer.setAttributeValue(attributeName, replacementGOTerm);
-						// update in db.
-						adaptor.updateInstanceAttribute(referrer, attributeName);
-						logger.debug("\"{}\" now refers to {} via {}, instead of referring to \"{}\"", referrer.toString(), replacementGOTerm, attributeName, this.goInstance.toString());
-					}
+					// the referrer could refer to many things via the attribute.
+					// we should ONLY remove *this* GO instance that will probably be deleted
+					// and add the replacement GO term. All other values should be left alone.
+					@SuppressWarnings("unchecked")
+					List<GKInstance> values = (List<GKInstance>) referrer.getAttributeValuesList(attribute);
```
cookersjs: 'values' isn't a Reactome-specific thing, so I think a more descriptive name like 'attributeValues' might be more informative

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -307,6 +328,21 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.IS_OBSOLETE) &&
 		return mainOutput;
 	}
 
+	protected static Map<GKSchemaAttribute, Integer> getReferrerCounts(GKInstance inst) throws Exception
+	{
+		Map<GKSchemaAttribute, Integer> referrersCount = new HashMap<GKSchemaAttribute, Integer>();
+		for (GKSchemaAttribute attrib : (Collection<GKSchemaAttribute>)inst.getSchemClass().getReferers())
```
cookersjs: Same thing I mentioned up above -- does inst.getSchemClass().getReferers() return GKSchemaAttributes?

cookersjs: Nevermind... I can see from the type that it does

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
```diff
@@ -28,7 +30,7 @@
 	@Before
 	public void setup() throws SQLException
 	{
-		adaptor = new MySQLAdaptor("localhost", "test_slice_64_Java_GO_Update", "root", "root", 3306);
+		adaptor = new MySQLAdaptor("localhost", "gk_central_Perl_GO_Update", "root", "root", 3306);
```
cookersjs: Should this be configged?

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
```diff
@@ -90,19 +97,25 @@ public void testGetDuplicateReferrers() throws Exception
 							{
 								for (GKInstance ref : refs)
 								{
-									System.out.println("\t\t"+ref.toString());
+									if (!Arrays.asList(ReactomeJavaConstants.GO_BiologicalProcess, ReactomeJavaConstants.GO_MolecularFunction, ReactomeJavaConstants.GO_CellularComponent, ReactomeJavaConstants.Compartment).contains(ref.getSchemClass().getName()))
+									{
+										System.out.println("\t\tvia "+refAtt.getName()+":\t"+ref.toString());
```
cookersjs: This looks like the end to a statement -- is it a possible continuation from line 85?

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
```diff
@@ -90,19 +97,25 @@ public void testGetDuplicateReferrers() throws Exception
...
 				System.out.print("\n");
 			}
-			else
-			{
-				System.out.println("Accession "+accession+ " has no instances with referrers!");
-			}
+//			else
+//			{
+//				System.out.println("Accession "+accession+ " has no instances with referrers!");
+//			}
```
cookersjs: Can this be removed?

go-update/src/test/java/org/reactome/release/goupdate/GoTermsUpdaterTest.java
```diff
@@ -91,14 +91,14 @@
...
-			"relationship: negatively_regulates GO:0048308 \n" +
-			"\n"
+//			+"[Term]\n"+
+//			"id: GO:0000033\n" + 
+//			"name: duplicate test!\n" + 
+//			"namespace: biological_process\n" + 
+//			"def: \"testing stuff.\n" + 
+//			"is_a: GO:0048309 \n" +
+//			"relationship: negatively_regulates GO:0048308 \n" +
+//			"\n"
```
cookersjs: Can this be removed?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -3,36 +3,39 @@
...
+public class JavaVsPerlComparison
+{
 	public static void main(String[] args) throws Exception
 	{
-		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "gk_central_Java_GO_Update", "root", "root");
-		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "gk_central_Perl_GO_Update", "root", "root");
-		int diffCount = 0;
-		int sameCount = 0;
+		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "gk_central.20181211.dump", "root", "root", 3306);//new MySQLAdaptor("localhost", "gk_central_after_uniprot_update.dump.sql", "root", "root", 3308);
+		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "gk_central.20181211.dump", "root", "root", 3307);//new MySQLAdaptor("localhost", "gk_central_R66_before_chebi_update.sql", "root", "root", 3307);
```
cookersjs: Configged DB names?

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
SolomonShorser-OICR: The `dateTime` column in the `InstanceEdit` table is of the type `timestamp`, so it _should_ be stored in a manner that is independent of formatting. From what I can tell, the Reactome Java API does not have any special handling for temporal data, so I imagine that it simply returns temporal data as a string, which I then format. Also, it seems that MySQL always returns in the format `YYYY-MM-DD HH:MM:SS` (https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html), so I think as long as no one tries to force MySQL to return timestamps in a different format, this should be OK.

cookersjs:
For the classes that are in the actual release repository (DuplicateCleaner, DuplicateReporter, GoTermInstanceModifier, GoTermsUpdater) are they a part of the actual GOUpdate step, or could they be moved to a 'datafix' location if they are one-offs? 

Aside from that and the minor suggestions I had in the code, things look good!

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: In the context `this.goInstance.getSchemClass().getReferers()`, the function `getSchemaClass` will return the data model class for _this_ instance. Then, `getReferrers` is called on the schema class (_not_ on the instance), so it will return the referrer Schema Attributes for the Schema Class. The key thing is: `getReferrers` is not being called on `goInstance`, it is being called on `goInstance`'s Schema Class.

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
SolomonShorser-OICR: Yes, probably. It was a low-priority task for me, since this code never runs in production, only locally, when testing. If we ever get to the point where we run automated tests on some server then this will definitely need to be cleaned up... I'll probably get to this one later.

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
SolomonShorser-OICR: You mean the print statement? Yes, line 85 prints a "header", and then inside the loop, everything for that header gets printed.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
SolomonShorser-OICR: Yes, probably a good idea.

SolomonShorser-OICR:
All classes are used. `DuplicateCleaner`  can be used to clean up duplicates and currently runs as separate from the main GO Update process (though it is a `ReleaseStep` class so it could be called like any other `ReleaseStep`) , and in theory it could also be integrated (probably with not much effort) to _automatically_ clean up duplicates. If the curators would ever be OK with such a thing.

SolomonShorser-OICR:
Is everyone else OK with this getting merged on Monday?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -282,6 +282,22 @@ public static boolean isGoTermDeleteable(GKInstance instance) throws Exception
 	}
 	
 	
+	public void deleteSecondaryGOInstance(GKInstance primaryGOTerm, StringBuffer deletionStringBuffer)
```
jweiser: I think StringBuilder is preferred to StringBuffer now for single threads.

https://stackoverflow.com/questions/355089/difference-between-stringbuilder-and-stringbuffer

SolomonShorser-OICR: I probably used the thread-safe structure in case I figured out a way to parallelize that bit of code, or what was calling it. Or there may have been a block of code that *did* execute on multiple threads and also referenced `GoTermsUpdater.deletionStringBuilder`, which is why it is a `StringBuffer` and not a `StringBuilder`.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+			}
+			// update list of things to delete with IDs that have no referrers.
+			dbIDsToDelete.addAll(dbIDsWithNoReferrers);
+			// If no DB IDs have significant referrers, we should delete ALL DB_IDs for the accession, except for the newest one.
+			if (dbIDsWithNoReferrers.size() == refCounts.keySet().size())
+			{
+				GKInstance newestInstance = null;
+				// Need to try all classes since we don't know which class a GO Accession might be. Note: this.goClasses also contains "Compartment", but here "Compartment" is not used.
+				// It might be safe to use this.goClasses here, but since this bit of code did not use Compartment, we are not going to use this.goClasses here, for now.
+				for (String reactomeClass : Arrays.asList(ReactomeJavaConstants.GO_BiologicalProcess, ReactomeJavaConstants.GO_MolecularFunction, ReactomeJavaConstants.GO_CellularComponent))
```
jweiser: This list is used in many places in the code base.  It would consolidate and make the code more succinct to pull it into the GoUpdateConstants like:

```
static final List<String> GO_CLASSES = Arrays.asList(
    ReactomeJavaConstants.GO_BiologicalProcess,
    ReactomeJavaConstants.GO_MolecularFunction,
    ReactomeJavaConstants.GO_CellularComponent
);
```

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+					}
+				}
+				logger.debug("Newest instance for accession {} is {} - this instance will NOT be deleted.", accession, newestInstance.toString());
+				// REMOVE the ID of the newest instance from the list of things to delete (it would have been automatically added if the "newest" instance
+				// had 0 referrers).
+				dbIDsToDelete.remove(newestInstance.getDBID());
+			}
+		}
+		logger.info("\n\nSummary:\nTotal number of duplicated accessions: {} \n"
+				+ "Number of instances with significant (non-GO Term) referrers: {}\n", countOfDuplicatesAccessions, instancesWithSignificantReferrers);
```
jweiser: The `countOfDuplicatesAccessions` variable can be removed and replaced directly with the query `duplicates.keySet().size()`.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -81,23 +84,29 @@ public DuplicateReporter(MySQLAdaptor adaptor)
...
-						if (referrerCounts.containsKey(i.getDBID()))
-						{
-							referrerCounts.put(i.getDBID(), referrerCounts.get(i.getDBID()) + referrers.size());
-						}
-						else
-						{
-							referrerCounts.put(i.getDBID(), referrers.size());
-						}
-					}
+					Long dbid = i.getDBID();
```
jweiser: Better to use primitives whenever possible

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
jweiser: Ah okay.  What do you think about using the stream in the comment for `getReferrerCountForAccession`?  If you like it, then the lambda would need to handle the exception thrown here or this method would need to handle the exception rather than throw it.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -81,23 +84,29 @@ public DuplicateReporter(MySQLAdaptor adaptor)
...
+	{
+		int refCount = 0;
+		
+		for (GKSchemaAttribute referrerAttrib : (Set<GKSchemaAttribute>) instance.getSchemClass().getReferers())
+		{
+			@SuppressWarnings("unchecked")
+			Collection<GKInstance> referrers = (Collection<GKInstance>) instance.getReferers(referrerAttrib);
+			if (classesToIgnore != null && classesToIgnore.length > 0)
+			{
+				referrers = referrers.stream().filter(r -> !(new HashSet<String>( Arrays.asList(classesToIgnore))).contains(r.getSchemClass().getName())).collect(Collectors.toList());
```
jweiser: It took me some time to understand this was checking the referrer class was not in the list of classes to ignore.

`!(new HashSet<String>( Arrays.asList(classesToIgnore))).contains(r.getSchemClass().getName())`

Could it be refactored to be more declarative?  Also, I think you can call contains directly on the List instead of wrapping in a HashSet.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
SolomonShorser-OICR: I'm not sure, but it might be possible (or it _was_ possible, at one time) for `duplicates` to be null, in which case `duplicates.keySet()...` would throw a `NullPointerException`.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
SolomonShorser-OICR: I'll have a look at this

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
```diff
@@ -0,0 +1,215 @@
...
+		Map<String, Integer> duplicates = dupeReporter.getDuplicateAccessions();
+		countOfDuplicatesAccessions = duplicates.keySet().size();
+		// loop through all duplicated accessions.
+		for (String accession : duplicates.keySet())
+		{
+			logger.info("Accession {} is duplicated {} times.", accession, duplicates.get(accession));
+			List<Long> dbIDsWithNoReferrers = new ArrayList<Long>();
+			Map <Long,Integer> refCounts = dupeReporter.getReferrerCountForAccession(accession, goClasses);
+			// if there are referrers... refCounts should never be null, unless something catastrophic happens inside getReferrerCountForAccession and in that case, there will probably be a thrown exception.
+			if (refCounts.size() > 0)
```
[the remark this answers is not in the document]
SolomonShorser-OICR: I'll have a closer look at this.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -292,18 +308,27 @@ public void deleteGoInstance(Map<String, Map<String,Object>> goTerms, Map<String
...
 				{
 					GKInstance replacementGOTerm = allGoInstances.get(replacementGOTermAccession).get(0);
-					pointAllReferrersToOtherInstance(replacementGOTerm);
+					this.pointAllReferrersToOtherInstance(replacementGOTerm);
 				}
 				deletionStringBuilder.append("Deleting GO instance: \"").append(this.goInstance.toString()).append("\" (GO:").append(goId).append(")\n");
 				adaptor.deleteInstance(this.goInstance);
 			}
+			// A GO term that has no replacement value can still be deleted if it has no referrers.
+			else if (goTerms.get(goId).get(GoUpdateConstants.REPLACED_BY) == null && GoTermsUpdater.getReferrerCounts(this.goInstance).isEmpty())
```
jweiser: The first condition here isn't needed since it would have to be null to fail the if check and go to this else if.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -318,38 +343,33 @@ public void deleteGoInstance(Map<String, Map<String,Object>> goTerms, Map<String
...
+					// the referrer could refer to many things via the attribute.
+					// we should ONLY remove *this* GO instance that will probably be deleted
+					// and add the replacement GO term. All other values should be left alone.
+					@SuppressWarnings("unchecked")
+					List<GKInstance> referrerAttributeValues = (List<GKInstance>) referrer.getAttributeValuesList(attribute);
+					// remove *this* goInstance from the referrer
+					referrerAttributeValues = referrerAttributeValues.parallelStream().filter(v -> !v.getDBID().equals(this.goInstance.getDBID())).collect(Collectors.toList());
+					// add the replacement to the referrer
+					referrerAttributeValues.add(replacementGOTerm);
+					referrer.setAttributeValue(attributeName, replacementGOTerm);
```
jweiser: `referrerAttributeValues` is populated but never used.  Should this line be `referrer.setAttributeValue(attributeName, referrerAttributeValues)`?  Otherwise, it looks like the attribute value is getting overwritten with the replacementGOTerm so other possible attribute values are getting wiped out.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -193,22 +192,53 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.PENDING_OBSOLET
...
-					obsoleteAccessionLogger.warn("GO:{} ({}) marked as OBSOLETE!{}",goID, goInstances.toString(), replacementValue);
+					StringBuilder attemptToDeleteObsoleteMessage = new StringBuilder();
+					Map<GKSchemaAttribute, Integer> referrersCount = new HashMap<>();
 					// Only add instance(s) to deletion list if they have a valid replacement.
 					if (goTermsFromFile.get(goID).get(GoUpdateConstants.REPLACED_BY) != null)
 					{
 						for (GKInstance inst : goInstances)
 						{
 							instancesForDeletion.add(inst);
 						}
```
jweiser: Could the for loop be done as: `instancesForDeletion.addAll(goInstances)`?

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -193,22 +192,53 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.PENDING_OBSOLET
...
+					else
+					{
+						// ...or, if an obsolete term has no replacement AND also has no referreres, it can be 
+						// safely be deleted because nothing will be affected.
+						// 
+						// (Check that the instance has not already been added to instancesForDeletion by some other path)
+						goInstances.stream().filter(inst -> !instancesForDeletion.contains(inst)).forEach( inst -> {
+							try
+							{
+								referrersCount.putAll( this.getReferrerCounts(inst) );
```
jweiser: Can `this` be removed since `getReferrerCounts` is a static method?

go-update/src/test/java/org/reactome/release/goupdate/DuplicateReporterIT.java
```diff
@@ -67,18 +72,20 @@ public void testGetDuplicateReferrers() throws Exception
 		for (String accession : duplicates.keySet())
 		{
 			assertNotNull(duplicates.get(accession));
+			totalDuplicatesCount = duplicates.keySet().size();
```
jweiser: Why does totalDuplicatesCount get reassigned in each loop iteration?  When the value is printed at the end, it will only have the value from the last cycle of the loop.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -51,138 +54,173 @@ public static void main(String[] args) throws Exception
...
+							@SuppressWarnings("unchecked")
+							List<GKInstance> perlList = ((List<GKInstance>) perlInst.getReferers(ReactomeJavaConstants.goBiologicalProcess));
+							if (javaList != null && perlList != null)
 							{
-								peSameCount ++;
+								javaList = javaList.stream().sorted(dbIdComparator).collect(Collectors.toList());
+								perlList = perlList.stream().sorted(dbIdComparator).collect(Collectors.toList());
+	
+								GKInstance javaReferringRegulation =  javaList.get(0);
+								GKInstance perlReferringRegulation = perlList.get(0);
```
jweiser: The referrers actually could be a regulation or an event (pathway or reaction like event).  Usually it is an event, so it may be clearer to change the name.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateCleaner.java
SolomonShorser-OICR: On closer inspection, this pattern (iterating over these 3 classes) only happens in two places - here and in `DuplicateReporter` (1 occurrence). `DuplicateCleaner` has `goClasses`, but as per the note, that also contains `Compartment`. In this particular case, `Compartment` is not used. So, I think it's OK to leave it as-is. `goClasses` takes care of the places where a longer list was used repeatedly.

go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
SolomonShorser-OICR: `getDBID` returns a `Long` not a `long`.

jweiser: Yeah, I saw that.  I tried out `long dbid = i.getDBID()` and it will auto unbox for you.  It's just better to have variables that use the value stored in a primitive so there's never concern about value vs reference equality issues.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: Yes, I think so. Good catch! :+1:

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
SolomonShorser-OICR: I think that loop used to do more than just adding to another list.

SolomonShorser-OICR:
@jweiser: Many changes made. Please let me know if there's anything else.

PR #23, opened 2019-01-29
Feature/orthopairs

cookersjs:
New script for Orthopairs. It downloads files from PANTHER as well as some ID files from various Model Organism Databases. This takes us off Biomart and Compara, and reduces the runtime quite a bit!


orthopairs/README.md
```diff
@@ -0,0 +1,54 @@
+<h2> Orthopairs: Protein ortholog file generation</h2>
+
+To see how to run this script, check out the <a href="https://github.com/reactome/data-release-pipeline/new/feature/orthopairs/orthopairs#-running-orthopairs-">Running Orthopairs</a> section.
```
SolomonShorser-OICR: Nitpick: Java applications are not scripts.

orthopairs/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+    <packaging>jar</packaging>
+
+  <name>orthopairs</name>
+  <url>http://maven.apache.org</url>
```
SolomonShorser-OICR: This is not the URL for the project. Either remove the url element or put the github URL for data-release-pipeline in here.

orthopairs/runOrthopairs.sh
```diff
@@ -0,0 +1,9 @@
+## Update repo
+git pull
+## Create new jar file with orthoinference code
+mvn clean compile assembly:single
+
+## Run Orthopairs file generating script
+echo "java -jar target/orthopairs-0.0.1-SNAPSHOT-jar-with-dependencies.jar"
```
SolomonShorser-OICR: you can echo commands in bash with `set -x / set +x`.

orthopairs/runOrthopairs.sh
```diff
@@ -0,0 +1,9 @@
+## Update repo
+git pull
```
SolomonShorser-OICR: This should probably be a checkout of a specific version tag.

orthopairs/src/main/java/org/reactome/release/orthopairs/AlternateIdMapper.java
```diff
@@ -0,0 +1,169 @@
...
+            altIdToEnsemblMap = mapYeastAlternateIds(br);
+        } else {
+            logger.warn(speciesKey + " does not have a method for mapping its alternate Ids to Ensembl Ids");
+        }
+
+        return altIdToEnsemblMap;
+    }
+
+    // This uses HGNC_homologene.rpt from http://www.informatics.jax.org/downloads/reports/
+    private static Map<String, Set<String>> mapMouseAlternateIds(BufferedReader br) throws IOException {
```
SolomonShorser-OICR: If you feel like it, you could create an common interface for doing these "alternate" mappings (which should not be too hard since all of these methods take the same input and return the same output), then create a few classes that implement it and contain only a single function for one species-specific mapping. Then, create instances of these special mappers and call their "map" function. Some might say it's cleaner than having a bunch of functions for different species all in one class, but I don't think it's necessary.

orthopairs/src/main/java/org/reactome/release/orthopairs/AlternateIdMapper.java
```diff
@@ -0,0 +1,169 @@
...
+public class AlternateIdMapper {
+    private static final Logger logger = LogManager.getLogger();
+    // Since we require some species-specific databases, there is a method for mapping the returned files. This wasn't
+    // abstractble, considering that the files all have different formats.
+    public static Map<String, Set<String>> getAltIdMappingFile(Object speciesKey, String alternateIdFilename) throws IOException {
+
+        File alternateIdFile = new File(alternateIdFilename);
+        BufferedReader br = new BufferedReader(new FileReader(alternateIdFile));
+        Map<String,Set<String>> altIdToEnsemblMap = new HashMap<>();
+        if (speciesKey.equals("mmus")) {
```
SolomonShorser-OICR: I think Java's switch/case syntax has supported switching on strings for a while. Personally, I find the switch/case a more appropriate choice for this sort of code.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthologyFileParser.java
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                entityMap.get(targetSpecies).get(keyEntity).add(valueEntity);
+            }
+        }
+
+        // Lines with an orthologType equal to 'LDO' mean that we only want that value in the Set since its the Least Diverged Ortholog, meaning we have a
+        // high degree of confidence in its homology. We remove all other values from the Set unless an 'LDO' value already exists. In this rare case,
+        // we will keep multiple LDOs.
+        Set<String> targetEntitys = entityMap.get(targetSpecies).get(keyEntity);
+        if (orthologType.equals("LDO")) {
```
SolomonShorser-OICR: "LDO" seems to appear often enough that it could be replaced with a constant variable whose value is "LDO"

orthopairs/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+`Mouse Genome Informatics (MGI)` (<a href="http://www.informatics.jax.org/">link</a>) -- Mapping file: http://www.informatics.jax.org/downloads/reports/HGNC_homologene.rpt
+
+`Rat Genome Database (RGD)` (<a href="https://rgd.mcw.edu/">link</a>) -- Mapping file: ftp://ftp.rgd.mcw.edu/pub/data_release/GENES_RAT.txt
+
+`Xenbase (Frog)` (<a href="http://www.xenbase.org/entry/">link</a>) -- Mapping file: ftp://ftp.xenbase.org/pub/GenePageReports/GenePageEnsemblModelMapping.txt
+
+`Zebrafish Information Network (ZFIN)` (<a href="https://zfin.org/">link</a>) Mapping file: https://zfin.org/downloads/ensembl_1_to_1.txt
+
+`Saccharomyces Genome Database (SGD)` (<a href="https://www.yeastgenome.org/">link</a>) Mapping file downloaded from <a href="https://yeastmine.yeastgenome.org/yeastmine/bagDetails.do?scope=all&bagName=ALL_Verified_Uncharacterized_Dubious_ORFs">here</a>. File not programmatically accessible, so mapping information is stored in`src/main/resources/sgd_ids.txt`
```
SolomonShorser-OICR: If it's not programmatically, are we allowed to post it to github?

SolomonShorser-OICR: Also, maybe I missed it, but it's not clear to me where this file gets loaded.

orthopairs/src/main/java/org/reactome/release/orthopairs/AlternateIdMapper.java
cookersjs: I wasn't aware of switch/case, I'll look into it

orthopairs/README.md
cookersjs: program?

orthopairs/pom.xml
cookersjs: Ya it gets auto-generated like that I believe, I'll make the fix

orthopairs/README.md
cookersjs: What I mean by this statement is that there isn't a 'download link' but rather a GUI interface that masks the download link

SolomonShorser-OICR: Oh, ok. Sometimes it's tricky... I know that for BRENDA, you have to click on some button that says you Accept the License before you are allowed to download, which means that you can't download programmatically (well, more like you _shouldn't_), and also, I imagine, redistributing is not allowed. If there's no license issues around that file, then posting it here is probably no issue.

orthopairs/src/main/java/org/reactome/release/orthopairs/AlternateIdMapper.java
jweiser: There's a classic refactoring pattern called 'replace conditional with polymorphism' that would be perfect here.  

https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html

A family of small, simple classes may be better than a switch statement with an abstract class or interface `AlternateIDMapper` and subclasses for each species that inherit from it.  You could do the mapping assignment in one line and have the map implementations more self-contained for each species.

jweiser: Yes, you're right.  I was just to show the idea.  I've usually seen this go with a factory class that is responsible for choosing, creating, and returning the correct class based on the species being processed.

https://www.tutorialspoint.com/design_pattern/factory_pattern.htm

SolomonShorser-OICR: Yes, that would move the logic out of the AlternateIdMapper and into a factory.

jweiser: Yup, but it would allow the methods that are specific to each species processing of alternate ids to have its own class.  It is good to keep the amount of responsibility specific and focused for any one class.

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+     */
+
+    private static final Logger logger = LogManager.getLogger();
+
+    public static void main( String[] args ) throws IOException, ParseException {
+
+        String pathToConfig = "";
+        String sourceMappingSpecies = "";
+        // If using an alternative source species, specify the 4-letter code as the second argument
+        //TODO: Better solution for this. What if user wants default config but to change source species?
```
jweiser: You could use something like the library JCommander for named command line arguments and I think you can specify defaults.

https://stackoverflow.com/a/3337023

SolomonShorser-OICR: Good point. Specifying the  path to config and an alternate source species should not be dependent on each other.

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+        String pathToSpeciesConfig = props.get("pathToSpeciesConfig").toString();
+        String pantherFilepath = props.get("pantherFileURL").toString();
+        String pantherQfOFilename = props.get("pantherQfOFilename").toString();
+        String pantherHCOPFilename = props.get("pantherHCOPFilename").toString();
+        String MGIFileURL = props.get("MGIFileURL").toString();
+        String RGDFileURL = props.get("RGDFileURL").toString();
+        String XenbaseFileURL = props.get("XenbaseFileURL").toString();
+        String ZFINFileURL = props.get("ZFINFileURL").toString();
+
+        new File(releaseNumber).mkdir();
```
jweiser: I think there is a new way to create directories for Java 7 and up that is preferred like this:

`Files.createDirectories(Paths.get(releaseNumber));`

SolomonShorser-OICR: https://docs.oracle.com/javase/tutorial/essential/io/dirs.html

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+
+        logger.info("Starting Orthopairs file generation");
+        // Download and extract homology files from Panther
+        List<String> pantherFiles = new ArrayList<String>(Arrays.asList(pantherQfOFilename, pantherHCOPFilename));
```
jweiser: You can use the diamond operator when instantiating a parameterized object.

`List<String> pantherFiles = new ArrayList<>(...);`

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthologyFileParser.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.logging.log4j.Logger;
+import org.json.simple.JSONObject;
+
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.*;
+
+public class OrthologyFileParser {
```
jweiser: It might be better to make this class use non-static fields and methods since the mapping data processed is being stored.  It makes more sense for that to belong to an instance of the class rather than the class itself.

You could add a constructor something like this:

```
public OrthologyFileParser(List<String> pantherFiles, String sourceMappingSpecies, JSONObject speciesJSONFile) {
    try {
        parsePantherOrthologFiles(pantherFiles, sourceMappingSpecies, speciesJSONFile);
    } catch (IOException e) {
        logger.fatal("Could not parse Panther Ortholog files", e);
	throw new RuntimeException(e);
    }
}
```

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
jweiser: Yup, java.nio.file is the package the 'Files' and 'Paths' classes comes from.

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
cookersjs: Yea the named arguments seemed like the way to go, just didn't know how to do them, thanks!

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
cookersjs: Yea I typically do that now, that one just got through

jweiser: I think IntelliJ will color it grey if it finds the type on the right side and let you know it can be removed.

You could try it out by pasting in something like this to see if it works.

`List<String> a = new ArrayList<String>();`

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
jweiser: There's actually seemingly endless ways to do it in Java!  

https://stackoverflow.com/a/7829772

JCommander looks like a good option, though.

jweiser: https://picocli.info/ might be a good option, too.  It has autocompletion and other neat features.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthologyFileParser.java
```diff
@@ -0,0 +1,119 @@
...
+        for (Object speciesKey : speciesJSONFile.keySet()) {
+            JSONObject speciesJSON = (JSONObject) speciesJSONFile.get(speciesKey);
+            if (!speciesKey.equals(sourceMappingSpecies)) {
+                pantherSpeciesNames.add(speciesJSON.get("panther_name").toString());
+            } else {
+                sourceSpeciesPantherName = speciesJSON.get("panther_name").toString();
+            }
+        }
+
+        if (!sourceSpeciesPantherName.equals("")) {
```
jweiser: You can use `isEmpty()` on a String in place of `.equals("")`.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthologyFileParser.java
```diff
@@ -0,0 +1,119 @@
...
+            // There are 2 Panther files at time of writing since Sus Scrofa (PIG) info is found in a seperate file.
+            // The program iterates through both since for overlapping species it produces the same data structure. This is where the redundancy reduction of Sets comes in handy.
+            for (String pantherFileTar : pantherFiles) {
+                String extractedPantherFile = pantherFileTar.replace(".tar.gz", "");
```
jweiser: Since the pantherFile is already extracted when this executes, it may be better to pass in the `pantherFiles` list to this method already stripped of the .tar.gz extension.

jweiser: In Main.java, where you call `downloadAndExtractTarFile` you could move the loop into a method called `downloadAndExtractTarFiles(pantherFiles)` and have it return the new list of file names.

```
List<String> pantherFiles = downloadAndExtractTarFiles(Arrays.asList(pantherQfOFilename, pantherHCOPFilename));
```

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+        // HGNC identifier file is downloaded as well.
+        List<String> alternativeIdMappingURLs = new ArrayList<>(Arrays.asList(MGIFileURL,RGDFileURL,XenbaseFileURL,ZFINFileURL));
+        for (String altIdURL : alternativeIdMappingURLs) {
+            File altIdFilepath = new File(altIdURL.substring(altIdURL.lastIndexOf("/")+1));
+            if (!altIdFilepath.exists()) {
+                logger.info("Downloading " + altIdURL);
+                FileUtils.copyURLToFile(new URL(altIdURL), altIdFilepath);
+            } else {
+                logger.info(altIdFilepath + " already exists");
+            }
```
jweiser: The code to download files above is used in two places, so it could be extracted to a method.

```
private static void downloadFile(URL resourceToDownload, File downloadTargetFile) throws IOException {
	// Download files
	if (!downloadTargetFile.exists()) {
		logger.info("Downloading " + resourceToDownload);
		FileUtils.copyURLToFile(resourceToDownload, downloadTargetFile);
	} else {
		logger.info(downloadTargetFile + " already exists");
	}
}
```

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+        List<String> pantherFiles = new ArrayList<String>(Arrays.asList(pantherQfOFilename, pantherHCOPFilename));
+        for (String pantherFilename : pantherFiles) {
+            downloadAndExtractTarFile(pantherFilename, pantherFilepath);
+        }
+
+        // Download ID files from various model organism databases (Mouse Genome Informatics, Rat Genome Database, Xenbase (frog), ZFIN (Zebrafish))
+        // HGNC identifier file is downloaded as well.
+        List<String> alternativeIdMappingURLs = new ArrayList<>(Arrays.asList(MGIFileURL,RGDFileURL,XenbaseFileURL,ZFINFileURL));
+        for (String altIdURL : alternativeIdMappingURLs) {
+            File altIdFilepath = new File(altIdURL.substring(altIdURL.lastIndexOf("/")+1));
```
jweiser: Something like this might be more declarative: `Paths.get(altIdURL).getFileName().toFile())`.

You may need to use `new URI(altIdURL)` for the URL to be handled properly, but the Paths.get() method may be smart enough to use the URL as a String.  We'd need to try it out.

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        logger.info("Finished Orthopairs file generation");
+    }
+
+    private static void downloadAndExtractTarFile(String pantherFilename, String pantherFilepath) throws IOException {
+
+        URL pantherFileURL = new URL(pantherFilepath + pantherFilename);
+        File pantherTarFile = new File(pantherFilename);
+
+        // Download files
```
jweiser: This is the other spot where a download method could be called if extracted out.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+            Set<String> targetProteinIds = speciesProteinHomologs.get(sourceProteinId);
+            targetProteinIds.remove("LDO");
+            String proteinOrthologLine = sourceProteinId.split("=")[1] + "\t";
+            List<String> cleanTargetProteinIds = new ArrayList<>();
+            for (String targetProteinId : targetProteinIds) {
+                cleanTargetProteinIds.add(targetProteinId.split("=")[1]);
+            }
+            Collections.sort(cleanTargetProteinIds);
+            proteinOrthologLine += String.join(" ", cleanTargetProteinIds) + "\n";
+            Files.write(Paths.get(sourceTargetProteinMappingFile.getPath()), proteinOrthologLine.getBytes(), StandardOpenOption.APPEND);
```
jweiser: The argument to Paths.get() can be replaced with `sourceTargetProteinMappingFilename` so there is no need to create a File object.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // Create target species gene-protein mapping file
+    public static void createSpeciesGeneProteinFile(String speciesKey, String targetGeneProteinMappingFilename, JSONObject speciesJSON, Map<String,Set<String>> speciesGeneProteinMap) throws IOException {
+
+        logger.info("\tGenerating " + targetGeneProteinMappingFilename);
+        File targetGeneProteinMappingFile = new File (targetGeneProteinMappingFilename);
+        if (targetGeneProteinMappingFile.exists()) {
+            targetGeneProteinMappingFile.delete();
+        }
```
jweiser: Similar to above, the Files.write() method will look after this

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                } else {
+                    String geneProteinLine = geneId + "\t" + String.join(" ", cleanTargetProteinIds) + "\n";
+                    targetGeneProteinLines.add(geneProteinLine);
+                }
+            }
+        }
+        Collections.sort(targetGeneProteinLines);
+        for (String targetGeneProteinLine : targetGeneProteinLines) {
+            Files.write(Paths.get(targetGeneProteinMappingFile.getPath()), targetGeneProteinLine.getBytes(), StandardOpenOption.APPEND);
```
jweiser: Like above, the `targetGeneProteinMappingFile` String can be used for the argument to Paths.get()

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.*;
+
+public class OrthopairFileGenerator {
+    private static final Logger logger = LogManager.getLogger();
+    // Create source-target protein mapping file
+    public static void createProteinHomologyFile(String speciesKey, String sourceTargetProteinMappingFilename, JSONObject speciesJSON, Map<String,Set<String>> speciesProteinHomologs ) throws IOException {
```
jweiser: Can speciesKey and speciesJSON be removed from the parameters?  It looks like they aren't used

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+
+        logger.info("\tGenerating " + sourceTargetProteinMappingFilename);
+        File sourceTargetProteinMappingFile = new File(sourceTargetProteinMappingFilename);
+        if (sourceTargetProteinMappingFile.exists()) {
+            sourceTargetProteinMappingFile.delete();
+        }
+        sourceTargetProteinMappingFile.createNewFile();
+
+        List<String> sourceProteinIds = new ArrayList<>();
+        sourceProteinIds.addAll(speciesProteinHomologs.keySet());
```
jweiser: The call to addAll can be replaced by initializing the ArrayList:

`List<String> sourceProteinIds = new ArrayList<>(speciesProteinHomologs.keySet());`

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+        Collections.sort(sourceProteinIds);
+        for (String sourceProteinId : sourceProteinIds) {
+            Set<String> targetProteinIds = speciesProteinHomologs.get(sourceProteinId);
+            targetProteinIds.remove("LDO");
+            String proteinOrthologLine = sourceProteinId.split("=")[1] + "\t";
+            List<String> cleanTargetProteinIds = new ArrayList<>();
+            for (String targetProteinId : targetProteinIds) {
+                cleanTargetProteinIds.add(targetProteinId.split("=")[1]);
+            }
+            Collections.sort(cleanTargetProteinIds);
```
jweiser: For transforming operations on collections, it can be more declarative to use streams (though not always) to show what's being done to the data.  The book Effective Java recommends that it's usually best to consider both ways and decide which way is clearer.  With streams, this could be done this way:

```
String targetProteinIds = speciesProteinHomologs.get(sourceProteinId)
	.stream()
	.filter(targetProteinId -> !targetProteinId.equals("LDO"))
	.map(targetProteinId -> targetProteinId.split("=")[1])
	.sorted()
	.collect(Collectors.joining(" "));
```

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        sourceTargetProteinMappingFile.createNewFile();
+
+        List<String> sourceProteinIds = new ArrayList<>();
+        sourceProteinIds.addAll(speciesProteinHomologs.keySet());
+        Collections.sort(sourceProteinIds);
+        for (String sourceProteinId : sourceProteinIds) {
+            Set<String> targetProteinIds = speciesProteinHomologs.get(sourceProteinId);
+            targetProteinIds.remove("LDO");
+            String proteinOrthologLine = sourceProteinId.split("=")[1] + "\t";
```
jweiser: The pattern `String.split("=")[1]` is done a few times in the class.  Maybe it could be a method with a name describing what the value after the equals sign is for explanation and consolidation to one place?

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
jweiser: Actually, this is done in both methods that create files, so a private method might be better:

```
private static String getTargetProteinsAsString(Set<String> targetProteins) {
	return targetProteins
		.stream()
		.filter(targetProteinId -> !targetProteinId.equals("LDO"))
		.map(targetProteinId -> targetProteinId.split("=")[1])
		.sorted()
		.collect(Collectors.joining(" "));
}
```

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+    // Create target species gene-protein mapping file
+    public static void createSpeciesGeneProteinFile(String speciesKey, String targetGeneProteinMappingFilename, JSONObject speciesJSON, Map<String,Set<String>> speciesGeneProteinMap) throws IOException {
+
+        logger.info("\tGenerating " + targetGeneProteinMappingFilename);
+        File targetGeneProteinMappingFile = new File (targetGeneProteinMappingFilename);
+        if (targetGeneProteinMappingFile.exists()) {
+            targetGeneProteinMappingFile.delete();
+        }
+
+        boolean altIdMappingExists = false;
```
jweiser: A method of the same name to query might be preferable to a boolean temp so there's no need to track the variable state.

```
private static boolean altIdMappingExists(JSONObject speciesJSON) {
	return speciesJSON.get("alt_id_file") != null;
}
```

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+            String geneId = geneSplit[geneSplit.length - 1];
+            Set<String> targetProteinIds = speciesGeneProteinMap.get(targetGeneId);
+            targetProteinIds.remove("LDO");
+            List<String> cleanTargetProteinIds = new ArrayList<>();
+            for (String targetProteinId : speciesGeneProteinMap.get(targetGeneId)) {
+                cleanTargetProteinIds.add(targetProteinId.split("=")[1]);
+            }
+            if (cleanTargetProteinIds.size() > 0) {
+                Collections.sort(cleanTargetProteinIds);
+                if (!geneSource.startsWith("Ensembl") && altIdMappingExists) {
```
jweiser: The `altIdMappingExists(speciesJSON)` could also be used here.

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+            altIdToEnsemblMap = AlternateIdMapper.getAltIdMappingFile(speciesKey, speciesJSON.get("alt_id_file").toString());
+            altIdMappingExists = true;
+        }
+        targetGeneProteinMappingFile.createNewFile();
+
+        List<String> targetGeneProteinLines = new ArrayList<>();
+        for (String targetGeneId : speciesGeneProteinMap.keySet()) {
+            String[] geneSplit = targetGeneId.split("=");
+            String geneSource = geneSplit[0];
+            String geneId = geneSplit[geneSplit.length - 1];
```
jweiser: From this spot, the body of the for loop could be done like this to make the code more declarative:

```
String targetProteinIds = getTargetProteinsAsString(speciesGeneProteinMap.get(targetGeneId));
if (!targetProteinIds.isEmpty()) {
	if (!geneSource.startsWith("Ensembl") && altIdMappingExists(speciesJSON)) {
		targetGeneProteinLines.addAll(
			createGeneProteinLines(altIdToEnsemblMap.get(geneId), targetProteinIds)
		);
	} else {
		targetGeneProteinLines.add(createGeneProteinLine(geneId, targetProteinIds));
	}
}
```
With the private methods:
```
private static Set<String> createGeneProteinLines(Set<String> geneIds, String targetProteinIds) {
	return geneIds
		.stream()
		.map(geneId -> createGeneProteinLine(geneId, targetProteinIds))
		.collect(Collectors.toSet());
}

private static String createGeneProteinLine(String geneId, String targetProteinIds) {
	return geneId + "\t" + targetProteinIds + "\n";
}
```

orthopairs/runOrthopairs.sh
jweiser: Adding these lines like we did for orthoinference would be good so wherever the current working directory is, the script always knows its directory context for execution.
```
#!/bin/bash
cd $DIR
```

orthopairs/src/test/java/org/reactome/release/orthopairs/OrthopairFileGeneratorTest.java
```diff
@@ -0,0 +1,81 @@
...
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void createProteinHomologyFileTest() throws IOException {
+        String sourceTargetProteinMappingFilename = "sourceTargetProteinMappingFilename";
+        mockSet.add("UniProtKB=Q6DEX5");
+        mockSet.add("LDO");
+        mockMap.put("UniProtKB=Q86YI8", mockSet);
+        OrthopairFileGenerator.createProteinHomologyFile("speciesKey", sourceTargetProteinMappingFilename, mockJSONObject, mockMap);
```
jweiser: Does the file have contents after this line is run?  I'm still getting my head around mocking.  If so, it might be good to assert the file contains what you expect?

jweiser: I think JUnit has an assertion plugin for checking files easily

orthopairs/src/test/java/org/reactome/release/orthopairs/AlternateIdMapperTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void mapMouseAlternateIdsTest() throws IOException {
+
+        AlternateIdMapper.getAltIdMappingFile("mmus", mouseTestFilename);
```
jweiser: It'd be helpful to test the contents of the maps retrieved from each test in addition to the success of retrieving a file.

orthopairs/src/test/java/org/reactome/release/orthopairs/OrthopairFileGeneratorTest.java
```diff
@@ -30,10 +30,10 @@
...
     JSONObject mockJSONObject;
 
-    @Before
-    public void setUp() {
-        MockitoAnnotations.initMocks(this);
-    }
+//    @Before
+//    public void setUp() {
+//        MockitoAnnotations.initMocks(this);
+//    }
```
jweiser: Why is the mock init commented out?

orthopairs/README.md
```diff
@@ -0,0 +1,54 @@
+<h2> Orthopairs: Protein ortholog file generation</h2>
+
+To see how to run this script, check out the <a href="https://github.com/reactome/data-release-pipeline/new/feature/orthopairs/orthopairs#-running-orthopairs-">Running Orthopairs</a> section.
+
+This step has been rewritten in Java and completely overhauled from the Perl <a href="https://github.com/reactome/Release/tree/master/scripts/release/orthopairs">version</a>.
```
SolomonShorser-OICR: Markdown tip: You don't need to use HTML anchor tags, you can use markdown notation. See: https://guides.github.com/features/mastering-markdown/#syntax - look for "Links".

orthopairs/src/test/java/org/reactome/release/orthopairs/OrthopairFileGeneratorTest.java
cookersjs: Its not used in the current tests. It was causing an error when I would run all tests, but when I ran the test individual it was fine

orthopairs/src/main/resources/config.properties
```diff
@@ -1,4 +1,4 @@
-releaseNumber=
+releaseNumber=90
```
jweiser: Just wondering if releaseNumber should be set in the checked-in file and if it should be set to 90?

cookersjs: No, it shouldn't be there

orthopairs/src/test/java/org/reactome/release/orthopairs/OrthologyFileParserTest.java
```diff
@@ -51,22 +52,11 @@ public void parseJSONObjectsTest() throws IOException {
...
-//        Mockito.when(mockJSONObject.get(mockObject)).thenReturn(mockJSONObject);
-//        Mockito.when(mockJSONObject.get("panther_name")).thenReturn("HUMAN");
-////        Mockito.when(mockStringSet.contains("DICDI")).thenReturn(true);
-//
-//        OrthologyFileParser.parsePantherOrthologFiles(mockList, "hsap", mockJSONObject);
-//    }
+    @Test
+    public void orthologyFileParserTest() throws Exception {
+        JSONObject speciesJSONFile = (JSONObject) parser.parse(new FileReader(pathToSpeciesConfig));
+        OrthologyFileParser.parsePantherOrthologFiles(mockList, "hsap", speciesJSONFile);
```
jweiser: Can you test the contents of the maps that the parsing populates?

orthopairs/runOrthopairs.sh
SolomonShorser-OICR: ...also, the version number should be parameterized.

orthopairs/src/test/java/org/reactome/release/orthopairs/OrthopairFileGeneratorTest.java
jweiser: Oh, I see - it's just using the collections that aren't mocked.  Got it, thanks!

orthopairs/src/main/java/org/reactome/release/orthopairs/OrthopairFileGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final Logger logger = LogManager.getLogger();
+    // Create source-target protein mapping file
+    public static void createProteinHomologyFile(String speciesKey, String sourceTargetProteinMappingFilename, JSONObject speciesJSON, Map<String,Set<String>> speciesProteinHomologs ) throws IOException {
+
+        logger.info("\tGenerating " + sourceTargetProteinMappingFilename);
+        File sourceTargetProteinMappingFile = new File(sourceTargetProteinMappingFilename);
+        if (sourceTargetProteinMappingFile.exists()) {
+            sourceTargetProteinMappingFile.delete();
+        }
+        sourceTargetProteinMappingFile.createNewFile();
```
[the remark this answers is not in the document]
cookersjs: When I comment out the above 4 lines, it only appends to existing files. Same if I only leave the `.createNewFile()` method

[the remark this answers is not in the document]
SolomonShorser-OICR: See: https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/file/Files.html 
The `write()` method takes an `OpenOption` argument, which could let you specify if you want to overwrite or append.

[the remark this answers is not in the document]
jweiser: Hmm, that's strange, though.  According to the JavaDocs, it should create or overwrite by default:

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#write-java.nio.file.Path-byte:A-java.nio.file.OpenOption...-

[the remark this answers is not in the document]
SolomonShorser-OICR: I suspect it's because this code is using `java.io.File` https://docs.oracle.com/javase/8/docs/api/index.html?java/io/File.html

NOT `java.nio.files`.

PR #27, opened 2019-02-26
Feature/orthoinference panther implementation

cookersjs:
Small PR that encompasses two things:

1) Implementation of PANTHER data into Orthoinference

2) Updated any 'Compara' bits (such as Summation text) to reference PANTHER

Also some tiny quality of life items

### DIFF
--- a/orthoinference/runOrthoinference.sh
+++ b/orthoinference/runOrthoinference.sh
@@ -12,7 +12,7 @@ allSpecies=(pfal spom scer ddis cele sscr btau cfam mmus rnor ggal xtro drer dme
 for species in "${allSpecies[@]}"
 do
 	echo "java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference_$species.out";
-	java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference.out;
+	java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference_$species.out;
 done
 echo "Orthoinference complete"
 
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
@@ -36,40 +36,30 @@ public class EWASInferrer {
 	private static Map<String,GKInstance> ewasIdenticals = new HashMap<String,GKInstance>();
 	private static Map<String,GKInstance> residueIdenticals = new HashMap<String,GKInstance>();
 
-    // Creates an array of inferred EWAS instances from the homologue mappings file (hsap_species_mapping.txt).
+	// Creates an array of inferred EWAS instances from the homologue mappings file (hsap_species_mapping.txt).
 	@SuppressWarnings("unchecked")
 	public static List<GKInstance> inferEWAS(GKInstance ewasInst) throws InvalidAttributeException, Exception
 	{
 		List<GKInstance> infEWASInstances = new ArrayList<GKInstance>();
 		String referenceEntityId = ((GKInstance) ewasInst.getAttributeValue(referenceEntity)).getAttributeValue(identifier).toString();
 		if (homologueMappings.get(referenceEntityId) != null)
-			{
-				// Iterate through the array of homologue mappings, attempting to infer EWAS instances for each.
-				for (String homologue : homologueMappings.get(referenceEntityId))
-				{
-					String[] splitHomologue = homologue.toString().split(":");
-					String homologueSource = splitHomologue[0];
-					String homologueId = splitHomologue[1];
-					
+		{
+			// Iterate through the array of homologue mappings, attempting to infer EWAS instances for each.
+			for (String homologueId : homologueMappings.get(referenceEntityId)) {
+				if (checkValidSpeciesProtein(homologueId)) {
 					GKInstance infReferenceGeneProductInst = null;
-					if (referenceGeneProductIdenticals.get(homologueId) == null)
-					{
+					if (referenceGeneProductIdenticals.get(homologueId) == null) {
 						infReferenceGeneProductInst = InstanceUtilities.createNewInferredGKInstance((GKInstance) ewasInst.getAttributeValue(referenceEntity));
 						infReferenceGeneProductInst.addAttributeValue(identifier, homologueId);
 						// Reference DB can differ between homologue mappings, but can be differentiated by the 'homologueSource' found in each mapping.
-						GKInstance referenceDatabaseInst = null;
-						if (homologueSource.equals("ENSP"))
-						{
-							referenceDatabaseInst = enspDbInst;
-						} else {
-							referenceDatabaseInst = uniprotDbInst;
-						}
-						infReferenceGeneProductInst.addAttributeValue(referenceDatabase,  referenceDatabaseInst);
-						
+						// With PANTHER data, the Protein IDs are exclusively UniProt
+						GKInstance referenceDatabaseInst = uniprotDbInst;
+						infReferenceGeneProductInst.addAttributeValue(referenceDatabase, referenceDatabaseInst);
+
 						// Creates ReferenceDNASequence instance from ReferenceEntity
 						List<GKInstance> inferredReferenceDNAInstances = createReferenceDNASequence(homologueId);
 						infReferenceGeneProductInst.addAttributeValue(referenceGene, inferredReferenceDNAInstances);
-						
+
 						infReferenceGeneProductInst.addAttributeValue(species, speciesInst);
 						infReferenceGeneProductInst.setAttributeValue(_displayName, "UniProt:" + homologueId);
 						infReferenceGeneProductInst = InstanceUtilities.checkForIdenticalInstances(infReferenceGeneProductInst);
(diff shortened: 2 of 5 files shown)


orthoinference/runOrthoinference.sh
```diff
@@ -12,7 +12,7 @@ allSpecies=(pfal spom scer ddis cele sscr btau cfam mmus rnor ggal xtro drer dme
 for species in "${allSpecies[@]}"
 do
 	echo "java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference_$species.out";
-	java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference.out;
+	java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar $species > orthoinference_$species.out;
```
SolomonShorser-OICR: I would assume at some point the version number of the Java application will change, and there will need to be a variable in this script that contains the version number and it gets used as 
```bash
java -jar target/orthoinference-$ORTHOINFERENCE_VERSION-jar-with-dependencies.jar $species > orthoinference_$species.out;
```

SolomonShorser-OICR:
What is the purpose of this PR?

cookersjs:
> What is the purpose of this PR?

The paragraph at the top outlined it -- Its implementation of Orthopair data into Orthoinference as well as updating some fields that mentioned Compara

SolomonShorser-OICR:
ah, sorry, I missed that!

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -80,118 +70,99 @@
...
-					for (int endCoord : (Collection<Integer>) ewasInst.getAttributeValuesList(endCoordinate))
-					{
+					for (int endCoord : (Collection<Integer>) ewasInst.getAttributeValuesList(endCoordinate)) {
 						infEWASInst.addAttributeValue(endCoordinate, endCoord);
 					}
-//					int startCoord = (int) infEWASInst.getAttributeValue(startCoordinate);
-//					int endCoord = (int) infEWASInst.getAttributeValue(endCoordinate);
-					if (infEWASInst.getAttributeValue(startCoordinate) != null && (int) infEWASInst.getAttributeValue(startCoordinate) > 1 || infEWASInst.getAttributeValue(endCoordinate) != null && (int) infEWASInst.getAttributeValue(endCoordinate) > 1) 
-					{
+					//					int startCoord = (int) infEWASInst.getAttributeValue(startCoordinate);
```
SolomonShorser-OICR: Can this and the line below be removed?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -92,8 +92,14 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 		String eligibleFilename = "eligible_" + species	+ "_75.txt";
 		String inferredFilename = "inferred_" + species + "_75.txt";
 		File eligibleFile = new File(eligibleFilename);
+		if (eligibleFile.exists()) {
+			eligibleFile.delete();
```
SolomonShorser-OICR: Just curious: why delete the files? Are they so large that they take up lots of storage? Does leaving them in place cause issues with other steps?

orthoinference/runOrthoinference.sh
cookersjs: Would it make sense to try and get the version number from the jar file in the `target` folder?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: The script will append lines to the files, which can make it annoying to try and find out what inferences were attempted/completed on the current run

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -290,61 +271,60 @@ public static void readENSGMappingFile(String toSpecies, String pathToOrthopairs
...
+			for (String proteinId : spaceSplit)
 			{
-				String[] colonSplit = proteinId.split(":");
-				if (ensgMappings.get(colonSplit[1]) == null)
+				if (ensgMappings.get(proteinId) == null)
 				{
 					List<String> singleArray = new ArrayList<String>();
 					singleArray.add(ensgKey);
-					ensgMappings.put(colonSplit[1], (ArrayList<String>) singleArray);
+					ensgMappings.put(proteinId, (ArrayList<String>) singleArray);
```
jweiser: For putting an array of length 1 into a map, you can do:
`ensgMappings.put(proteinId, Collections.singleton(ensgKey));`

`ensgMappings` can also have its declaration at the start of the class as `Map<String, List<String>> ensgMappings = new HashMap<>();` so a cast to `ArrayList<String>` isn't needed.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -290,61 +271,60 @@ public static void readENSGMappingFile(String toSpecies, String pathToOrthopairs
 			String[] tabSplit = currentLine.split("\t");
 			String ensgKey = tabSplit[0];
 			String[] spaceSplit = tabSplit[1].split(" ");
```
jweiser: Can `spaceSplit` be renamed to `proteinIds`?

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -356,27 +336,27 @@ public static void createAlternateReferenceDBInstance(String toSpeciesLong, JSON
...
-	
-	public static void resetVariables() 
+
+	public static void resetVariables()
 	{
-		homologueMappings = new HashMap<String,String[]>();
+		homologueMappings = new HashMap<>();
 		ensgMappings = new HashMap<String,ArrayList<String>>();
 		referenceGeneProductIdenticals = new HashMap<String,GKInstance>();
 		ewasIdenticals = new HashMap<String,GKInstance>();
```
jweiser: For consistency, can the typing be removed from the right side for these three lines?

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -218,10 +186,23 @@
...
+		}
 		return infEWASInstances;
 	}
 
-    // Creates ReferenceGeneSequence instance based on ENSG identifier mapped to protein. Creates an instance for the primary database and an alternate, if it exists.
+	// Homologous Protein IDs can exist in ${source}_${target}_mapping.txt but the corresponding Gene ID might not exist in ${target}_gene_protein_mapping.txt.
+	// This is different from when we built Orthopairs files using Compara, since the homology mapping file was generated using IDs from the gene-protein file.
+	// This function prevents a Null Exception from killing the entire Reaction's inference, rather than just the EWAS inference.
+	private static boolean checkValidSpeciesProtein(String homologueId) {
+		List<String> ensgIds = ensgMappings.get(homologueId);
```
jweiser: Is it always the case that if a homologueId key returns a null value from the map it means it isn't in the map or do homologue ids get included in the ensg map even if they have no gene ids to map to?

If homologue ids without gene ids aren't in the map as keys, I think you can do this check like this: 

`return ensgMappings.containsKey(homologueId);`

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
cookersjs: Yes, intellij has helpfully grayed them out. Is there a way to purge a project of unnecessary code?

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
cookersjs: Very cool, wasn't aware of that function

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
jweiser: I haven't tried it myself, but according to Google there is a drop down option for "Code | Inspect Code" that should show things picked up by IntelliJ.  I don't know if there's a way to purge in one step, though.

orthoinference/runOrthoinference.sh
SolomonShorser-OICR: Me, I'd just have a variable in the script and ensure that it's updated every time this project is released. It *could* be automated, if you want, probably by getting maven to to a find-and-replace on a token in the shell script.

SolomonShorser-OICR: But more importantly, update the version number in the pom file if this is no longer a snapshot. If it's ready to be used in production, you can set the version number to 1.0.0.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
SolomonShorser-OICR: If that's the problem, it could put separators between the sections it appends. Or it could use timestamps...

But since you're deleting the files, I'm guessing that it's not really necessary to keep them around for the long term.

PR #29, opened 2019-03-07
Feature/add ortho stable ids

cookersjs:
- This is a  sub-branch of feature/generate_orthoinference_stable_identifiers, and I want to merge it with it's parent

- This code creates a Stable Identifier instance for inferred PhysicalEntitys, ReactionlikeEvents and Pathways during orthoinference

- It adds a single class with 3 entry points in the code; it modifies slightly InstanceUtilities.checkForIdenticalInstances, which now receives the inferred AND original instance (instead of just the inferred instance) when a stableIdentifier needs to be generated. In cases where a stableIdentifier does not need to be generated, 'null' is provided as the second argument

### DIFF
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -103,6 +103,7 @@ public class EventsInferrer
 		inferredFile.createNewFile();
 		ReactionInferrer.setEligibleFilename(eligibleFilename);
 		ReactionInferrer.setInferredFilename(inferredFilename);
+		StableIdentifierGenerator.setSpeciesAbbreviation((String) speciesObject.get("abbreviation"));
 
 		// Set static variables (DB/Species Instances, mapping files) that will be repeatedly used
 		setInstanceEdits(personId);
@@ -236,6 +237,7 @@ public class EventsInferrer
 		SkipInstanceChecker.setAdaptor(dbAdaptor);
 		InstanceUtilities.setAdaptor(dbAdaptor);
 		OrthologousEntityGenerator.setAdaptor(dbAdaptor);
+		StableIdentifierGenerator.setAdaptor(dbAdaptor);
 		EWASInferrer.setAdaptor(dbAdaptor);
 		HumanEventsUpdater.setAdaptor(dbAdaptor);
 		
@@ -312,6 +314,7 @@ public class EventsInferrer
 		instanceEditInst = InstanceEditUtils.createInstanceEdit(dbAdaptor, personId, "org.reactome.orthoinference");
 		InstanceUtilities.setInstanceEdit(instanceEditInst);
 		OrthologousEntityGenerator.setInstanceEdit(instanceEditInst);
+		StableIdentifierGenerator.setInstanceEdit(instanceEditInst);
 		EWASInferrer.setInstanceEdit(instanceEditInst);
 		HumanEventsUpdater.setInstanceEdit(instanceEditInst);
 	}
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -191,6 +191,7 @@ public class EventsInferrer
 				return;
 			} catch (Exception e) {
 				e.printStackTrace();
+				return;
 			}
 		}
 		HumanEventsUpdater.setInferredEvent(ReactionInferrer.getInferredEvent());
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -272,7 +272,7 @@ public class EventsInferrer
 		speciesInst.addAttributeValue(created, instanceEditInst);
 		speciesInst.addAttributeValue(name, toSpeciesLong);
 		speciesInst.addAttributeValue(_displayName, toSpeciesLong);
-		speciesInst = InstanceUtilities.checkForIdenticalInstances(speciesInst);
+		speciesInst = InstanceUtilities.checkForIdenticalInstances(speciesInst, null);
 		logger.info("Species instance is " + speciesInst);
 		OrthologousEntityGenerator.setSpeciesInstance(speciesInst);
 		EWASInferrer.setSpeciesInstance(speciesInst);
@@ -287,7 +287,7 @@ public class EventsInferrer
 		String summationText = "This event has been computationally inferred from an event that has been demonstrated in another species.<p>The inference is based on the homology mapping from PANTHER. Briefly, reactions for which all involved PhysicalEntities (in input, output and catalyst) have a mapped orthologue/paralogue (for complexes at least 75% of components must have a mapping) are inferred to the other species. High level events are also inferred for these events to allow for easier navigation.<p><a href='/electronic_inference_compara.html' target = 'NEW'>More details and caveats of the event inference in Reactome.</a> For details on PANTHER see also: <a href='http://www.pantherdb.org/about.jsp' target='NEW'>http://www.pantherdb.org/about.jsp</a>";
 		summationInst.addAttributeValue(text, summationText);
 		summationInst.addAttributeValue(_displayName, summationText);
-		summationInst = InstanceUtilities.checkForIdenticalInstances(summationInst);
+		summationInst = InstanceUtilities.checkForIdenticalInstances(summationInst, null);
 		
 		ReactionInferrer.setSummationInstance(summationInst);
 		HumanEventsUpdater.setSummationInstance(summationInst);
@@ -302,7 +302,7 @@ public class EventsInferrer
 		evidenceTypeInst.addAttributeValue(name, evidenceTypeText);
 		evidenceTypeInst.addAttributeValue(name, "IEA");
 		evidenceTypeInst.addAttributeValue(_displayName, evidenceTypeText);
-		evidenceTypeInst = InstanceUtilities.checkForIdenticalInstances(evidenceTypeInst);
+		evidenceTypeInst = InstanceUtilities.checkForIdenticalInstances(evidenceTypeInst, null);
 		ReactionInferrer.setEvidenceTypeInstance(evidenceTypeInst);
 		HumanEventsUpdater.setEvidenceTypeInstance(evidenceTypeInst);
 	}
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -186,9 +186,6 @@ public class EventsInferrer
 			try {
 				logger.info("Attempting to infer " + reactionInst);
 				ReactionInferrer.inferReaction(reactionInst);
-			} catch (RuntimeException e) {
-				e.printStackTrace();
-				return;
 			} catch (Exception e) {
 				e.printStackTrace();
 				return;
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -50,6 +50,7 @@ public class EventsInferrer
 	private static GKInstance speciesInst;
 	private static Map<GKInstance,GKInstance> manualEventToNonHumanSource = new HashMap<GKInstance,GKInstance>();
 	private static List<GKInstance> manualHumanEvents = new ArrayList<GKInstance>();
+	private static StableIdentifierGenerator stableIdentifierGenerator;
 
 	@SuppressWarnings("unchecked")
 	public static void inferEvents(Properties props, String pathToConfig, String species) throws Exception
@@ -103,7 +104,9 @@ public class EventsInferrer
 		inferredFile.createNewFile();
 		ReactionInferrer.setEligibleFilename(eligibleFilename);
 		ReactionInferrer.setInferredFilename(inferredFilename);
-		StableIdentifierGenerator.setSpeciesAbbreviation((String) speciesObject.get("abbreviation"));
+
+
+		stableIdentifierGenerator = new StableIdentifierGenerator(dbAdaptor, (String) speciesObject.get("abbreviation"));
 
 		// Set static variables (DB/Species Instances, mapping files) that will be repeatedly used
 		setInstanceEdits(personId);
@@ -183,6 +186,9 @@ public class EventsInferrer
 			try {
 				logger.info("Attempting to infer " + reactionInst);
 				ReactionInferrer.inferReaction(reactionInst);
+			} catch (RuntimeException e) {
+				e.printStackTrace();
+				return;
 			} catch (Exception e) {
 				e.printStackTrace();
 			}
@@ -193,6 +199,10 @@ public class EventsInferrer
 		resetVariables();
 		System.gc();
 		logger.info("Finished orthoinference of " + speciesName + ".");
+	}
+
+	public static StableIdentifierGenerator getStableIdentifierGenerator() {
+		return stableIdentifierGenerator;
 	}
 
 	private static void setReleaseDates(String dateOfRelease) 
@@ -237,7 +247,6 @@ public class EventsInferrer
 		SkipInstanceChecker.setAdaptor(dbAdaptor);
 		InstanceUtilities.setAdaptor(dbAdaptor);
 		OrthologousEntityGenerator.setAdaptor(dbAdaptor);
-		StableIdentifierGenerator.setAdaptor(dbAdaptor);
 		EWASInferrer.setAdaptor(dbAdaptor);
 		HumanEventsUpdater.setAdaptor(dbAdaptor);
 		
@@ -314,7 +323,6 @@ public class EventsInferrer
 		instanceEditInst = InstanceEditUtils.createInstanceEdit(dbAdaptor, personId, "org.reactome.orthoinference");
 		InstanceUtilities.setInstanceEdit(instanceEditInst);
 		OrthologousEntityGenerator.setInstanceEdit(instanceEditInst);
-		StableIdentifierGenerator.setInstanceEdit(instanceEditInst);
 		EWASInferrer.setInstanceEdit(instanceEditInst);
 		HumanEventsUpdater.setInstanceEdit(instanceEditInst);
 	}
--- a/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
@@ -8,6 +8,8 @@ import org.gk.persistence.MySQLAdaptor;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
+
 import static org.gk.model.ReactomeJavaConstants.*;
 
 /*
@@ -40,14 +42,11 @@ public class StableIdentifierGenerator {
 
         // Paralogs will have the same base stable identifier, but we want to denote when that happens.
         // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
-        if (seenOrthoIds.get(targetIdentifier) == null) {
-            seenOrthoIds.put(targetIdentifier, 1);
-        } else {
-            int paralogCount = seenOrthoIds.get(targetIdentifier);
-            paralogCount++;
-            seenOrthoIds.put(targetIdentifier, paralogCount);
+        int paralogCount = Optional.ofNullable(seenOrthoIds.get(targetIdentifier)).orElse(0) + 1;
+        if (paralogCount > 1) {
             targetIdentifier += "-" + paralogCount;
         }
+        seenOrthoIds.put(targetIdentifier, paralogCount);
 
         // Check that the stable identifier instance does not already exist in DB
         // TODO: Performance check
@@ -69,7 +68,6 @@ public class StableIdentifierGenerator {
 
         // Populate inferred instance with new StableIdentifier instance
         inferredInst.addAttributeValue(stableIdentifier, orthoStableIdentifierInst);
-
         return inferredInst;
     }
 }
--- a/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
@@ -1,10 +1,66 @@
 package org.reactome.orthoinference;
 
 import org.gk.model.GKInstance;
+import org.gk.persistence.MySQLAdaptor;
 
+import java.util.HashMap;
+import java.util.Map;
+import static org.gk.model.ReactomeJavaConstants.*;
+
+/*
+ *  All PhysicalEntitys, ReactionlikeEvents and Pathways are routed to this class to generate their stable identifiers
+ */
 public class StableIdentifierGenerator {
 
-    public static void generateOrthologousStableId(GKInstance inferredInst, GKInstance stableIdentifierInst) {
+    private static MySQLAdaptor dba;
+    private static GKInstance instanceEditInst;
+    private static Map<String,Integer> seenOrthoIds = new HashMap<>();
+    private static String speciesAbbreviation = null;
 
+    public static GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstance originalInst) throws Exception {
+
+        // All Human PhysicalEntitys and Events will have a StableIdentifier instance in the stableIdentifier attribute
+        GKInstance stableIdentifierInst = (GKInstance) originalInst.getAttributeValue(stableIdentifier);
+
+        // For now, Human is hard-coded as the source species, so we replace the stableIdentifier source species based on that assumption
+        String sourceIdentifier = (String) stableIdentifierInst.getAttributeValue(identifier);
+        String targetIdentifier = sourceIdentifier.replace("HSA", speciesAbbreviation);
+        // Paralogs will have the same base stable identifier, but we want to denote when that happens.
+        // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
+        if (seenOrthoIds.get(targetIdentifier) == null) {
+            seenOrthoIds.put(targetIdentifier, 1);
+        } else {
+            int paralogCount = seenOrthoIds.get(targetIdentifier);
+            paralogCount++;
+            seenOrthoIds.put(targetIdentifier, paralogCount);
+            targetIdentifier += "-" + paralogCount;
+        }
+
+        // Create new StableIdentifier instance
+        GKInstance orthoStableIdentifierInst = InstanceUtilities.createNewInferredGKInstance(stableIdentifierInst);
+        orthoStableIdentifierInst.addAttributeValue(identifier, targetIdentifier);
+        String identifierVersionNumber = "1";
+        orthoStableIdentifierInst.addAttributeValue(identifierVersion, identifierVersionNumber);
+        String orthoStableIdentifierName = targetIdentifier + "." + identifierVersionNumber;
+        orthoStableIdentifierInst.setDisplayName(orthoStableIdentifierName);
+        dba.storeInstance(orthoStableIdentifierInst);
+
+        // Populate inferred instance with new StableIdentifier instance
+        inferredInst.addAttributeValue(stableIdentifier, orthoStableIdentifierInst);
+
+        return inferredInst;
+
+    }
+
+    public static void setSpeciesAbbreviation(String speciesAbbreviationCopy) {
+        speciesAbbreviation = speciesAbbreviationCopy;
+    }
+
+    public static void setInstanceEdit(GKInstance instanceEditInstCopy) {
+        instanceEditInst = instanceEditInstCopy;
+    }
+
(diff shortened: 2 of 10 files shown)


orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private static MySQLAdaptor dba;
+    private static GKInstance instanceEditInst;
+    private static Map<String,Integer> seenOrthoIds = new HashMap<>();
+    private static String speciesAbbreviation = null;
+
+    public static GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstance originalInst) throws Exception {
+
+        // All Human PhysicalEntitys and Events will have a StableIdentifier instance in the stableIdentifier attribute
+        GKInstance stableIdentifierInst = (GKInstance) originalInst.getAttributeValue(stableIdentifier);
```
jweiser: It might be good to have a check here to see if `stableIdentifierInst` is null and log a fatal error with an exception being thrown if so.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // For now, Human is hard-coded as the source species, so we replace the stableIdentifier source species based on that assumption
+        String sourceIdentifier = (String) stableIdentifierInst.getAttributeValue(identifier);
+        String targetIdentifier = sourceIdentifier.replace("HSA", speciesAbbreviation);
```
jweiser: I think the source species was configurable in the old implementation.  Could the same be done here and the sourceSpecies abbreviation taken from the species info file the way it is for the target/inferred species?  

Plant Reactome might want to use the new orthoinference Java implementation, so being able to configure would be helpful for them.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+
+/*
+ *  All PhysicalEntitys, ReactionlikeEvents and Pathways are routed to this class to generate their stable identifiers
+ */
+public class StableIdentifierGenerator {
+
+    private static MySQLAdaptor dba;
+    private static GKInstance instanceEditInst;
+    private static Map<String,Integer> seenOrthoIds = new HashMap<>();
+    private static String speciesAbbreviation = null;
```
jweiser: Since the setting of `speciesAbbreviation` is required for the `generateOrthologousStableId` method, it would probably be good to extract it from the `inferredInst` parameter and set it locally in the method rather than as a class field.  As a class field, we run the risk of the `speciesAbbreviation` being null or set to an incorrect value if the `setSpeciesAbbreviaiton` method hasn't been called to sync the value with the call to `generateOrthologousStableId`.

orthoinference/src/test/java/org/reactome/orthoinference/StableIdentifierGeneratorTest.java
```diff
@@ -0,0 +1,67 @@
...
+        PowerMockito.mockStatic(InstanceUtilities.class);
+        Mockito.when(mockOriginalInst.getAttributeValue("stableIdentifier")).thenReturn(mockStableIdentifierInst);
+        Mockito.when(mockStableIdentifierInst.getAttributeValue("identifier")).thenReturn(identifier);
+
+        StableIdentifierGenerator.setSpeciesAbbreviation(fakeAbbreviation);
+        StableIdentifierGenerator.setAdaptor(mockAdaptor);
+
+        PowerMockito.when(InstanceUtilities.createNewInferredGKInstance(mockStableIdentifierInst)).thenReturn(mockOrthoStableIdentifierInst);
+
+        StableIdentifierGenerator.generateOrthologousStableId(mockInferredInst, mockOriginalInst);
```
jweiser: Can the return value of this call be captured and tested it was updated correctly with an assert?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -103,6 +103,7 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 		inferredFile.createNewFile();
 		ReactionInferrer.setEligibleFilename(eligibleFilename);
 		ReactionInferrer.setInferredFilename(inferredFilename);
+		StableIdentifierGenerator.setSpeciesAbbreviation((String) speciesObject.get("abbreviation"));
```
SolomonShorser-OICR: Is there a `ReactomeJavaConstants` constant for "abbreviation"? If not, maybe one should be added to `ReactomeJavaConstants`.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
SolomonShorser-OICR: Or make it an argument to the methods that need it.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+import static org.gk.model.ReactomeJavaConstants.*;
+
+/*
+ *  All PhysicalEntitys, ReactionlikeEvents and Pathways are routed to this class to generate their stable identifiers
+ */
+public class StableIdentifierGenerator {
+
+    private static MySQLAdaptor dba;
+    private static GKInstance instanceEditInst;
```
jweiser: The `instanceEditInst` doesn't seem to be used.  Can it be added to the modified attribute after the generated stable id is added to the inferred instance?

jweiser: Also as the created attribute for the new stable id instance?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: Yeah, that works too but I would just worry that the inferredInstance and the passed speciesAbbreviation could be out of sync.  The speciesAbbreviation can be derived from the inferredInstance, so passing it as its own parameter isn't really necessary.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+import org.gk.persistence.MySQLAdaptor;
+
+import java.util.HashMap;
+import java.util.Map;
+import static org.gk.model.ReactomeJavaConstants.*;
+
+/*
+ *  All PhysicalEntitys, ReactionlikeEvents and Pathways are routed to this class to generate their stable identifiers
+ */
+public class StableIdentifierGenerator {
```
jweiser: Making this an instantiable class with a constructor that sets instanceEdit and dba on creation would be better than using the static fields and methods.  This would ensure the instanceEdit and dba are set before calling the generate method.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+        // Paralogs will have the same base stable identifier, but we want to denote when that happens.
+        // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
+        if (seenOrthoIds.get(targetIdentifier) == null) {
+            seenOrthoIds.put(targetIdentifier, 1);
+        } else {
+            int paralogCount = seenOrthoIds.get(targetIdentifier);
+            paralogCount++;
+            seenOrthoIds.put(targetIdentifier, paralogCount);
+            targetIdentifier += "-" + paralogCount;
+        }
```
jweiser: This section could be written like this if you like this style.  I like handling nulls conditions in one statement when I can.
```
        int paralogCount = Optional.ofNullable(seenOrthoIds.get(targetIdentifier)).orElse(0) + 1;
        if (paralogCount > 1) {
            targetIdentifier += "-" + paralogCount;
        }
        seenOrthoIds.put(targetIdentifier, paralogCount);
```

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Yes thats something I still need to do with the whole of Orthoinference. I kept it with the source species hard-coded actually to keep it in line with the current implementation. In the near future I'm going to do some work to make it agnostic to source species

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: The speciesAbbreviation comes from the species config file, which already contributes things like the species Name, URL, reference DB, etc. Since the process is called on a per-species basis, if the abbreviation was out of sync that would mean that these other variables are out of sync as well

orthoinference/src/test/java/org/reactome/orthoinference/StableIdentifierGeneratorTest.java
cookersjs: Right, I will do that. I need to kill this coverage priority now before it turns into a bad habit

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: Yea there isn't. In the old code it extracted the abbreviation from the species name ('H'omo 'SA'piens; 'M'us 'MU'sculus) using a regex. I figured having it in a config made more sense

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Right, the 'created' attribute is given to it in the InstanceUtilities call, the instance edit survived from when I was doing the code seperate

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
SolomonShorser-OICR: Ok, I think `ReactomeJavaConstants` comes from the code in the CuratorTool that gets exported to reactome.jar or reactome-base.jar. We can look into adding a constant for the value "abbreviation".

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
SolomonShorser-OICR: Hmmm good point...

SolomonShorser-OICR: Still, if `speciesAbbreviation` is only used in one static method, it's probably  better to just make it an argument to that method. It would mean you don't have to call `setSpeciesAbbreviation` anymore.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: Ah okay, great.  Does the modified attribute need to be updated for the inferred instance after it gets the stable identifier added to it on line 49?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: Cool, I can open a GitHub issue so it's on the to do list if you like?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
jweiser: There are a few constants for newer attributes that could be added to `ReactomeJavaConstants`.  It might be good to look over the current db schema at some point and make sure we have all classes and attributes covered.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: That sounds good to me, thanks

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: I think this is a philosophical question. The StableIdentifier instance will now be created right before the inferred instance, which means that the inferred instance will already have the stable identifier attribute populated by the time its stored in the database. Does it make sense to denote a 'modify' for an instance when its now a part of the creation process?

jweiser: Oh, I see.  I thought the inferred instance was created and stored at this point, but if not then no need for a modified value!

Is there any chance that the inferred instance won't be stored because of an aborted reaction or pre-existing identical instance after the stable identifier instance is already created and stored?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: Done! https://github.com/reactome/data-release-pipeline/issues/30

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: I positioned the stable identifier generator right before the 'store' method is called on the dba for the instance in question. See:
[PhysicalEntitys](https://github.com/reactome/data-release-pipeline/blob/feature/add_ortho_stableIds/orthoinference/src/main/java/org/reactome/orthoinference/InstanceUtilities.java#L136), [Reactions](https://github.com/reactome/data-release-pipeline/blob/feature/add_ortho_stableIds/orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java#L93), [Pathways](https://github.com/reactome/data-release-pipeline/blob/feature/add_ortho_stableIds/orthoinference/src/main/java/org/reactome/orthoinference/HumanEventsUpdater.java#L141)

So while obviously there is always a chance, I think that if a Stable Identifier instance is being created it should mean that its referring instance will be inserted

jweiser: Okay, great.  Yeah, that makes sense.  From a code logic standpoint, it shouldn't happen but an interruption of running could cause it.

In that case, what would happen if the stable id got created and stored but the process was interrupted before storing the inferred instance?  When re-running and resuming creating the inferred instance would it find the already created stable identifier instance and attach it or would it create a new stable identifier instance?

cookersjs: It would create a new one, since the Stable Identifier code doesn't do any identical instance checks. I can't really think of a way around that situation though that wouldn't cause a notable runtime increase, as the identical instance checks are the expensive ones

jweiser: We could try searching for a stable identifier with the same identifier attribute value.  Maybe a cached set of all existing stable identifiers with the species prefix so it's just one DB look-up?

SolomonShorser-OICR: If an interruption happens, should the database be restored and the process restarted from scratch, or is the process generally robust enough to survive an abort and be safely re-run on the same database as before the interruption?

jweiser: I think it should usually be robust enough to continue.  If the interruption happens because of a Reactome data related problem rather than a system one, though, it may be better to start from scratch.

cookersjs: Yea, I'm going to create an issue of this just as a reminder to test that the script is interruption proof

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: I'll add that in. Should it be fatal to orthoinference though? Is it because if a human instance doesn't have a stable identifier, then that is a bigger problem that needs to be addressed?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: StableIdentifierGeneration is called from 3 separate classes, so to have it as an argument would be adding more code. Unless there is a strong belief against setting it using the method, I'd like to leave it as is

SolomonShorser-OICR: ok

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
SolomonShorser-OICR: And the speciesAbbreviation could also be a constructor arg.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: Yeah, it probably should be fatal to orthoinference.  Potentially, we could flag it and continue with other inferences, but it might indicate a bigger underlying problem that would be harder to patch after all the rest of the inference has happened.

We should probably investigate immediately to see why that situation happened.

SolomonShorser-OICR: I vote for :skull: ...with a check that runs at the before orthoinference to ensure that it should enver actually happen.

jweiser: I think that's probably best.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
jweiser: It would probably be good to have this class be instantiable and make the static methods into instance methods.  If you want to ensure only one object of this class exists, though, you could use the singleton pattern.

https://www.tutorialspoint.com/design_pattern/singleton_pattern.htm

I'm just concerned that static methods leave it open to being in an inconsistent state.

cookersjs: We made this class instantiable so now this abbreviation thing isn't done with a setter

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Done

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Done

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Done

orthoinference/src/test/java/org/reactome/orthoinference/StableIdentifierGeneratorTest.java
cookersjs: So when testing that the 'stableIdentifier' attribute exists in the inferred instance it returns null, both in the actual code and in the test code as well. I think that since it's a mock GKInstance, that it never actually gets populated. I could get around this by either making it a real GKInstance or maybe by populating the mock instance ahead of time in the test code (which would not prove anything actually worked in the code). Suggestions?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -183,6 +186,9 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 			try {
 				logger.info("Attempting to infer " + reactionInst);
 				ReactionInferrer.inferReaction(reactionInst);
+			} catch (RuntimeException e) {
+				e.printStackTrace();
+				return;
```
jweiser: Just curious - why does this RuntimException catch block have a return but the Exception block doesn't?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -1,67 +1,73 @@
...
+        this.speciesAbbreviation = speciesAbbreviation;
+    }
+
+    public GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstance originalInst) throws Exception {
 
         // All Human PhysicalEntitys and Events will have a StableIdentifier instance in the stableIdentifier attribute
         GKInstance stableIdentifierInst = (GKInstance) originalInst.getAttributeValue(stableIdentifier);
+        if (stableIdentifierInst == null) {
+            logger.fatal("No stable identifier instance found for " + originalInst + " -- terminating");
+            throw new RuntimeException("Could not find stable identifier instance");
```
jweiser: It might be good to keep the logging message and the exception message the same.  The log statement is nice since it gives the instance that didn't have a stable id, but maybe the "terminating" part can be removed in case the exception is ever caught and dealt with so that the program keeps going?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -1,67 +1,73 @@
...
-        orthoStableIdentifierInst.addAttributeValue(identifierVersion, identifierVersionNumber);
-        String orthoStableIdentifierName = targetIdentifier + "." + identifierVersionNumber;
-        orthoStableIdentifierInst.setDisplayName(orthoStableIdentifierName);
-        dba.storeInstance(orthoStableIdentifierInst);
+        // Check that the stable identifier instance does not already exist in DB
+        // TODO: Performance check
+        Collection<GKInstance> existingStableIdentifier = (Collection<GKInstance>) dba.fetchInstanceByAttribute("StableIdentifier", "identifier", "=", targetIdentifier);
+
+        GKInstance orthoStableIdentifierInst;
+        if (existingStableIdentifier.size() == 0) {
```
jweiser: Could the logic in this if block be a method for creating the stable id?

orthoinference/src/test/java/org/reactome/orthoinference/StableIdentifierGeneratorTest.java
```diff
@@ -56,12 +59,26 @@ public void generateOrthologousStableIdTest() throws Exception {
...
-        StableIdentifierGenerator.setAdaptor(mockAdaptor);
+        Mockito.when(mockAdaptor.fetchInstanceByAttribute("StableIdentifier", "identifier", "=", "R-ABC-123456")).thenReturn(mockInstanceCollection);
 
         PowerMockito.when(InstanceUtilities.createNewInferredGKInstance(mockStableIdentifierInst)).thenReturn(mockOrthoStableIdentifierInst);
+        stIdGenerator.generateOrthologousStableId(mockInferredInst, mockOriginalInst);
+    }
 
-        StableIdentifierGenerator.generateOrthologousStableId(mockInferredInst, mockOriginalInst);
+    @Test
+    public void generateOrthologousStableIdRuntimeExceptionTest() throws Exception {
```
jweiser: For testing an exception, you can use this pattern:
```
@Rule
public ExpectedException expectedException = ExpectedException.none();

@Test
public void generateOrthologousStableIdRuntimeExceptionTest() {
    expectedException.expect(RuntimeException.class);  // only passes with this type of exception
    expectedException.expectMessage("Could not find stable identifier instance"); // checks the exception contains this message
    stIdGenerator.generateOrthologousStableId(mockInferredInst, mockOriginalInst);
}
```

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: Good point

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: I removed the terminating part and made it the same message for the logger and exception message

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
cookersjs: Done

orthoinference/src/test/java/org/reactome/orthoinference/StableIdentifierGeneratorTest.java
cookersjs: Yep I used this logic and adjusted it for the updated logging message

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -191,6 +191,7 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 				return;
 			} catch (Exception e) {
 				e.printStackTrace();
+				return;
```
jweiser: Since the catch block body for Exception and RuntimeException are the same now, they could be combined into one:
```
} catch (RuntimeException | Exception e) {
    e.printStackTrace();
    return;
}
```
The RuntimeException block could also just be removed if you don't mind not having it explicitly since RuntimeException is a child of Exception.  The Exception catch block will catch any kind of exception thrown.

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -70,4 +64,15 @@ public GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstanc
...
     }
+
+    private GKInstance createOrthologousStableIdentifierInstance(GKInstance stableIdentifierInst, GKInstance orthoStableIdentifierInst, String targetIdentifier) throws Exception {
+        orthoStableIdentifierInst = InstanceUtilities.createNewInferredGKInstance(stableIdentifierInst);
+        orthoStableIdentifierInst.addAttributeValue(identifier, targetIdentifier);
+        String identifierVersionNumber = "1";
+        orthoStableIdentifierInst.addAttributeValue(identifierVersion, identifierVersionNumber);
+        String orthoStableIdentifierName = targetIdentifier + "." + identifierVersionNumber;
+        orthoStableIdentifierInst.setDisplayName(orthoStableIdentifierName);
+        dba.storeInstance(orthoStableIdentifierInst);
```
jweiser: Maybe the storing of the new stable id instance can happen outside the method so there is no side effect to the creation and the method is only responsible for creating the Java object?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -32,8 +32,8 @@ public GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstanc
         // All Human PhysicalEntitys and Events will have a StableIdentifier instance in the stableIdentifier attribute
         GKInstance stableIdentifierInst = (GKInstance) originalInst.getAttributeValue(stableIdentifier);
         if (stableIdentifierInst == null) {
-            logger.fatal("No stable identifier instance found for " + originalInst + " -- terminating");
-            throw new RuntimeException("Could not find stable identifier instance");
+            logger.fatal("No stable identifier instance found for " + originalInst);
+            throw new RuntimeException("No stable identifier instance found for " + originalInst);
```
jweiser: Fantastic!  Now that the messages are in sync, you could extract it to a String variable and pass that to the log and exception.

If the message needs to be changed, it will keep them in sync with each other

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -52,16 +52,10 @@ public GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstanc
         // TODO: Performance check
         Collection<GKInstance> existingStableIdentifier = (Collection<GKInstance>) dba.fetchInstanceByAttribute("StableIdentifier", "identifier", "=", targetIdentifier);
 
-        GKInstance orthoStableIdentifierInst;
+        GKInstance orthoStableIdentifierInst = null;
```
jweiser: Just curious why the variable is initialized to null here.  Will the if/else always give it a value?  If so, it can be left just as a declaration.  Usually setting to null indicates that there's a possibility it will remain null and it had to be set so that the variable has a value when it gets used later.

cookersjs: This was an IDE thing. Once I moved `orthoStableIdentifierInst` to an argument of the new method, it required it to be initialized in some way. ¯\_(ツ)_/¯

jweiser: Oh okay.  Actually, you should now be able to remove `orthoStableIdentifierInst` as an argument to the new method.  You can declare it locally in the method and return it.  That should also get rid of the IDE forcing the initialization to null.

PR #32, opened 2019-03-11
Test mode for UpdateDOIs

cookersjs:
Small PR that allows the user to run a test mode of UpdateDOIs. This will output all DOIs that would be updated during this run. This list can be confirmed by the curators and can all be used to populate the UpdateDOIs.report file that a real run will use. Also added some additional logging statements and a sample config.properties file

### DIFF
--- a/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
+++ b/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
@@ -32,6 +32,7 @@ public class Main {
     MySQLAdaptor dbaGkCentral = null;
     long authorIdTR = 0;
     long authorIdGK = 0;
+    boolean testMode = true;
 
     // Properties file should contain information needed to access current Test Reactome and GKCentral databases
     try 
@@ -50,6 +51,7 @@ public class Main {
       authorIdTR = Integer.valueOf(props.getProperty("authorIdTR"));
       authorIdGK = Integer.valueOf(props.getProperty("authorIdGK"));
       int port = Integer.valueOf(props.getProperty("port"));
+      testMode = Boolean.valueOf(props.getProperty("testMode"));
 
       // Set up db connections.
       dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, port);
@@ -58,8 +60,8 @@ public class Main {
       e.printStackTrace();
     }
       UpdateDOIs.setAdaptors(dbaTestReactome, dbaGkCentral);
-      UpdateDOIs.findAndUpdateDOIs(authorIdTR, authorIdGK, pathToReport);
-
+      logger.info("Starting UpdateDOIs");
+      UpdateDOIs.findAndUpdateDOIs(authorIdTR, authorIdGK, pathToReport, testMode);
       logger.info( "UpdateDOIs Complete" );
     }
 }
--- a/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
+++ b/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
@@ -1,8 +1,5 @@
 package org.reactome.release.updateDOIs;
 
-//import org.apache.logging.log4j.LogManager;
-//import org.apache.logging.log4j.Logger;
-
 import java.io.FileInputStream;
 import java.util.Properties;
 
--- a/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
+++ b/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
@@ -51,7 +51,11 @@ public class Main {
       authorIdTR = Integer.valueOf(props.getProperty("authorIdTR"));
       authorIdGK = Integer.valueOf(props.getProperty("authorIdGK"));
       int port = Integer.valueOf(props.getProperty("port"));
-      testMode = Boolean.valueOf(props.getProperty("testMode"));
+      if (props.getProperty("testMode") == null) {
+        testMode = true;
+      } else {
+        testMode = Boolean.valueOf(props.getProperty("testMode"));
+      }
 
       // Set up db connections.
       dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, port);
@@ -62,6 +66,10 @@ public class Main {
       UpdateDOIs.setAdaptors(dbaTestReactome, dbaGkCentral);
       logger.info("Starting UpdateDOIs");
       UpdateDOIs.findAndUpdateDOIs(authorIdTR, authorIdGK, pathToReport, testMode);
-      logger.info( "UpdateDOIs Complete" );
+      if (!testMode) {
+        logger.info("UpdateDOIs Complete");
+      } else {
+        logger.info("Finished test run of UpdateDOIs");
+      }
     }
 }
--- a/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
+++ b/update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
@@ -48,9 +48,7 @@ public class Main {
       authorIdTR = Integer.valueOf(props.getProperty("authorIdTR"));
       authorIdGK = Integer.valueOf(props.getProperty("authorIdGK"));
       int port = Integer.valueOf(props.getProperty("port"));
-      if (props.getProperty("testMode") == null) {
-        testMode = true;
-      } else {
+      if (props.getProperty("testMode") != null) {
         testMode = Boolean.valueOf(props.getProperty("testMode"));
       }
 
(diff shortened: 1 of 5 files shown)


update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -95,19 +103,26 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
 									gkDOI.getAttributeValuesList("modified");
 									gkDOI.addAttributeValue("modified", instanceEditGK);
 									gkDOI.setAttributeValue("doi", updatedDoi);
-									dbaGkCentral.updateInstanceAttribute(gkDOI, "modified");
-									dbaGkCentral.updateInstanceAttribute(gkDOI, "doi");
+									if (!testMode) {
```
SolomonShorser-OICR: Your code will still perform a commit at the end, eve if `testMode` is `true`. Is there a possibility that something might get committed in testMode? Maybe there should be a check for `testMode` at the end, and perform a rollback if `testMode` is `true`.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -43,15 +47,17 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
 		GKInstance instanceEditGK = UpdateDOIs.createInstanceEdit(UpdateDOIs.dbaGkCentral, authorIdGK, creatorFile);
 		// Gets the updated report file if it was provided for this release
 		HashMap<String, HashMap<String,String>> expectedUpdatedDOIs = UpdateDOIs.getExpectedUpdatedDOIs(pathToReport);
+		if (expectedUpdatedDOIs.size() == 0) {
+			logger.warn("No DOIs listed in UpdateDOIs.report. Please add expected DOI and displayName to UpdateDOIs.report.");
+		}
```
jweiser: If there are no expected DOIs, should the program exit or can it continue?

SolomonShorser-OICR: I'm not sure but I think that if `expectedUpdatedDOIs` is empty, then any DOIs will end up in `notUpdated`. I don't think anything would get updated in the database, but maybe it's a good idea to just exit early if there's nothing to do.

...unless I'm wrong, and something _will_ happen, even when `expectedUpdatedDOIs` is empty.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
cookersjs: This was a good point, I needed to actually do a test mode condition around the 'createInstanceEdit' calls above

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
cookersjs: Nothing will happen, I'll add a return statement

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -67,17 +75,17 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
...
 							notUpdated.add(doiWithName);
-							continue;
+							if (!testMode) {
+								continue;
+							}
 						}
 						// This updates the 'modified' field for Pathways instances, keeping track of when changes happened for each instance
-						// TODO: Put this after the update to GK Central so that we make sure values match
 						trDOI.getAttributeValuesList("modified");
 						trDOI.addAttributeValue("modified", instanceEditTR);
```
SolomonShorser-OICR: When `testMode` is set, will `addAttributeValue` complain if `instanceEditTR` is NULL?

cookersjs: It works the same if the `GKInstance` is null. I'm not sure what would happen if this was actually attempted in a DB commit, but no problems come up using the `addAttributeValue` method

SolomonShorser-OICR: Ok

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -67,17 +75,17 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
...
 							String doiWithName = updatedDoi + ":" + nameFromDb;
 							notUpdated.add(doiWithName);
-							continue;
+							if (!testMode) {
+								continue;
+							}
 						}
 						// This updates the 'modified' field for Pathways instances, keeping track of when changes happened for each instance
-						// TODO: Put this after the update to GK Central so that we make sure values match
 						trDOI.getAttributeValuesList("modified");
```
SolomonShorser-OICR: Not urgent, but most of the attribute names that exist in `ReactomeJavaConstants` should probably be replaced with the appropriate constant.

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -51,7 +51,11 @@ public static void main( String[] args ) {
       authorIdTR = Integer.valueOf(props.getProperty("authorIdTR"));
       authorIdGK = Integer.valueOf(props.getProperty("authorIdGK"));
       int port = Integer.valueOf(props.getProperty("port"));
-      testMode = Boolean.valueOf(props.getProperty("testMode"));
+      if (props.getProperty("testMode") == null) {
+        testMode = true;
+      } else {
+        testMode = Boolean.valueOf(props.getProperty("testMode"));
```
jweiser: Since `testMode` is initialized to true, could this be:
```
if (props.getProperty("testMode") != null) {
    testMode = Boolean.valueOf(props.getProperty("testMode"));
}
```

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -32,26 +30,36 @@ public static void setAdaptors(MySQLAdaptor adaptorTR, MySQLAdaptor adaptorGK) {
...
+		}
+		List<String> updated = new ArrayList<>();
+		List<String> notUpdated = new ArrayList<>();
 		try 
 		{
 			// Get all instances in Test Reactome in the Pathway table that don't have a 'doi' attribute starting with 10.3180, the Reactome DOI standard
-			 doisTR = dbaTestReactome.fetchInstanceByAttribute("Pathway", "doi", "NOT REGEXP", "^10.3180");
-			// Used during testing
-//			doisTR = dbaTestReactome.fetchInstanceByAttribute("Pathway", "DB_ID", "REGEXP", "1912408|3232118|3232142|4085377|4090294|4655427|4755510|8985947|9013694|9034015");
+			 doisTR = dbaTestReactome.fetchInstanceByAttribute(ReactomeJavaConstants.Pathway, "doi", "NOT REGEXP", "^10.3180");
```
jweiser: The 10.3180 could be extracted to a constant like "REACTOME_DOI_PREFIX" to document its meaning.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -61,59 +69,70 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
 					for (GKInstance trDOI : doisTR)
 					{
 						// The dois are constructed from the instances 'stableIdentifier', which should be in the db already
-						String stableIdFromDb = ((GKInstance) trDOI.getAttributeValue("stableIdentifier")).getDisplayName();
-						String nameFromDb = trDOI.getAttributeValue("name").toString();
+						String stableIdFromDb = ((GKInstance) trDOI.getAttributeValue(ReactomeJavaConstants.stableIdentifier)).getDisplayName();
+						String nameFromDb = trDOI.getAttributeValue(ReactomeJavaConstants.name).toString();
 						String updatedDoi = "10.3180/" + stableIdFromDb;
```
jweiser: That constant could also be used here, too.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
cookersjs: Yea good idea

cookersjs:
> Changes look good -- just a couple of minor comments.
> 
> I noticed that testMode's state is tested in multiple places. It might be nice to consolidate so the state only has to be checked once, but it's not essential and it doesn't need to be done for this PR.

Yea I did think of that when I was doing it. I expect that through a second run-through of much of the code I will find efficiencies like that that I will want to implement

PR #34, opened 2019-03-14
Feature/fix reports

SolomonShorser-OICR:
Fixed the format of the reports, as per Lisa's request: "Creator" information is now presented in other reports. Corrected the order of two columns in refEntChangeLog. Some minor cleanup and refactoring.


chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+	{
+		public void setFormulae(List<DataItem> formulae)
+		{
+			this.formulae = formulae;
+		}
+	}
+	
+	
+	/**
+	 * Makes calls to the ChEBI web service to get info for speciefied ChEBI
```
jweiser: Small typo for specified

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+		{
+			AtomicInteger counter = new AtomicInteger(0);
+			// The web service calls are a bit slow to respond, so do them in parallel.
+			refMolecules.parallelStream().forEach(molecule ->
+			{
+				String identifier = null;
+				try
+				{
+					identifier = (String) molecule.getAttributeValue("identifier");
+					if (identifier != null && !identifier.trim().equals(""))
```
jweiser: If you don't mind using apache commons, you could do this with `!StringUtils.isBlank(identifier)`:

https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+		
+		// BufferedWriter is supposed to be thread-safe.
+		try(FileWriter fileWriter = new FileWriter("chebi-cache", true);
+			BufferedWriter bw = new BufferedWriter(fileWriter);)
+		{
+			AtomicInteger counter = new AtomicInteger(0);
+			// The web service calls are a bit slow to respond, so do them in parallel.
+			refMolecules.parallelStream().forEach(molecule ->
+			{
+				String identifier = null;
```
jweiser: The identifier String should be declarable inside the try on line 92, I think?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+				catch (InvalidAttributeException e)
+				{
+					logger.error("InvalidAttributeException caught while trying to get the \"identifier\" attribute on " + molecule.toString());
+					// stack trace should be printed, but I don't think this should break execution, though the only way I can think
+					// of this happening is if the data model changes - otherwise, this exception would probably never be caught.
+					e.printStackTrace();
+				}
+				catch (Exception e)
+				{
+					// general exceptions - print stack trace but keep going.
```
jweiser: Should a general exception be logged too?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+	 * @param identifier
+	 * @param e
+	 */
+	private static void handleWSException(Map<GKInstance, String> failedEntitiesList, GKInstance molecule, String identifier, ChebiWebServiceFault_Exception e)
+	{
+		// "invalid ChEBI identifier" probably shouldn't break execution but should be logged for further investigation.
+		if (e.getMessage().contains("invalid ChEBI identifier"))
+		{
+			logger.error("ERROR: ChEBI Identifier \"{}\" is not formatted correctly.", identifier);
+			failedEntitiesList.put(molecule, "ChEBI Identifier \""+identifier+"\" is not formatted correctly.");
```
jweiser: Could the common error message be extracted into a temp and used in the logger and failedEntitiesList here.   That way it only needs to be edited in one place.  The same for the next branch as well.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+	 * @param molecule - the Molecule to query for.
+	 * @return - The Entity, from ChEBI.
+	 * @throws ChebiWebServiceFault_Exception
+	 * @throws IOException
+	 * @throws Exception
+	 */
+	private Entity getChEBIDataFromWebService(Map<GKInstance, String> failedEntitiesList, final Map<String, List<String>> chebiCache, BufferedWriter bw, GKInstance molecule) throws ChebiWebServiceFault_Exception, IOException, Exception
+	{
+		String identifier = (String) molecule.getAttributeValue(ReactomeJavaConstants.identifier);
+		Entity entity;
```
jweiser: Could this declaration be moved to line 198?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+		}
+		entity = this.chebiClient.getCompleteEntity(identifier);
+		if (entity != null && this.useCache)
+		{
+			bw.write("CHEBI:"+identifier+"\t"+entity.getChebiId()+"\t"+entity.getChebiAsciiName()+"\t"+ (entity.getFormulae().size() > 0 ? entity.getFormulae().get(0).getData() : "") + "\t" + LocalDateTime.now().toString() + "\n");
+			bw.flush();
+		}
+		else
+		{
+			if (entity == null)
```
jweiser: Could this be combined to an `else if (entity == null)`?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,264 @@
...
+
+	/**
+	 * Extract a ChEBI entity from the cache, based on the identifier.
+	 * @param chebiCache - The cache from a file.
+	 * @param identifier - The chebi Identifier.
+	 * @return A ChEBI Entity that will be built from the data in the cache. It will only have: an Identifier, ChEBI Name, Formula.
+	 */
+	private Entity extractChEBIEntityFromCache(Map<String, List<String>> chebiCache, String identifier)
+	{
+		Entity entity;
```
jweiser: Could the declaration be combined to the next line, the same as above?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -324,12 +317,14 @@ private void reportIfMoleculeIdentifierChanged(GKInstance molecule, String newCh
 				for (GKInstance refMol : refMolsWithNewIdentifier)
 				{
 					String newIdentifierReferrersString = referrerIDJoiner(refMol);
-					refMolIdentChangeLog.info("{}\t{}\t{}\t{}\t{}\t{}\t{}", molecule.getDBID(), molecule.toString(), oldMoleculeIdentifier, newChebiID, refMol.getDBID(), oldIdentifierReferrersString, newIdentifierReferrersString);
+					GKInstance creator = ChebiUpdater.getCreator(molecule);
+					refMolIdentChangeLog.info("{}\t{}\t{}\t{}\t{}\t{}\t{}\t{}", molecule.getDBID(), creator.toString(), molecule.toString(), oldMoleculeIdentifier, newChebiID, refMol.getDBID(), oldIdentifierReferrersString, newIdentifierReferrersString);
```
jweiser: Could you use `String.join("\t", ...);` so the delimiter is given once?

chebi-update/src/test/java/org/reactome/release/chebiupdate/ChebiDataRetrieverTest.java
```diff
@@ -0,0 +1,203 @@
...
+		String identifier = "112233";
+		when(molecule1.getAttributeValue(ReactomeJavaConstants.identifier)).thenReturn(identifier);
+		Map<GKInstance, String> failedEntities = new HashMap<>();
+		try
+		{
+			when(chebiClient.getCompleteEntity(identifier)).thenReturn(null);
+			ChebiDataRetriever retriever = new ChebiDataRetriever(false);
+			retriever.retrieveUpdatesFromChebi(Arrays.asList(molecule1), failedEntities);
+		}
+		catch (ChebiWebServiceFault_Exception e)
```
jweiser: Could you combine the catch statements using `|` since they have the same body?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: But then it couldn't be passed to `ChebiDataRetriever.handleWSException` in the catch block. Admittedly, it _might_ be null there, but it might not be, in case it was set and something bad happened within the try.

jweiser: Ah okay, I didn't see it was being used in the ChEBI exception catch block.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -458,35 +458,35 @@ public void checkForDuplicates() throws SQLException, Exception
...
-			String identifier = duplicates.getString(1);
-			int numberOfDuplicates = duplicates.getInt(2);
-			this.duplicatesSB.append("\n** ReferenceMolecule with identifier " + identifier + " occurs " + numberOfDuplicates + " times:\n\n");
-			String operator = identifier == null ? "IS NULL" : "=";
+			duplicatesLog.info("# DB_ID\tCreator\tDuplicated Identifier\tReferenceMolecule");
+	
+			// Should only be one, but API returns collection.
+			@SuppressWarnings("unchecked")
+			Collection<GKInstance> chebiDBInsts = (Collection<GKInstance>)adaptor.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceDatabase, ReactomeJavaConstants.name, "=", "ChEBI");
+			GKInstance chebiDBInst = chebiDBInsts.stream().findFirst().orElse(null);
```
jweiser: Is it possible that chebiDBInsts could be null triggering an NPE?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -458,35 +458,35 @@ public void checkForDuplicates() throws SQLException, Exception
...
+				String identifier = duplicates.getString(1);
+				int numberOfDuplicates = duplicates.getInt(2);
+				this.duplicatesSB.append("\n** ReferenceMolecule with identifier " + identifier + " occurs " + numberOfDuplicates + " times:\n\n");
+				String operator = identifier == null ? "IS NULL" : "=";
+				
+				AttributeQueryRequest identifierAQR = adaptor.new AttributeQueryRequest(ReactomeJavaConstants.ReferenceMolecule, ReactomeJavaConstants.identifier, operator, identifier);
+				
+				@SuppressWarnings("unchecked")
+				Collection<GKInstance> dupesOfIdentifier = (Collection<GKInstance>) adaptor._fetchInstance(Arrays.asList(chebiAQR, identifierAQR));
+				for (GKInstance duplicate : dupesOfIdentifier)
```
jweiser: Most of the logic here seems to be setting up getting the `dupesofIdentifier` collection.  Could that functionality be its own method that takes in the identifier value from each loop?  The method could handle the construction of the details of building the AttributeQueryRequests.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: I suppose it _could_ happen if there are no ReferenceDatabase objects with the name "ChEBI", in which case I think it's perfectly acceptable for the application to crash with an NPE. The presence of the "ChEBI" ReferenceDatabase object is necessary for the rest of the application to run.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: It's only 3 or 4 lines, I'm not sure what the problem here is.

jweiser: I was just getting lost in the details of creating the AQRs.  It's okay to keep it as is if you prefer.  I just thought it would help focus the intent of the logic here to capture the details fetching of the duplicate identifier into a method and then call it.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: A similar query is performed earlier in the program (near the beginning of `updateChebiReferenceMolecules`). I'll added an explicit RuntimeException _there_ to ensure the program terminates if there is no ChEBI RefDB object.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
SolomonShorser-OICR: Some comments have been added, hopefully that will make it easier to follow.

SolomonShorser-OICR:
@cookersjs Any thoughts on this one?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,267 @@
...
+	 * Makes calls to the ChEBI web service to get info for specified ChEBI
+	 * identifiers.
+	 * 
+	 * @param updator
+	 * @param refMolecules - a list of ReferenceMolecules. The Identifier of each of these molecules will be sent to ChEBI to get up-to-date information for that Identifier.
+	 * @param failedEntitiesList - A list of ReferenceMolecules for which no information was returned by ChEBI. Will be updated by this method.
+	 * @return A ReferenceMolecule DB_ID-to-ChEBI Entity map.
+	 * @throws IOException 
+	 */
+	public Map<Long, Entity> retrieveUpdatesFromChebi(Collection<GKInstance> refMolecules, Map<GKInstance, String> failedEntitiesList) throws IOException
```
cookersjs: What is this Entity type?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,267 @@
...
+			BufferedWriter bw = new BufferedWriter(fileWriter);)
+		{
+			AtomicInteger counter = new AtomicInteger(0);
+			// The web service calls are a bit slow to respond, so do them in parallel.
+			refMolecules.parallelStream().forEach(molecule ->
+			{
+				String identifier = null;
+				try
+				{
+					identifier = (String) molecule.getAttributeValue("identifier");
```
cookersjs: Is there any preference to using casting vs using the `.toString()` method from a style/convention point-of-view,

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,267 @@
...
+					logger.error("Exception was caught: {}", e.getMessage());
+					e.printStackTrace();
+				}
+			});
+		}
+		return entityMap;
+	}
+
+	/**
+	 * Handles exceptions from the ChEBI web service. Some of them are ok, as as "invalid ChEBI identifier",
```
cookersjs: as as type

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,267 @@
...
+	 * @throws IOException
+	 */
+	private Map<String,List<String>> loadCacheFromFile() throws IOException
+	{
+		Map<String,List<String>> chebiCache = Collections.synchronizedMap(new HashMap<String, List<String>>());
+		if (this.useCache)
+		{
+			logger.info("useCache is TRUE - chebi-cache file will be read, and populated. Identifiers not in the cache will be queried from ChEBI.");
+			// if the cache exists, load it.
+			if (Files.exists(Paths.get("chebi-cache")))
```
cookersjs: Should 'chebi-cache' string be stored in a variable since it is used multiple times?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
```diff
@@ -0,0 +1,267 @@
...
+	 * Extract a ChEBI entity from the cache, based on the identifier.
+	 * @param chebiCache - The cache from a file.
+	 * @param identifier - The chebi Identifier.
+	 * @return A ChEBI Entity that will be built from the data in the cache. It will only have: an Identifier, ChEBI Name, Formula.
+	 */
+	private Entity extractChEBIEntityFromCache(Map<String, List<String>> chebiCache, String identifier)
+	{
+		// Use the AccessibleEntity to set the formula.
+		Entity entity = new AccessibleEntity();
+		entity.setChebiId(chebiCache.get("CHEBI:"+identifier).get(0));
```
cookersjs: I expect this wouldn't be a problem based on code flow, but would there ever be a worry of chebiCache not containing "CHEBI:`identifier`"?

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: `uk.ac.ebi.chebi.webapps.chebiWS.model.Entity` comes from the ChEBI WebService Client dependency. It's a part of their data model for ChEBI data.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: I think that casting to `String` assumes that the object was a `String` all along (or is of a type that is castable to String), but is being returned as an `Object`. Calling `.toString()` suggests that the receiver of the method call might _not_ be a string, but has a valid string representation. In this case, I think that identifiers _are_ strings, but `getAttributeValue()` will return an `Object`, not a `String`, so I need to force it back to being a `String` so that I can treat it like a `String`.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: :thinking:

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: This method gets called when `chebiCache.containsKey("CHEBI:"+identifier)` evaluates to TRUE, so I'm not sure how this scenario could happen. If the logic surrounding the call to this method changed (such that there was no check before calling), then it _could_ be a problem. But currently, I don't think it could happen that the first time the cache is checked, it contains `CHEBI:${identifier}` and then it suddenly _doesn't_ on the second lookup.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiDataRetriever.java
SolomonShorser-OICR: I had another look, and I think in this case it is better to check the cache inside the method, so I tweaked the logic a bit.

chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -450,218 +464,61 @@ public void checkForDuplicates() throws SQLException, Exception
 				+ "where ReferenceDatabase_2_name.name = 'ChEBI' and identifier is not null\n" + "group by ReferenceEntity.identifier\n"
 				+ "having count(ReferenceMolecule.DB_ID) > 1;\n";
```
cookersjs: Why is this bare query used?

SolomonShorser-OICR: I think this type of query is too complex for the Reactome Java API.

PR #37, opened 2019-03-27
Generalize new orthoinference for Plant Reactome use

preecej:
Mod'd config file, shell script, and ref species references in java to accept a variable reference species ("Oryza sativa/osat" instead of "Homo sapiens/hsap") and a variable inference threshold (1% instead of 75%).

Did NOT generalize new PANTHER code and references to allow for Ensembl Compara. @cookersjs is doing that.


preecej:
Local test run of orthoinference jar with revisions seems to work well, at least at creating data instances in projected species. Did note 3 types of NPE during test run; will need to follow up on those, but first waiting for PANTHER/Compara grandfathering work to be complete before retest.

preecej:
More general question about code in /datafix: does that code need to be updated for continued use by both projects, or was it a one-off for human Reactome?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -129,18 +129,18 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
...
 /**
  *  Start of ReactionlikeEvent inference. Retrieves all human ReactionlikeEvents, and attempts to infer each for the species.
  */
 		// Gets DB instance of source species (human)
-		Collection<GKInstance> sourceSpeciesInst = (Collection<GKInstance>) dbAdaptor.fetchInstanceByAttribute("Species", "name", "=", "Homo sapiens");
+		Collection<GKInstance> sourceSpeciesInst = (Collection<GKInstance>) dbAdaptor.fetchInstanceByAttribute("Species", "name", "=", refSpeciesName);
 		if (sourceSpeciesInst.isEmpty())
 		{
-			logger.info("Could not find Species instance for Homo sapiens");
+			logger.info("Could not find Species instance for reference species");
```
SolomonShorser-OICR: Should the _name_ of the reference species be used here?

cookersjs:
> More general question about code in /datafix: does that code need to be updated for continued use by both projects, or was it a one-off for human Reactome?

For posterity's sake: The code was a one-off fix for when java orthoinference was first integrated into the release pipeline

orthoinference/runOrthoinference.sh
```diff
@@ -7,15 +7,17 @@ git pull
 ## Create new jar file with orthoinference code
 mvn clean compile assembly:single
 
+## set args - all required
+configPath="src/main/resources/config.properties"	# Path to main configuration file
+refSpecies="orig"	# Reference species (4-digit abbv)
+projSpecies=(prj1 prj2 prj3)	# Projected species (4-digit abbv); may contain multiple species, space-delimited
+threshold="1"		# Minimum percentage of orthologous catalysts required to project event
```
cookersjs: This should be a variable we can modify by calling the script or that is retrieved from the config files

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -132,7 +132,7 @@
 						infModifiedResidueInst.setAttributeValue(_displayName, modifiedResidueInst.getAttributeValue(_displayName));
 						// Update name to reflect that coordinate values are taken from humans. This takes place after cache retrieval, since the name from DB won't contain updated name.
 						if (modifiedResidueInst.getAttributeValue(coordinate) != null) {
-							String newModifiedResidueDisplayName = modifiedResidueInst.getAttributeValue(_displayName).toString() + " (in Homo sapiens)";
+							String newModifiedResidueDisplayName = modifiedResidueInst.getAttributeValue(_displayName).toString() + " (in reference species)";
```
cookersjs: This will need to be replaced with a variable containing the reference species full name (I can add this if you want @preecej )

orthoinference/runOrthoinference.sh
```diff
@@ -7,15 +7,17 @@ git pull
 ## Create new jar file with orthoinference code
 mvn clean compile assembly:single
 
+## set args - all required
+configPath="src/main/resources/config.properties"	# Path to main configuration file
+refSpecies="orig"	# Reference species (4-digit abbv)
+projSpecies=(prj1 prj2 prj3)	# Projected species (4-digit abbv); may contain multiple species, space-delimited
```
jweiser: Apologies for the nit-pick, but could the comments say "4-character" rather than "4-digit" just so it's clear the abbreviations are not numeric?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -181,14 +181,14 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 			
 			try {
 				logger.info("Attempting to infer " + reactionInst);
-				ReactionInferrer.inferReaction(reactionInst);
+				ReactionInferrer.inferReaction(reactionInst, threshold);
```
cookersjs: I think rather than including the threshold as an argument I'd prefer it set as a global variable in the necessary classes (see `setSummationInstance()` or `setInstanceEdits` in this class) . That's the way I've done it with other, similar values and I'd like to keep things uniform

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
jweiser: Can the specific reference species value be used here?

orthoinference/src/main/resources/Species.json
```diff
@@ -1,4 +1,50 @@
...
+				"TAIR"
+			]
+		},
+		"mart_group":"athaliana_eg_gene",
+		"name":[
+			"Arabidopsis thaliana"
+		],
+		"mart_virtual_schema":"plants_mart",
+		"group":"Fungi\/Plants"
+	},
```
jweiser: @cookersjs If we add rice and arabidopsis in the Species.json file, will this cause them to be inferred for human orthoinference or will they be ignored because they aren't included in the shell script list of target species?

cookersjs:
As mentioned in the comments, I'd like the threshold to be set as a global variable in the classes that it is needed rather than passed as an argument. I also think a more descriptive variable name than 'threshold' would be good as well. Thanks for identifying all the areas that needed to be genericized though -- wish I had done it from the beginning. If you want me to make any of the changes just let me know @preecej

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
jweiser: I think that's a good idea.  If we could localize it to one place and reference it from there by parts of the code that need it, it would be handy.

orthoinference/src/main/resources/Species.json
cookersjs: Yea the bash script is the determinant of which species is projected to so we won't need to worry about it

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
preecej: Go for it. I considered it, but was moving  fast.

orthoinference/runOrthoinference.sh
preecej: Absolutely. I can change.

orthoinference/runOrthoinference.sh
preecej: Which variable are you referring to?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
preecej: Agreed. It was a pain to thread it through all of those static methods. I had looked for a good place to set it gloabally, but I would like it to exist outside of the compiled code if possible. We might need to adjust this value in the future.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
preecej: Ideally. Just need to pass it in from a global config. I added it to the shell script, but we could pull it from the Species instance, too. Preference?

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
SolomonShorser-OICR: :racing_car:

orthoinference/src/main/resources/Species.json
preecej: This file should have a separate version in the PR fork. It will have 80-100 entries!

orthoinference/runOrthoinference.sh
preecej: threshold, got it.

preecej:
So, big picture: waiting on preferences from Reactome dev group on the location for global threshold value, and how best to source the species name of the ref species. If we can answer those, Justin Cook and I can address.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
jweiser: For sure, the value should live outside the compiled code.  The threshold could be declared in one place, initialized by parsing the configuration, and referenced where needed by the code from the single field containing it.

preecej: on it.

preecej:
Also, good general practice is to avoid using "human" in namespaces. I recommend "reference" or "ref". I am ignoring that in othoinference, however (except where I had to generalize for variable vlaues).

orthoinference/runOrthoinference.sh
preecej: moving to config file...

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
preecej: grabbing from config.properties now

orthoinference/runOrthoinference.sh
preecej: done

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
preecej: done. in config.properties.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
preecej: done

orthoinference/src/main/resources/Species.json
cookersjs: The species file is a config variable so we could always have a 'ReactomeSpecies.json' and a 'PlantReactomeSpecies.json' that still exists on the main branch

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -181,14 +184,14 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 			
 			try {
 				logger.info("Attempting to infer " + reactionInst);
-				ReactionInferrer.inferReaction(reactionInst);
+				ReactionInferrer.inferReaction(reactionInst, inferenceThreshold);
```
cookersjs: Just this inferenceThreshold argument just needs to get removed now

cookersjs:
Looks ready to merge

cookersjs:
@preecej  Are these changes you still want us to review, or are you getting to work on the new code and want us to hold back until finished?

preecej:
@cookersjs Sry, I didn't release my commits last Friday would be associated with the same pull request. I mean for them to be separate, but...they could be taken together if you want to test them. Can we discuss today before you leave?

cookersjs:
Closed, with permission from @preecej

PR #38, opened 2019-03-28
EWASInferrer changes to accomodate varying Orthopair file formats

cookersjs:
This is in response to a request from @preecej to allow the original and new orthopair files to both be used in orthoinference.


preecej:
But firstly, thank you!
Here's that Uniprot:AT issue in RGPs:
orthoinference_Uniprot:ATxxxx_example

SolomonShorser-OICR:
(image does not load)

PR #45, opened 2019-04-08
Feature/ncbi

jweiser:
Creates post-release data exports for Reactome UniProt and/or Pathway data to upload to NCBI, UCSC, and Europe PC.


ncbi/pom.xml
```diff
@@ -0,0 +1,208 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>org.reactome.release</groupId>
+	<artifactId>ncbi</artifactId>
+	<packaging>jar</packaging>
+	<version>1.0-SNAPSHOT</version>
+	<name>ncbi</name>
+	<url>http://maven.apache.org</url>
```
SolomonShorser-OICR: This element should be used to refer to the URL for the project that is built by this pom.

ncbi/pom.xml
```diff
@@ -0,0 +1,208 @@
...
+		<resources>
+			<resource>
+				<directory>src/main/resources</directory>
+				<filtering>true</filtering>
+			</resource>
+		</resources>
+	</build>
+
+
+	<!-- This repo is necessary for the chebiWS-client -->
```
SolomonShorser-OICR: This project requires the ChEBI web service client?

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+
+	/**
+	 * Produces Europe PMC profile XML content
+	 * @return XML string describing Reactome as a data provider to Europe PMC
+	 */
+	private String getEuropePMCProfileXML() {
+		return String.join(System.lineSeparator(),
+			getXMLDeclaration(),
+			"<providers>",
+			indentString("<provider>", 1),
```
SolomonShorser-OICR: For future reference, you can use JAXB to convert Java objects to XML. A simple example: https://www.mkyong.com/java/jaxb-hello-world-example/

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+	private Path getEuropePMCLinkFilePath() {
+		return Paths.get(outputDir, "europe_pmc_links_reactome_" + reactomeVersion + ".xml");
+	}
+
+	/**
+	 * Class to hold the relationship between a Reactome pathway and a literature reference.  The class can generate
+	 * a Europe PMC Link XML Node based on this relationship to be used in the exported Europe PMC Link File.
+	 * @author jweiser
+	 */
+	public static class EuropePMCLink {
```
SolomonShorser-OICR: Is there a reason this is public? I thought inner classes usually had a more limited scope.

ncbi/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,104 @@
...
+	 * NCBI Gene XML
+	 *     "Link" XML nodes describing NCBI Gene Identifier relationships to either UniProt entries or Top Level
+	 *     Pathways in Reactome
+	 * NCBI Gene Protein File (not uploaded to NCBI)
+	 *     Tab delimited file of UniProt accession to NCBI Gene identifiers in Reactome
+	 * NCBI Protein File
+	 *     Entries of all UniProt accessions in Reactome associated with an NCBI Gene Identifier (file only contains
+	 *     UniProt accessions)
+	 * UCSC Entity File
+	 *     Entires of all human, rat, and mouse UniProt accessions in Reactome
```
SolomonShorser-OICR: *Entries

ncbi/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,104 @@
...
+
+		// Write UCSC Entity and Event Files
+		UCSC.getInstance(outputDir, version).writeUCSCFiles(graphDBSession);
+		// Write Europe PMC Profile and Link Files
+		EuropePMC.getInstance(outputDir, version).writeEuropePMCFiles(graphDBSession);
+
+		graphDBSession.close();
+		logger.info("Finished NCBI, UCSC, and Europe PMC export step");
+
+		System.exit(0);
```
SolomonShorser-OICR: I think 0 is a default exit code, so I'm not sure this is necessary. On the other hand, I don't think it will cause any problems...

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
```diff
@@ -0,0 +1,201 @@
...
+
+	/**
+	 * Returns the XML String describing the relationship between the object's UniProt entry and an NCBI Gene
+	 * @param ncbiGene NCBI Gene identifier
+	 * @return XML String for NCBI Gene Entity Link
+	 */
+	public String getEntityLinkXML(String ncbiGene) {
+		return getLinkXML(
+			ncbiGene,
+			"&entity.base.url;",
```
SolomonShorser-OICR: Does the string "&entity.base.url" actually get emitted into the output XML? I think the `&` might give you problems. You might need to encode it as `&amp;`.

ncbi/src/main/java/org/reactome/release/ReactomeConstants.java
```diff
@@ -0,0 +1,6 @@
+package org.reactome.release;
+
+public class ReactomeConstants {
```
SolomonShorser-OICR: We already have a `ReactomeJavaConstants` in Guanming's library. Maybe renaming this to `ReactomeURLs` would reduce the potential for confusion?

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+ * File generator for Europe PMC.  This class will produce files for
+ * Europe PMC profile XML describing Reactome as a data provider and
+ * Europe PMC link XML describing Reactome pathways annotated directly
+ * (i.e. not on parent or child pathways) with PubMed literature references.
+ * @author jweiser
+ */
+public class EuropePMC {
+	private static final Logger logger = LogManager.getLogger();
+	private static final String rootTag = "links";
+	private static final int reactomeProviderID = 1903;
```
SolomonShorser-OICR: What is this?

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+
+		logger.info("Finished writing Europe PMC Link file");
+	}
+
+	/**
+	 * Standard XML declaration header
+	 * @return XML declaration as String
+	 */
+	private String getXMLDeclaration() {
+		return "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>";
```
SolomonShorser-OICR: Does this need to be a function, or could it be String constant?

ncbi/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,104 @@
...
+		logger.info("Files for Reactome version " + version + " will be output to the directory " + outputDir);
+
+		Session graphDBSession = getGraphDBDriver(props).session();
+
+		List<NCBIEntry> ncbiEntries = NCBIEntry.getUniProtToNCBIGeneEntries(graphDBSession);
+
+		// Write NCBI Gene related Protein File
+		NCBIGene.getInstance(ncbiEntries, outputDir, version).writeProteinFile();
+
+		// Write NCBI Gene Files (split into multiple files to conform with 15MB upload maximum)
```
SolomonShorser-OICR: Why not write everything to a single file, then split into chunks < 15 MB after? Isn't it a risk that the user might guess wrong for `numGeneXMLFiles` and produce files that are too large?

ncbi/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,104 @@
...
+			props.load(new FileInputStream(pathToResources));
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		int version = Integer.parseInt(props.getProperty("reactomeVersion", "67"));
+		String outputDir = props.getProperty("outputDir", "archive");
+		Files.createDirectories(Paths.get(outputDir));
+		logger.info("Files for Reactome version " + version + " will be output to the directory " + outputDir);
+
+		Session graphDBSession = getGraphDBDriver(props).session();
```
SolomonShorser-OICR: Not sure, but can this type of Session be used in a try-with-resources statement? That way you don't have to worry about explicitly closing it.

ncbi/src/main/java/org/reactome/release/NCBIGene.java
```diff
@@ -0,0 +1,184 @@
...
+						   "<!ENTITY event.base.url \"" + ReactomeConstants.PATHWAY_BROWSER_URL + "\">",
+						   "]>"
+		).concat(System.lineSeparator());
+	}
+
+	/**
+	 * Returns the opening XML root level tag for the NCBI Gene file
+	 * @return XML opening root level tag as String
+	 */
+	public static String getOpenRootTag() {
```
SolomonShorser-OICR: There are two different classes that have a getOpenRootTag method. This suggests there is a common interface lurking in the design...

ncbi/src/main/java/org/reactome/release/NCBIProtein.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	private Path getNCBIProteinFilePath() {
+		return Paths.get(outputDir, "protein_reactome" + reactomeVersion + ".ft");
+	}
+
+	private String getProteinFileHeader() {
+		return String.join(System.lineSeparator(),
+			getProteinFileSeparator(),
+			"prid:\t4914",
```
SolomonShorser-OICR: What is the significance of the number 4914?

ncbi/src/main/java/org/reactome/release/ReactomeEvent.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/**
+	 * Transforms Reactome event names that contain key words or phrases to standardized names
+	 * @param name Name to be considered for standardization
+	 * @return Standardized Reactome event display name if it contains pre-set words or phrases or the name passed as
+	 * a parameter, otherwise.
+	 */
+	private String fixName(String name) {
+		Map<String, String> patternToReplacement = new TreeMap<>();
```
SolomonShorser-OICR: Since it never changes, this map could probably be a class-level constant, initialized in a static initializer.

ncbi/src/main/java/org/reactome/release/UCSC.java
```diff
@@ -0,0 +1,231 @@
...
+			return ucscUniProtReactomeEntries;
+		}
+
+		logger.info("Fetching UniProt Reactome Entries for UCSC");
+
+		ucscUniProtReactomeEntries = graphDBSession.run(
+			String.join(System.lineSeparator(),
+				"MATCH (ewas:EntityWithAccessionedSequence)-[:referenceEntity]->(rgp:ReferenceGeneProduct)" +
+				"-[:referenceDatabase]->(rd:ReferenceDatabase)",
+				"WHERE ewas.speciesName IN ['Homo sapiens', 'Rattus norvegicus', 'Mus musculus'] AND rd.displayName =" +
```
SolomonShorser-OICR: What's special about these three species?

ncbi/src/main/java/org/reactome/release/Utilities.java
```diff
@@ -0,0 +1,84 @@
...
+		int numberOfExtraKeys = set.size() % numOfSubSets;
+		if (numberOfExtraKeys > 0) {
+			subSetSize += 1;
+		}
+
+		List<Set<String>> splitSets = new ArrayList<>();
+
+		Set<String> subSet = new LinkedHashSet<>();
+		int keyCount = 0;
+		for(String ncbiGeneXMLNodeString : set) {
```
SolomonShorser-OICR: The name `ncbiGeneXMLNodeString` makes some very specific assumptions inside a rather generic-looking function, no? ;)

ncbi/src/test/java/org/reactome/release/EuropePMCTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+		EuropePMC.EuropePMCLink expectedEuropePMCLink = new EuropePMC.EuropePMCLink(
+			"p53-Dependent G1 DNA Damage Response",
+			"R-HSA-69563",
+			"9153395"
+		);
+
+		Set<EuropePMC.EuropePMCLink> europePMCLinks = EuropePMC.fetchEuropePMCLinks(dummyGraphDBServer.getSession());
+
+		assertThat(europePMCLinks, contains(expectedEuropePMCLink));
```
SolomonShorser-OICR: How is this different from using assertTrue, as in `assertTrue(europePMCLinks.contains(expectedEuropePMCLink))`? Actually, where is `contains` even defined?

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+			)
+		)
+		.stream()
+		.map(record ->
+			new EuropePMCLink(
+				record.get("p.displayName").asString(),
+				record.get("p.stId").asString(),
+				Integer.toString(record.get("lr.pubMedIdentifier").asInt())
+			)
+		).collect(Collectors.toCollection(LinkedHashSet::new));
```
cookersjs: What is this `::new` notation?

SolomonShorser-OICR: A new thing in Java 8, still not in very common usage (but becoming more common):
https://stackoverflow.com/questions/20001427/double-colon-operator-in-java-8

ncbi/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,375 @@
...
+		}
+
+		/**
+		 * Checks equality based on object type and value of pathway display name, stable id, and PubMed identifier
+		 * @param o Object to check for equality with the calling Europe PMC Link object.
+		 * @return <code>true</code> if the same object or a Europe PMC Link object with the same pathway display name,
+		 * stable, id and PubMed identifier.  Returns <code>false</code> otherwise.
+		 */
+		@Override
+		public boolean equals(Object o) {
```
cookersjs: More descriptively named variable

ncbi/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,104 @@
...
+				Main.class.getClassLoader().getResource("sample_config.properties")
+			).getPath();
+
+		Properties props = new Properties();
+		try {
+			props.load(new FileInputStream(pathToResources));
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		int version = Integer.parseInt(props.getProperty("reactomeVersion", "67"));
```
cookersjs: Why the hard-coded 67?

ncbi/src/main/java/org/reactome/release/EuropePMC.java
SolomonShorser-OICR: I think this been an old Java convention when overriding "equals". But I agree, it's not the prettiest convention. In newer documentation, I see "obj" used instead of "o".

ncbi/src/main/java/org/reactome/release/Main.java
cookersjs: Why is there a need for the exit?

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	/**
+	 * Checks equality based on object type and value of UniProt accession, UniProt display name, and NCBI Gene ids
+	 * @param o Object to check for equality with the calling NCBI Entry object.
+	 * @return <code>true</code> if the same object or an NCBI Entry object with the same UniProt accession,
+	 * UniProt display name, and NCBI Gene ids.  Returns <code>false</code> otherwise.
+	 */
+	@Override
+	public boolean equals(Object o) {
```
cookersjs: similar as above: variable naming isn't descriptive. As well, this code seems similar to the one above (though I haven't checked if it is or not). Would it make sense to create a 'utilities' class for this step for re-used code?

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
```diff
@@ -0,0 +1,201 @@
...
+			ncbiGene,
+			"&event.base.url;",
+			pathway.getStableIdentifier(),
+			"Reactome Event:" + pathway.getName()
+		);
+	}
+
+	private String getLinkXML(String ncbiGene, String base, String rule, String urlName) {
+		return String.join(System.lineSeparator(),
+			"\t<Link>",
```
cookersjs: You used an 'indent' method up above for a similar problem as here. That looked a bit cleaner and would make the code more consistent

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
SolomonShorser-OICR: A Utilities class _could_ be created that has equality methods comparing instances of various objects, but that's not as object-oriented as having the objects _themselves_ know how to check for equality. ;)
Also, there are libraries that do object comparison and rely on you implementing your own `equals` method for your classes. Failure to implement `equals` means object comparison might not work the way you want/expect it to. I've noticed that Joel used a `TreeMap` in a few places. Since that structure is implemented using a Red-Black tree, it is necessary to be able to compare objects properly.

ncbi/src/main/java/org/reactome/release/NCBIGene.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Writes NCBI Gene XML files describing the relationships between NCBI Gene identifiers and UniProt entries as
+	 * well as their Reactome pathways to pre-set output directory
+	 * @param graphDBSession Neo4J Driver Session object for querying the graph database
+	 * @param numGeneXMLFiles Number of files to divide the Gene XML entries among
+	 * @throws IOException Thrown if creating or appending for any file fails
+	 */
+	public void writeGeneXMLFiles(Session graphDBSession, int numGeneXMLFiles) throws IOException {
+		Path geneErrorFilePath = getGeneErrorFilePath();
+		Files.deleteIfExists(geneErrorFilePath);
```
cookersjs: A potentially trivial suggestion: In the idea of a 'Utilities' class there could be a `deleteAndCreateFile()' method since I've seen it appear multiple times

ncbi/src/main/java/org/reactome/release/PathwayHierarchyUtilities.java
```diff
@@ -0,0 +1,203 @@
...
+	 * to the Set of Pathway database identifiers in which each Reaction Like Event participates
+	 * @param graphDBSession Neo4J Driver Session object for querying the graph database
+	 * @return Map of Reaction Like Event identifier to Set of Pathway identifiers
+	 */
+	public static Map<Long, Set<Long>> fetchRLEIdToPathwayId(Session graphDBSession) {
+		if (rleToPathwayIdCache.containsKey(graphDBSession)) {
+			return rleToPathwayIdCache.get(graphDBSession);
+		}
+
+		if (graphDBSession == null) {
```
cookersjs: Can null be used as a key to a Map? Would it make more sense to have this null check before the above `containsKey` condition?

SolomonShorser-OICR: I am pretty sure that a Java Map can have a key that is null.

ncbi/src/main/java/org/reactome/release/ReactomeConstants.java
cookersjs: `ReactomeURLConstants`?

ncbi/src/main/java/org/reactome/release/ReactomeEvent.java
```diff
@@ -0,0 +1,159 @@
...
+		patternToReplacement.put("amino acids", "Metabolism of nitrogenous molecules");
```
cookersjs: Is there any worry about slightly off lettering or dashes, etc? This is safe as is, without a regex?

ncbi/src/main/java/org/reactome/release/UniProtReactomeEntry.java
```diff
@@ -0,0 +1,390 @@
...
+				uniprotAccessionToReactionLikeEventId.computeIfAbsent(uniprotAccession, k -> new HashSet<>());
+
+			// Add RLEs containing UniProt accession
+			uniprotAccessionToReactomeEvent
+				.computeIfAbsent(uniprotAccession, k -> new HashSet<>())
+				.addAll(reactionLikeEventIds.stream().map(eventCache::get).collect(Collectors.toSet()));
+
+			for (long reactionLikeEventId : reactionLikeEventIds) {
+				Set<Long> pathwayIds = rleToPathwayId.computeIfAbsent(reactionLikeEventId,
+																			  k -> new HashSet<>());
```
cookersjs: The spacing here brings this line far to the right on github

ncbi/src/main/java/org/reactome/release/Utilities.java
```diff
@@ -0,0 +1,84 @@
...
+	/**
+	 * Takes a set of Strings and splits it into a list of the requested number of sub-sets.  If an ordered set, order
+	 * is preserved.  If the set does not split evenly into the number of set-sets requested, the remainder elements
+	 * will be added one element each to the beginning sets of the returned list.
+	 * @param set Set of Strings to split
+	 * @param numOfSubSets Number of sub-sets to return
+	 * @return List of String sub-sets
+	 */
+	public static List<Set<String>> splitSet(Set<String> set, int numOfSubSets) {
+		int subSetSize = set.size() / numOfSubSets ;
```
cookersjs: Why is the `numOfSubSets` variable needed?

ncbi/src/test/java/org/reactome/release/ReactomeEventTest.java
```diff
@@ -0,0 +1,53 @@
...
+		assertThat(reactomeEventCache, hasEntry(EVENT_DB_ID, expectedEvent));
+	}
+
+	@Test
+	public void reactomeEventNameCorrection() {
+		final long DUMMY_DB_ID = 1L;
+		final String DUMMY_STABLE_ID = "R-HSA-123456";
+
+		ReactomeEvent reactomeEvent =
+			new ReactomeEvent(DUMMY_DB_ID, "something to do with sugars", DUMMY_STABLE_ID);
```
cookersjs: The middle string could be a descriptive variable

ncbi/pom.xml
jweiser: Removed!

ncbi/pom.xml
jweiser: Nope, this was a holdover when I used the ChEBI POM file as a template :)

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: I've kept it as an inner class because it does really belong as part of the EuropePMC class, but I've made it package-private since there is some functionality I need to test for the inner class itself.

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: Corrected :)

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: Yes, I think it is.  I tried just `System.exit()` but that method doesn't exist.  I added an explicit exit call here because the program just seemed to continue running in the IDE without it.

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
jweiser: Yes, the string with the ampersand needs to be in the output file as part of the XML content.  I'm a little reluctant to change it as NCBI hasn't said it's a problem for them, but I can send them an e-mail to ask if the `&amp;` encoding would work for them so it is proper.

ncbi/src/main/java/org/reactome/release/ReactomeConstants.java
jweiser: Changed to `ReactomeURLConstants`!

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: This is the identifier Europe PMC has given to Reactome so they have an id to associate with our submissions.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: I've re-worked how the XML is generated, so this is produced programmatically.  I'll need you guys to look over the new XML generation logic when all changes are done and the PR is ready for the second round of review.  I will let you know when to look it over.

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: You're right - that approach would be better.  I'll have to spend some time re-working this, though.  It might be good as an enhancement in a separate branch and PR after this one.

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: Done!

ncbi/src/main/java/org/reactome/release/NCBIGene.java
jweiser: Yup, this has been removed in the generation of XML for one of the export files.  I have to re-work the generation of XML here as well.  I will commit those changes soon before requesting a second round of review.

ncbi/src/main/java/org/reactome/release/NCBIProtein.java
jweiser: This is Reactome's provider id for NCBI.  I've extracted it to a constant so it is documented.

ncbi/src/main/java/org/reactome/release/ReactomeEvent.java
jweiser: Done!

ncbi/src/main/java/org/reactome/release/UCSC.java
jweiser: I'm not sure, to be honest.  This has been in the UCSC data export since I've started with Reactome.  I can ask Robin and Peter if this is still correct.

ncbi/src/main/java/org/reactome/release/Utilities.java
jweiser: Yeah, this was a holdover from when this was extracted from the NCBIGene class.  Changed to be more generic :).

ncbi/src/test/java/org/reactome/release/EuropePMCTest.java
jweiser: `assertThat` and `contains` are methods from the Hamcrest library.  

`assertThat` gives more detail on failure than `assertTrue` in terms of what was expected and what was found during the test.  It's very useful for diagnosing the failure and fixing the issue.

`contains` is a matcher that looks for exactly the provided values occurring in the collection given in the first parameter.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: Yeah, it's a method reference to the constructor to get a new object (used instead of a lamba which would use the traditional syntax with the new keyword).  It creates a new LinkedHashSet to be filled with the results of the stream.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: I've changed `o` to `obj` so it's a little more verbose.

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: This was a holdover from testing.  It's been removed!

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: The program seemed to continue running in my IDE without an explicit exit, so I added it to ensure the program terminated.

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
jweiser: Yeah, unfortunately, there is some duplication in the patterns for the boilerplate of the `equals` methods for the different objects, but it is kind of necessary to ensure objects of those classes knows how to be compared.

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
jweiser: Yes, you're right.  I actually think I'm going to try to generate the XML in a way that's more commonly done in Java.  I'll ask you guys to take a look after I've made those changes.

ncbi/src/main/java/org/reactome/release/NCBIGene.java
jweiser: Done!

ncbi/src/main/java/org/reactome/release/PathwayHierarchyUtilities.java
jweiser: You're right -- the null check should be done first.  I've changed the order.

ncbi/src/main/java/org/reactome/release/ReactomeEvent.java
jweiser: I don't think so.  This has been in place for a long time and I think it might just be a failsafe in case old names are still present.

I may add checks to see if this is called and if it needs updating to be more flexible.

ncbi/src/test/java/org/reactome/release/ReactomeEventTest.java
jweiser: Done!

ncbi/src/main/java/org/reactome/release/UniProtReactomeEntry.java
jweiser: Yeah, this was introduced by IntelliJ from line wrapping.  It's fixed now!

ncbi/src/main/java/org/reactome/release/Utilities.java
jweiser: This is to let the method know into how many sub-sets the set that's being passed should be split.

ncbi/src/main/java/org/reactome/release/Main.java
SolomonShorser-OICR: Really? That's strange. Might be an issue with the IDE? Does it do that when you run it from the command line?

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
SolomonShorser-OICR: Ok. Because I'm quite certain that an XML file containing `&entity.base.url` will fail basic XML validation (probably complain that `&entity.base.url` is not a valid entity, _or_ that it's missing a `;` at the end).

But if NCBI is already accepting files with this in them, I guess it's not a problem.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
SolomonShorser-OICR: Ok.

ncbi/src/main/java/org/reactome/release/Main.java
SolomonShorser-OICR: Ok.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
jweiser: The JAXB looks good!  It might be a little too involved for this kind of XML, but I did find I can more flexibly create XML structures with org.w3c.dom package classes: https://docs.oracle.com/javase/8/docs/api/index.html?org/w3c/dom/package-summary.html

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
jweiser: Actually, I found this is okay because the headers NCBI specified to use are for defining special values of which entity.base.url and event.base.url are two.

https://www.w3schools.com/xml/xml_dtd_entities.asp

ncbi/src/main/java/org/reactome/release/Main.java
jweiser: Seems to happen on the command line as well.

data-exporter/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+						<phase>package</phase> <!-- bind to the packaging phase -->
+						<goals>
+							<goal>single</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-help-plugin</artifactId>
```
SolomonShorser-OICR: What is this plugin for?

SolomonShorser-OICR: Eclipse complains:
> Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-help-plugin:3.1.0:active-profiles (execution: show-profiles, phase: compile)

And if I comment this plugin out, the project builds just fine. So, what does it do?

ncbi/src/main/java/org/reactome/release/Main.java
SolomonShorser-OICR: That's... weird. Probably warrants further investigation.

ncbi/src/main/java/org/reactome/release/NCBIEntry.java
SolomonShorser-OICR: Ah, so they declare their own entity types.

ncbi/src/main/java/org/reactome/release/UCSC.java
SolomonShorser-OICR: Ok.

data-exporter/src/main/java/org/reactome/release/EuropePMC.java
```diff
@@ -0,0 +1,398 @@
+package org.reactome.release;
```
SolomonShorser-OICR: I just noticed this now, but the package for everything in this project should probably be `org.reactome.release.dataexport`. Other projects contain all of their code in their own package, such as `org.reactome.release.chebiupdate` or `org.reactome.release.updateDOIs`.

data-exporter/src/main/java/org/reactome/release/Utilities.java
```diff
@@ -0,0 +1,166 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.*;
+
+/**
+ * General Utilities class.  This class contains methods needed for general manipulation of data or common operations
+ * used by the data export process.
+ * @author jweiser
+ */
+public class Utilities {
```
SolomonShorser-OICR: Possibly rename to `DataExportUtilities`, if the functionality within is only useful for this data-export project? Or would it be useful in more general projects? Are any of these functions useful enough that they could go into release-common-lib?

data-exporter/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.*;
+
+/**
+ * Generates post-release export files for NCBI, UCSC and Europe PMC.
+ * @author jweiser
+ */
+public class Main {
```
SolomonShorser-OICR: release-common-lib has a ReleaseStep abstract class that can be extended for other steps. You can see how it's been used in GO Update and ChEBI Update. I'd like to try and keep the code for the various steps of the release process conforming to a common interface. One reasons is that each ReleaseStep could share common functionality. If there is a reason the ReleaseStep was not used for this project, could you explain why not? Are there improvements that could be made to ReleaseStep that would have made it easier to use in this project?

data-exporter/README.md
```diff
@@ -0,0 +1,61 @@
+# Data Exporter (post-release)
+
+This program will produce data exports for submission to NCBI, UCSC, and Europe PMC.  Files provided to these resources are:
+* **NCBI:** Gene (NCBI gene identifier to Reactome top level pathways) and Protein (all UniProt entries in Reactome associated with any NCBI Gene identifier).
+* **UCSC:** Entity (UniProt entires in Reactome) and Events (UniProt entries in Reactome in relation to Reactome pathways and reactions).
```
cookersjs: *entries

data-exporter/src/main/java/org/reactome/release/NCBIGene.java
```diff
@@ -0,0 +1,204 @@
...
+	 * @param ncbiGeneXMLNodeStrings Set of all the Link XML nodes strings to include in the NCBI Gene XML file(s)
+	 * @return Number of NCBI Gene XML files to create (i.e. split the XML nodes amongst to be under the file size
+	 * limit)
+	 */
+	private int getNumberOfGeneXMLFiles(Set<String> ncbiGeneXMLNodeStrings) {
+		final int BYTES_TO_KILOBYTES = 1024;
+		final int KILOBYTES_TO_MEGABYTES = 1024;
+		final double MAX_FILE_SIZE_IN_MEGABYTES = 15.0;
+
+		int sizeInBytes = ncbiGeneXMLNodeStrings.stream().mapToInt(str -> str.getBytes().length).sum();
```
SolomonShorser-OICR: I haven't tested this - is it an accurate reflection of how many bytes will be required to write this to disc? I would expect that different character sets/encodings might change the size reported here, so make sure you write with the same charset/encoding that this line uses.

 Or, just lower MAX_FILE_SIZE_IN_MEGABYTES to, I dunno, 13 or 14? for a little wiggle room. ;)

data-exporter/src/main/java/org/reactome/release/NCBIGene.java
```diff
@@ -0,0 +1,204 @@
...
+
+	private Path getGeneErrorFilePath() {
+		return Paths.get(outputDir, "geneentrez_" + reactomeVersion + ".err");
+	}
+
+	/**
+	 * Returns the XML header for the NCBI Gene file containing the Reactome entity and event base URLs
+	 * @return XML header as String
+	 */
+	public static String getXMLHeader() {
```
SolomonShorser-OICR: It sounds like the NCBI file can't use the standard Dom XML libraries for output, maybe a little comment explaining why would be nice for future developers, lest they try to rewrite this code and run into problems that you have already experienced.

data-exporter/configureDataExporter.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+echo -n "Enter Neo4J password: "
+# Outputs asterisks instead of plain text
+# Taken from https://stackoverflow.com/a/4316755
+unset password;
+while IFS= read -r -s -n1 pass; do
+  if [[ -z $pass ]]; then
+     echo
+     break
```
cookersjs: I don't really understand this condition. If they do not provide a password, then echo a blank line and break the iterative cycle? Why is there an iteration, and why echo nothing?

data-exporter/configureDataExporter.sh
```diff
@@ -0,0 +1,72 @@
...
+read output_directory
+if [ -z $output_directory ]; then
+    output_directory="$DIR/archive"
+fi
+
+echo -n "Enter Reactome version: "
+read reactome_version
+if [ -z $reactome_version ]; then
+    echo "ERROR: Reactome version required" >&2
+    exit 1;
```
cookersjs: Exiting after they have filled out all the other questions could be annoying -- maybe move this earlier in the question order?

data-exporter/configureDataExporter.sh
```diff
@@ -0,0 +1,72 @@
...
+# Removes comments from the original configuration file
+sed -i '/^###/d' $config_file
+
+# Replaces the dummy database configuration values
+sed -i "s|\(user=\).*|\1$user|" $config_file
+sed -i "s|\(password=\).*|\1$password|" $config_file
+sed -i "s|\(host=\).*|\1$host|" $config_file
+sed -i "s|\(port=\).*|\1$port|" $config_file
+sed -i "s|\(reactomeVersion=\).*|\1$reactome_version|" $config_file
+sed -i "s|\(outputDir=\).*|\1$output_directory|" $config_file
```
cookersjs: This is more out of curiosity: Why use a regex to replace values in the file instead of just creating a new file? At the end of the day with this its basically the same amount of work but in other situations it could be a more relevant question

data-exporter/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+			</plugin>
+		</plugins>
+
+		<resources>
+			<resource>
+				<directory>src/main/resources</directory>
+				<filtering>true</filtering>
+			</resource>
+		</resources>
+	</build>
```
cookersjs: Something we would probably want to standardize is how the maven build and jar execution happens, for consistency's sake

data-exporter/configureDataExporter.sh
jweiser: I got this from an answer on Stack Overflow on how to show asterisks during a password prompt. 
 I don't know the logic inside out, but I believe it is:

* reading a single character at a time into $pass
* outputting an asterisk on screen for that character
* appending the actual character in $pass to $password
* when the user hits enter, doing an empty "echo" for outputting a newline and breaking out of the loop that reads one character per iteration

If they provide no password, then $password will also be empty.

data-exporter/configureDataExporter.sh
jweiser: Good point.  I'll make that change.

data-exporter/configureDataExporter.sh
jweiser: You're right - creating a new file would be simpler.  I'll make the change.

data-exporter/pom.xml
jweiser: Sure, standardizing our builds as much as we can sounds like a good idea.  I'm still learning with maven, so I don't have strong opinions or suggestions on what we should do.

jweiser:
> I'm a bit concerned about the way the application just hangs if you don't explicitly call exit. Is it possible to investigate a little deeper into possible causes? Normally, it's not necessary to need this.
> Hmmm... Maybe there's a resource (file handle, data source connection, etc...) that's being left open, and if that doesn't get closed properly, then the whole application hangs?
> 
> Also, I'd like to see all Java code moved to a more specific Java package. (better to move the files with `git mv`).

data-exporter/src/main/java/org/reactome/release/NCBIGene.java
jweiser: Good point -- I will add that in.

data-exporter/src/main/java/org/reactome/release/NCBIGene.java
jweiser: That's a good idea.  In testing with release 67 data, it worked as expected but I was worried about an edge case of adding the XML header and root tag pushing it just over the edge of 15MB :).  A megabyte of wiggle room sounds good to me.

data-exporter/README.md
jweiser: I keep doing that :)!  I'll correct the spelling.

data-exporter/src/main/java/org/reactome/release/Main.java
jweiser: For sure, I can have it extend the ReleaseStep class.  I remember we talked about that, but I just wasn't thinking of it while making changes here.

data-exporter/src/main/java/org/reactome/release/Utilities.java
jweiser: No problem -- I'll rename it.  There are some general methods that could be re-used, but I'd like to keep them here until they're needed in more than one place in the data-release-pipeline.  I've created an issue just to keep in mind the methods can be moved to release-common-lib if they are needed elsewhere.

https://github.com/reactome/data-release-pipeline/issues/49

data-exporter/src/main/java/org/reactome/release/EuropePMC.java
jweiser: No problem -- I'll create the subpackage.

data-exporter/pom.xml
jweiser: I think this was a carryover from another POM file I used as a template.  I'll remove it.

jweiser:
> I'm a bit concerned about the way the application just hangs if you don't explicitly call exit. Is it possible to investigate a little deeper into possible causes? Normally, it's not necessary to need this.
> 
> Also, I'd like to see all Java code moved to a more specific Java package. (better to move the files with `git mv`).

The cause of the program hanging looks to have been the neo4j driver connection object not having its closed method called.  When I added the driver into the try-with-resources statement before getting a session object from it, the program terminated as expected with an explicit exit call.

The Java code has also been moved to the package 'org.reactome.release.dataexport'.

cookersjs:
New changes look good, my approval status _remains_

data-exporter/src/main/java/org/reactome/release/Main.java
SolomonShorser-OICR: Usually, I have one class that extends the ReleaseStep (such as `ChebiUpdateStep`) and the Main method uses that.

ncbi/src/main/java/org/reactome/release/EuropePMC.java
SolomonShorser-OICR: The DOM packages also work well for creating XML.

data-exporter/src/main/java/org/reactome/release/Main.java
SolomonShorser-OICR: And if you think that ReleaseStep's functionality could be improved, go ahead and do it! For example, the current ReleaseStep provides functionality to create a MySQL driver from the values in a Properties file. If you're using Neo4j, you could add similar functionality to ReleaseStep.

data-exporter/runDataExporter.sh
```diff
@@ -11,10 +38,17 @@ git pull
...
 ## If the config file does not exist, run configuration script
 if [ ! -f $config_file ]; then
-    ./configureDataExporter.sh
+	./configureDataExporter.sh
+fi
+
+jar_file="data-exporter.jar"
+## Generate the jar file if it doesn't exist or a re-build is requested
+if [ ! -f $jar_file ] || [ ! -z $build_jar ]; then
+	mvn clean package
```
cookersjs: Is there any need for a git pull?

jweiser: The git pull happens earlier in the script.  I was going to put it in this block, but it should happen regardless of a build happening or not so that the configuration shell script is up to date.  Also, so the source code is up to date in case anything is done manually.

PR #51, opened 2019-04-24
Feature/orthodiagrams

cookersjs:
Smallish PR that has the orthologous PathwayDiagram generation take place in orthoinference. Rather then call the `PredictedPathwayDiagramGeneratorFromDB.main()` method directly, this now only calls the downstream method `generatedPredictedDiagram` for the Pathways that pertain to the current target species.


orthoinference/pom.xml
```diff
@@ -23,6 +28,11 @@
 		    <artifactId>json-simple</artifactId>
 		    <version>1.1.1</version>
 		</dependency>
+		<dependency>
+			<groupId>org.jdom</groupId>
```
SolomonShorser-OICR: I quickly skimmed the code and didn't see any references to jdom so I'm not sure why it's a dependency here... did i miss it?

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+import org.gk.persistence.MySQLAdaptor;
+
+import java.util.Collection;
+
+public class PathwayDiagramGenerator {
+
+    private static final Logger logger = LogManager.getLogger();
+
+    private static MySQLAdaptor dba;
+    private static GKInstance speciesInst;
```
SolomonShorser-OICR: Is it necessary for speciesInst, personId, and referenceSpeciesId to be declared as static? I am not sure I understand why they are declared like this.

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+        logger.info("Generating diagrams for inferred " + speciesInst.getDisplayName() + "pathways");
+        // Create PredictedPathwayDiagramGeneratorFromDB object and set db adaptor and author ID.
+        PredictedPathwayDiagramGeneratorFromDB diagramGenerator = new PredictedPathwayDiagramGeneratorFromDB();
+        diagramGenerator.setMySQLAdaptor(dba);
+        diagramGenerator.setDefaultPersonId(personId);
+
+        GKInstance referenceSpeciesInst = dba.fetchInstance(referenceSpeciesId);
+
+        // Iterate through each PathwayDiagram instance looking for those associated with the reference species.
+        for (GKInstance diagramInst: (Collection<GKInstance>) dba.fetchInstancesByClass(ReactomeJavaConstants.PathwayDiagram)) {
```
SolomonShorser-OICR: There is a method, `generateDiagramsForOtherSpecies` inside `PredictedPathwayDiagramGeneratorFromDB` which generates diagrams for a given species. It looks like the code in `EventsInferrer` is passing a species to _this_ class, `PathwayDiagramGenerator`.  From my point of view, it looks like there is some duplication of functionality: Your code and the code in CuratorTool will generate diagrams on a per-species basis.

Is there a reason that you couldn't use `PredictedPathwayDiagramGeneratorFromDB.generateDiagramsForOtherSpecies`? If that method would work in this situation, it would simplify this PR significantly. If that method _cannot_ be used, could you add a brief note or comment explaining why?

orthoinference/src/test/java/org/reactome/orthoinference/PathwayDiagramGeneratorTest.java
```diff
@@ -0,0 +1,81 @@
...
+        Mockito.when(mockSpeciesInst.getDisplayName()).thenReturn("Species");
+        Mockito.when(mockAdaptor.fetchInstancesByClass(ReactomeJavaConstants.PathwayDiagram)).thenReturn(mockDiagramInstances);
+        Mockito.when(mockDiagramInst.getAttributeValue(ReactomeJavaConstants.representedPathway)).thenReturn(mockPathwayInst);
+        Mockito.when(mockPathwayInst.getAttributeValue(ReactomeJavaConstants.species)).thenReturn(mockSpeciesInst);
+        Mockito.when(mockPathwayInst.getAttributeValuesList(ReactomeJavaConstants.orthologousEvent)).thenReturn(mockOrthoEventInstances);
+        Mockito.when(mockOrthoEventInst.getAttributeValue(ReactomeJavaConstants.species)).thenReturn(mockSpeciesInst);
+        Mockito.when(mockOrthoEventInst.getAttributeValue(ReactomeJavaConstants.evidenceType)).thenReturn(mockEvidenceTypeInst);
+//        Mockito.when(mockDiagramGenerator.generatePredictedDiagram(mockOrthoEventInst, mockPathwayInst, mockDiagramInst)).thenReturn(mockOrthoDiagramInst);
+//
+        PathwayDiagramGenerator.generateOrthologousPathwayDiagrams();
```
SolomonShorser-OICR: Is there a way to assert that the diagrams were actually generated correctly? Maybe if `generateOrthologousPathwayDiagrams` returned some value, such as the number of diagrams that were generated, then this would be easier to verify?

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
cookersjs: Yea so that method is named a bit misleadingly. When you supply an id for an 'other' species, it will try to use that species as the reference. Since we always want Human as the reference, and since we want to only create diagrams for inferred Pathways in the target species of interest, we need to go a bit deeper. The generatePredictedDiagram method is the actual method that generates the orthologous diagram, but you need to first have the reference diagram, pathway and orthologous pathway instances.

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
cookersjs: Because the static concept is still a weird one to me -- I'll remove them

orthoinference/pom.xml
cookersjs: When I was testing the code outside orthoinference, there was a required jdom dependency in `PredictedPathwayDiagramGeneratorFromDB`

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
SolomonShorser-OICR: You can read more here: https://docs.oracle.com/javase/tutorial/java/javaOO/classvars.html
"static" is not a great word for the concept, I think.

A static class member will be shared by all instances of that class. For some things, such as  variables that access common resources (log files, databases, large caches of stuff), declaring them as static is a Good Idea. The variables speciesInst, personId, and referenceSpeciesId  could _potentially_ be different for each instance. So they probably should _not_ be static. 

One way to decide if something should be static is to ask: If two instances of _this_ class share the variable (if one instance modifies it, all others will see that modification), is that going to be a problem?

In this case, if you created multiple `PathwayDiagramGenerator` objects, the current code would force them to all share the values of speciesInst, personId, and referenceSpeciesId. This might be a problem if you wanted each on to create diagrams for different species, or using different reference species. The personId will _probably_ always be the same for all of them, but my personal rule is I usually do _not_ make things static, unless there is a good reason to, such as a shared database connection, or logger.

cookersjs: Something about the way I've written the code requires these instances to be static. I get the message 'non-static field cannot be referenced from a static context'. I think I might need to refactor a bit in order to get around this. We did something similar for StableIdentifierGenerator, and I don't think it was too much work. I'll look into doing that since I think we want to move towards that style

orthoinference/src/test/java/org/reactome/orthoinference/PathwayDiagramGeneratorTest.java
cookersjs: That would be quite nice, but I think since it is all mocked objects I would need a way to check that doesn't involve accessing a GKInstance object. I'll look into a better way of testing this so there is some assertion that can be made.

orthoinference/src/main/java/org/reactome/orthoinference/PathwayDiagramGenerator.java
SolomonShorser-OICR: > non-static field cannot be referenced from a static context

...means that you are trying to use an _instance_ member (declared without the "Static" modifier) in a _static_ method (the "static context" that the message is referring to). The method `generateOrthologousPathwayDiagrams` is a static method, and that is probably causing the problem. 

So the question is: why is `generateOrthologousPathwayDiagrams` a static method? A static method operates outside the context of a specific instance. In fact, a static method in Java is probably better referred to as a "function", since it does not operate in the context of any instance of its class.

In this specific case here, it probably makes sense to _remove_ the static modifier from `generateOrthologousPathwayDiagrams` and make it into a proper instance method. Removing the static modifier from personId, speciesInst, and referenceSpeciesInst also makes sense. Then you can create _instances_ of `PathwayDiagramGenerator` that generate diagrams for different species. Each object will then have its _own_ internal state describing which species, reference species, and person to use when creating diagrams.

An older article, but well-written:
https://javarevisited.blogspot.com/2013/07/when-to-make-method-static-in-java.html

See also this discussion, on when to use static for methods: https://stackoverflow.com/questions/2671496/java-when-to-use-static-methods

cookersjs: Thanks I'll look at both these articles and then I'll refactor a bit to make the method non-static

cookersjs: I've converted the method so that it is non-static now

orthoinference/README.md
```diff
@@ -2,29 +2,33 @@
...
+
+After all valid ReactionlikeEvents instances have been inferred for a species, the final step is to populate the <b>Pathway</b> instances these RlEs are found in. Pathway inference takes place in its entirety in the <a href="https://github.com/reactome/data-release-pipeline/blob/develop/orthoinference/src/main/java/org/reactome/orthoinference/HumanEventsUpdater.java">HumanEventsUpdater</a> class, and involves creating the hierarchy structure from the Human Pathway equivalent (this will be further expanded at a future date).
```
jweiser: Just wondering what it means when saying it will be expanded at a future date?  What needs to be added?

orthoinference/README.md
```diff
@@ -85,11 +90,10 @@ Once all scripts in the previous step have been run there is a QA process that s
...
 
-Next, we want to make sure that the new `test_reactome database` can be imported to the <b>graphDb</b> in neo4j and that it reports acceptable graph QA numbers. <br><b>It is recommended that the following steps be run on your workstation.</b>
+Next, we want to make sure that the new `release_current` database can be imported to the <b>graphDb</b> in neo4j and that it reports acceptable graph QA numbers. <br><br><b>It is recommended that the following steps be run on your workstation.</b>
 
-3) Run the <a href="https://github.com/reactome/graph-importer">graph-importer</a> module. This should take some time (30+ minutes) and will output the results from <a href="https://github.com/reactome/database-checker">database-checker</a> as well as the imported graphDb in `target/graph.db/`.
+2) Run the <a href="https://github.com/reactome/graph-importer">graph-importer</a> module. This should take some time (30ish minutes) and will output the results from <a href="https://github.com/reactome/database-checker">database-checker</a> as well as the imported graphDb in `target/graph.db/`.
```
jweiser: You could say approx. 30 minutes in place of 30ish minutes.

orthoinference/README.md
```diff
@@ -85,11 +90,10 @@ Once all scripts in the previous step have been run there is a QA process that s
...
-2) Inspect the `deleted_unsed_pe_test_reactome_(releaseNumber).txt` file that was output by the <a href="https://github.com/reactome/Release/blob/master/scripts/release/orthoinference/remove_unused_PE.pl">remove_unusued_PE.pl</a> step. It should be a one-liner stating that a considerable amount of PhysicalEntity instances were deleted. At time of writing (December 2018), the typical value was between 500,000 - 600,000 instances deleted.
 
-Next, we want to make sure that the new `test_reactome database` can be imported to the <b>graphDb</b> in neo4j and that it reports acceptable graph QA numbers. <br><b>It is recommended that the following steps be run on your workstation.</b>
+Next, we want to make sure that the new `release_current` database can be imported to the <b>graphDb</b> in neo4j and that it reports acceptable graph QA numbers. <br><br><b>It is recommended that the following steps be run on your workstation.</b>
```
jweiser: The "n" in Neo4j is usually capitalized

orthoinference/README.md
```diff
@@ -100,28 +104,22 @@ Next, we want to make sure that the new `test_reactome database` can be imported
...
 
-5) Once you are satisfied with the results from each of these steps, send the <b>graph-qa</b> and <b>database-checker</b> results to the Curator overseeing release. <a href="https://github.com/reactome/database-checker">Database-checker</a> can be re-run fairly painlessly by following the instructions on its github page. If the Curator is satisfied with the QA results, you can move onto the next step of release. At time of writing (December 2018) this is the <i>UpdateConfig</i> step. 
+4) Once you are satisfied with the results from each of these steps, send the <b>graph-qa</b> and <b>database-checker</b> results to the Curator overseeing release. <a href="https://github.com/reactome/database-checker">Database-checker</a> can be re-run fairly painlessly by following the instructions on its github page. If the Curator is satisfied with the QA results, you can move onto one of the next steps of release. At this point in the release pipeline that could be <a href="https://github.com/reactome/data-release-pipeline/tree/develop/update-dois">UpdateDOIs</a> or <a href="https://github.com/reactome/add-links">AddLinks</a>.
```
jweiser: GitHub can be capitalized in the sentence "...following the instructions on its github page".

orthoinference/README.md
```diff
@@ -100,28 +104,22 @@ Next, we want to make sure that the new `test_reactome database` can be imported
...
+-  The <b>WebELVTool</b> found in the <a href="https://github.com/reactome/CuratorTool">CuratorTool</a> repo can be used to check the validity of the `test_reactome` database. The WebELVTool jar is built by running an <b>ant</b> build on the <a href="https://github.com/reactome/CuratorTool/blob/master/ant/WebELVDiagram.xml">WebELVDiagram.xml</a> file. The jar file will appear in the `WebELVTool/` folder above the `CuratorTool/` folder. To run the program, navigate to the directory with the jar file and run:<br> `java -jar webELVTool.jar (host) (release_current) (mysql_username) (mysql_password) (mysql_port) (reactome_author_ID)`<br>The output should be many lines of 'Predicting pathway' or 'Working on pathway'. If the script runs successfully, then it is implied `test_reactome` is appropriately populated.
```
jweiser: You could say "test_reactome" or "release_current" database depending on the server.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -64,7 +64,7 @@
 						infReferenceGeneProductInst.addAttributeValue(species, speciesInst);
 						String referenceGeneProductSource = homologueSource.equals("ENSP") ? "ENSEMBL:" : "UniProt:";
 						infReferenceGeneProductInst.setAttributeValue(_displayName, referenceGeneProductSource + homologueId);
-						infReferenceGeneProductInst = InstanceUtilities.checkForIdenticalInstances(infReferenceGeneProductInst);
+						infReferenceGeneProductInst = InstanceUtilities.checkForIdenticalInstances(infReferenceGeneProductInst, null);
```
jweiser: It's a little unclear what the null parameter is here.  Maybe an overloaded `checkForIdenticalInstances` method that has a signature with just the first argument can be added in the InstanceUtilities class and it can call the existing method passing null as the second parameter.  That way you can pick the appropriate method signature depending on whether you need to pass one or two arguments.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -148,6 +150,7 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 			return;
 		}
 		String humanInstanceDbId = sourceSpeciesInst.iterator().next().getDBID().toString();
+		orthologousPathwayDiagramGenerator = new OrthologousPathwayDiagramGenerator(dbAdaptor, dbAdaptorPrev, speciesInst, personId, Long.valueOf(humanInstanceDbId));
```
jweiser: Is the `humanInstanceDbId` needed as specifically a String later on?  If not, maybe it would be better to have it be a `long` by removing the call to `.toString()` in the line before this and removing the need to parse it to a long value in this line.

orthoinference/src/test/java/org/reactome/orthoinference/OrthologousPathwayDiagramGeneratorTest.java
```diff
@@ -0,0 +1,95 @@
...
+    List<GKInstance> mockOrthoEventInstances = new ArrayList<>();
+
+
+    OrthologousPathwayDiagramGenerator testDiagramGenerator;
+
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+        testDiagramGenerator = new OrthologousPathwayDiagramGenerator(mockAdaptor, mockPrevAdaptor, mockSpeciesInst, 12345L, 12345L);
```
jweiser: Named constants to indicate what the last two parameters are would be good.

orthoinference/README.md
cookersjs: I've felt like the Pathway inference explanation is mostly just describing what happens as I see it in the code. I don't feel like I've explained much of the _why_ this is being done

orthoinference/README.md
cookersjs: Yea I had that until a few weeks ago, but I thought that since we had officially moved to the AWS server and switched to the current/previous notation that I should update the docs. I can add that back in

orthoinference/README.md
jweiser: Okay, got it.  We could work together on writing the reasoning for the pathway inference.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
cookersjs: Yea that is a cleaner way of doing it. I'll make that change

orthoinference/README.md
jweiser: Sounds good.  You could also say the "release database" if you want to be non-specific.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
cookersjs: This I think will be added in its own PR in the future (added it to release notes)

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: Yep good point

orthoinference/README.md
```diff
@@ -100,28 +104,22 @@ Next, we want to make sure that the new `test_reactome database` can be imported
...
+-  The <b>WebELVTool</b> found in the <a href="https://github.com/reactome/CuratorTool">CuratorTool</a> repo can be used to check the validity of the `release_current` database. The WebELVTool jar is built by running an <b>ant</b> build on the <a href="https://github.com/reactome/CuratorTool/blob/master/ant/WebELVDiagram.xml">WebELVDiagram.xml</a> file. The jar file will appear in the `WebELVTool/` folder above the `CuratorTool/` folder. To run the program, navigate to the directory with the jar file and run:<br> `java -jar webELVTool.jar (host) (release_current) (mysql_username) (mysql_password) (mysql_port) (reactome_author_ID)`<br>The output should be many lines of 'Predicting pathway' or 'Working on pathway'. If the script runs successfully, then it is implied `test_reactome` is appropriately populated.
```
jweiser: The "test_reactome" on the last line just needs to be changed to "release_current", I think?

cookersjs: Good catch, thanks

PR #52, opened 2019-04-29
Feature/update release current

cookersjs:
Creates `scripts` directory for general use scripts. Includes `updateDatabash.sh` which is the spiritual successor of `restore_database.pl`. This will archive and then update any database with an existing database dump


scripts/README.md
```diff
@@ -0,0 +1,21 @@
+<h2> General-purpose scripts for Reactome's release </h2>
```
SolomonShorser-OICR: Tip: Markdown has its own set of formatting instructions - you don't need to use HTML. Not necessary, but I find it is often slightly faster than HTML. Here's what github supports: https://guides.github.com/features/mastering-markdown/ (not all markdown renderers are the same).

scripts/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+This repository contains scripts that are used by multiple release steps.
+
+
+<h3>updateDatabase.sh</h3>
```
SolomonShorser-OICR: Is `updateDatabase.sh` the best name for it? I would expect that something with `update` in the name will update the data in an existing database. A drop-&-replace operation is _not_ what I expect from something named "updated". for this kind of script, I'd probably name it `replaceDatabase.sh`.

SolomonShorser-OICR: Mind you, if it could also _create_ a database, then maybe `replaceDatabase.sh` is also not the best name. Worst case: `createOrReplace.sh`, but if you can think of something better that's fine too.

scripts/updateDatabase.sh
```diff
@@ -0,0 +1,92 @@
...
+dbArchiveFile="$dbName.backup.dump"
+mysqldump -u$username -p$password -h$host -P$port $dbName > $dbArchiveFile
+echo "Compressing $dbArchiveFile";
+eval "gzip -f $dbArchiveFile";
+echo "Finished backing up $dbName";
+
+echo "Updating $dbName with $dbFilepath";
+mysql -u$username -p$password -h$host -P$port -e 'drop database if exists $dbName; create database $dbName'
+
+## If dump is gzipped, must use 'zcat'
```
SolomonShorser-OICR: Would it run faster if you unzip it and then `cat` the unzipped file, vs. using zcat? Is zcat always installed or should that be checked first?

cookersjs: I can't say that I've timed the two, though it doesn't take very long either way. And I know zcat is installed in base linux though I just found that there might be alternative behaviour for zcat on macs: https://serverfault.com/questions/570024/zcat-gzcat-works-in-linux-not-on-osx-general-linux-osx-compatibility

SolomonShorser-OICR: hmm ok probably not a problem to leave it this way for now since the _real_ release server runs linux (not OS X). BUT... if it's possible to find a way of calling zcat that is compatible for both platforms that would be nice to have, but not necessary right now.

cookersjs: Yea that makes sense. Alternatively I could just unzip the file if it has the *.gz pattern, I would probably just want to zip it up again at the end -- which would add a bit of time if it was a very large dump file

SolomonShorser-OICR: Hmm good point about re-zipping at the end. OK, I guess just leave it as is for now.

scripts/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+<b>Purpose</b>: Create or replace a mySQL database with a mysql dump file.
```
jweiser: Small nitpick - I think the capitalization is usually MySQL -- could that be updated through the PR?

scripts/config.properties.sample
```diff
@@ -0,0 +1,6 @@
+# Example config file for updateDatabase.sh
```
jweiser: Could the file name be changed, maybe to sample_config.properites, so .properties is the extension?

scripts/updateDatabase.sh
```diff
@@ -0,0 +1,92 @@
...
+## Parse config file for username and password properties
+echo "Reading $configFilepath";
+username=
+password=
+host=localhost
+port=3306
+while read line; do
+	if [[ $line == username* ]]
+ 	then
+ 		username=${line#*=}
```
jweiser: Just curious, what does the #*= do?

scripts/updateDatabase.sh
```diff
@@ -0,0 +1,92 @@
...
+	echo "Could not find username and/or password in $configFilepath";
+	exit 1
+fi
+
+echo "Updating $dbName with $dbFilepath";
+echo "Backing up $dbName";
+
+## Take archive of DB, drop it and create a new, empty one
+dbArchiveFile="$dbName.backup.dump"
+mysqldump -u$username -p$password -h$host -P$port $dbName > $dbArchiveFile
```
jweiser: I'm not sure if it's doable, but is there a way to do the mysqldump without specifying the password as an argument?  It's potentially visible to users of the system while it's running.

scripts/updateDatabase.sh
cookersjs: It is a weird one: https://stackoverflow.com/questions/41806725/what-is-the-meaning-of-1-in-bash

As I interpret it, the expression is three  parts, the `#`, `*` and `{}`. Together, they are saying find the pattern matching the `=`, trim anything preceding it and capture the rest. I think the `{}` surrounding the whole expression is what captures the remainder of the string. The 'drop anything preceding it' is apparently due to the `#` combined with the `*`. More technically, the # means drop _shortest_ preceding string, while having the operator `##` would mean drop the _longest_ preceding string.

cookersjs: The expected string by the way is username=mySQLUsername

scripts/updateDatabase.sh
cookersjs: Yea we had talked about doing this. I'll try and find a better way of doing this before I merge

jweiser: I think you can specify a my.cnf file containing the username and password to the mysqldump command that it will use to parse the credentials.  The updateDatabase.sh script could take the path to the my.cnf file as an argument.

https://stackoverflow.com/a/24369804

scripts/README.md
cookersjs: Yea I liked createOrReplace, so I changed the file names to that -- I'll update the README accordingly

scripts/updateDatabase.sh
cookersjs: Done

scripts/README.md
```diff
@@ -2,20 +2,22 @@
...
 <br>
 <b>Perl version</b>: <a href="https://github.com/reactome/Release/blob/master/scripts/restore_database.pl">restore_database.pl</a>
 <br>
 <b>Arguments</b>:
-- `-f|--file`  Filepath of mySQL dump file that will be restored
-- `-d|--database`  Name of mySQL database that will be created/updated
-- `-c|--config`  (Optional) Filepath of config file that contains mySQL parameters (username, password, database, port)
+- `-f|--file`  Filepath of MySQL dump file that will be restored
+- `-d|--database`  Name of MySQL database that will be created/updated
+- `-c|--config`  (Optional) Filepath of config file that contains mySQL parameters (username, database, port)
```
jweiser: If password isn't taken from the config file, it may be better not to allow username to be changed either?  Both should probably be configured in the ".my.cnf" file.

scripts/createOrReplaceDatabase.sh
```diff
@@ -59,24 +56,28 @@ while read line; do
...
 ## Take archive of DB, drop it and create a new, empty one
 dbArchiveFile="$dbName.backup.dump"
-mysqldump -u$username -p$password -h$host -P$port $dbName > $dbArchiveFile
-echo "Compressing $dbArchiveFile";
-eval "gzip -f $dbArchiveFile";
-echo "Finished backing up $dbName";
+if ! mysql -u$username -h$host -P$port -e "use $dbName" 2> /dev/null;
+then
+	echo "Creating $dbName";
+	mysql -u$username -h$host -P$port -e "create database $dbName"
```
jweiser: Is this branch necessary since the db will get created on line 80?

scripts/createOrReplaceDatabase.sh
```diff
@@ -0,0 +1,93 @@
...
+## Improperly formatted or missing config file
+if [ -z "$username" ]
+then
+	echo "No username in $configFilepath";
+	exit 1
+fi
+
+## Take archive of DB, drop it and create a new, empty one
+dbArchiveFile="$dbName.backup.dump"
+if ! mysql -u$username -h$host -P$port -e "use $dbName" 2> /dev/null;
```
SolomonShorser-OICR: What is this trying to do? Check that a database exists?  I think `mysqlshow -u$username -p | grep $dbName` might be a little clearer.

scripts/createOrReplaceDatabase.sh
cookersjs: It was throwing a MySQL error when it was trying to a backup of a non-existent DB, so I had to put that branch in

SolomonShorser-OICR: Could the logic just be rewritten as 
```
if $databaseDOESExists then
  # do backup stuff...
fi
```

scripts/README.md
cookersjs: Yea I thought to do that but wasn't sure if there was some situation where you might want to be a different user... but that doesn't make sense if your using a specific user's password, so I'll make that change

scripts/createOrReplaceDatabase.sh
jweiser: Right, but you could reverse the condition for controlling when you take a backup and remove this branch for creating a db since that happens later on line 80.

scripts/README.md
cookersjs: Done

scripts/createOrReplaceDatabase.sh
cookersjs: I swapped the logic to remove the `!` part. I think it looks more readable now

scripts/createOrReplaceDatabase.sh
cookersjs: With the configuration stuff removed and the conditional swapped, now it is just `if mysql -e "use db" 2> /dev/null`, which is a bit more readable. I am hesitant to add a third parameter to the `.my.cnf` file (for [mysqlshow]), only cause that adds more configuration

scripts/README.md
```diff
@@ -2,22 +2,19 @@
 
 This repository contains scripts that are used by multiple release steps.
 
-<b>Note:</b> Make sure you have a `~/.my.cnf` file populated with `mysqldump` and `mysql` parameters to use this script. See https://stackoverflow.com/a/24369804
+<h3>createOrReplaceDatabase.sh</h3>
 
 
-<h3>createOrReplaceDatabase.sh</h3>
+<b>Note:</b> Make sure you have a `.my.cnf` file populated, in your <b>home directory</b>, with `mysqldump` and `mysql` parameters to use this script. See https://stackoverflow.com/a/24369804
```
jweiser: Do you want to mention that the `my.cnf` parameters accepted include "user" and "password" with optional "host" and "port"?

PR #53, opened 2019-04-29
Feature/data model changes for go

SolomonShorser-OICR:
Sync GO Update with the latest data model changes for GO_* entities.
Also, some updates for report formats.

...might was well start reviewing. If Lisa wants changes to the reports, then that will just require a re-review.

### DIFF
--- a/downloadDirectory/pom.xml
+++ b/downloadDirectory/pom.xml
@@ -1,6 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>org.reactome.release</groupId>
@@ -109,24 +107,6 @@
 
 	<build>
 		<plugins>
-			<plugin>
-				<artifactId>exec-maven-plugin</artifactId>
-				<groupId>org.codehaus.mojo</groupId>
-				<version>1.6.0</version>
-				<executions>
-					<execution>
-						<id>Build PathwayExchange</id>
-						<phase>generate-sources</phase>
-						<goals>
-							<goal>exec</goal>
-						</goals>
-						<configuration>
-							<executable>build_pathway_exchange.sh</executable>
-						</configuration>
-					</execution>
-				</executions>
-			</plugin>
-
 			<plugin>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.7.0</version>
--- a/downloadDirectory/pom.xml
+++ b/downloadDirectory/pom.xml
@@ -1,5 +1,6 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>org.reactome.release</groupId>
@@ -79,7 +80,7 @@
 			<artifactId>jdom</artifactId>
 			<version>1.1</version>
 		</dependency>
-			<dependency>
+		<dependency>
 			<groupId>org.eclipse.jgit</groupId>
 			<artifactId>org.eclipse.jgit</artifactId>
 			<version>4.8.0.201706111038-r</version>
@@ -89,7 +90,7 @@
 			<artifactId>mysql-connector-java</artifactId>
 			<version>5.1.47</version>
 		</dependency>
-			<dependency>
+		<dependency>
 			<groupId>org.apache.logging.log4j</groupId>
 			<artifactId>log4j-core</artifactId>
 			<version>2.11.0</version>
@@ -99,7 +100,7 @@
 			<artifactId>commons-io</artifactId>
 			<version>2.4</version>
 		</dependency>
-			<dependency>
+		<dependency>
 			<groupId>com.googlecode.json-simple</groupId>
 			<artifactId>json-simple</artifactId>
 			<version>1.1.1</version>
@@ -109,12 +110,30 @@
 	<build>
 		<plugins>
 			<plugin>
+				<artifactId>exec-maven-plugin</artifactId>
+				<groupId>org.codehaus.mojo</groupId>
+				<version>1.6.0</version>
+				<executions>
+					<execution>
+						<id>Build PathwayExchange</id>
+						<phase>generate-sources</phase>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+						<configuration>
+							<executable>build_pathway_exchange.sh</executable>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+
+			<plugin>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.7.0</version>
 				<configuration>
-				<!--<release>8</release>-->
-				<source>8</source>
-				<target>8</target>
+					<!--<release>8</release> -->
+					<source>8</source>
+					<target>8</target>
 				</configuration>
 			</plugin>
 			<plugin>
@@ -123,11 +142,11 @@
 				<version>3.0.2</version>
 				<configuration>
 					<archive>
-					<manifest>
-						<mainClass>org.reactome.release.downloadDirectory.Main</mainClass>
-						<addClasspath>true</addClasspath>
-						<classpathPrefix>lib/</classpathPrefix>
-					</manifest>
+						<manifest>
+							<mainClass>org.reactome.release.downloadDirectory.Main</mainClass>
+							<addClasspath>true</addClasspath>
+							<classpathPrefix>lib/</classpathPrefix>
+						</manifest>
 					</archive>
 					<finalName>downloadDirectory</finalName>
 				</configuration>
--- a/go-update/pom.xml
+++ b/go-update/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.reactome.release</groupId>
 	<artifactId>go-update</artifactId>
-	<version>0.0.1-SNAPSHOT</version>
+	<version>1.0.0</version>
 	<name>GOUpdate</name>
 	<description>This program will update GO data.</description>
 	<properties>
--- a/go-update/pom.xml
+++ b/go-update/pom.xml
@@ -59,6 +59,11 @@
 			<version>${powermock.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-csv</artifactId>
+			<version>1.6</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
@@ -117,13 +122,5 @@
 				<enabled>true</enabled>
 			</snapshots>
 		</repository>
-		<!-- <repository>
-			<name>EBI</name>
-			<url>http://www.ebi.ac.uk/Tools/maven/repos/content/repositories/releases/</url>
-			<id>EBI</id>
-			<releases>
-				<enabled>true</enabled>
-			</releases>
-		</repository> -->
 	</repositories>
 </project>
--- a/go-update/pom.xml
+++ b/go-update/pom.xml
@@ -27,7 +27,7 @@
 		<dependency>
 			<groupId>org.reactome.base</groupId>
 			<artifactId>reactome-base</artifactId>
-			<version>RELEASE</version>
+			<version>1.2.20</version>
 		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
@@ -117,13 +117,13 @@
 				<enabled>true</enabled>
 			</snapshots>
 		</repository>
-		<repository>
+		<!-- <repository>
 			<name>EBI</name>
 			<url>http://www.ebi.ac.uk/Tools/maven/repos/content/repositories/releases/</url>
 			<id>EBI</id>
 			<releases>
 				<enabled>true</enabled>
 			</releases>
-		</repository>
+		</repository> -->
 	</repositories>
 </project>
--- a/go-update/pom.xml
+++ b/go-update/pom.xml
@@ -27,7 +27,7 @@
 		<dependency>
 			<groupId>org.reactome.base</groupId>
 			<artifactId>reactome-base</artifactId>
-			<version>1.2.20</version>
+			<version>1.2.21</version>
 		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
--- a/go-update/pom.xml
+++ b/go-update/pom.xml
@@ -11,7 +11,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<powermock.version>2.0.0-beta.5</powermock.version>
+		<powermock.version>2.0.0</powermock.version>
 	</properties>
 	<dependencies>
 		<dependency>
--- a/go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
+++ b/go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
@@ -53,13 +53,13 @@ public class DuplicateReporter
 	{
 		Map<String, Integer> duplicateAccessions = new HashMap<>();
 		
-		ResultSet results = this.adaptor.executeQuery(DuplicateReporter.duplicateAccessionQuery, null);
-		
-		while (results.next())
+		try(ResultSet results = this.adaptor.executeQuery(DuplicateReporter.duplicateAccessionQuery, null);)
 		{
-			duplicateAccessions.put(results.getString("accession"), results.getInt("instance_count"));
+			while (results.next())
+			{
+				duplicateAccessions.put(results.getString("accession"), results.getInt("instance_count"));
+			}
 		}
-		
 		return duplicateAccessions;
 	}
 	
@@ -86,13 +86,21 @@ public class DuplicateReporter
 				{
 					long dbid = i.getDBID();
 					int refCount = getReferrerCountforInstance(i, classesToIgnore);
-					referrerCounts.put(dbid,refCount);
+					referrerCounts.put(dbid, refCount);
 				}
 			}
 		}
 		return referrerCounts;
 	}
 	
+	// TODO: Move this function to release-common-lib, maybe. It's a generic utility function, could be useful somewhere else.
+	/**
+	 * Gets the number of referrers for an instance. That is, it returns how many objects refer to <code>instance</code>.
+	 * @param instance - the object to query about.
+	 * @param classesToIgnore - a list of class names to ignore. Referrers whose SchemaClass name match a name in this list will *not* be included in the final count.
+	 * @return
+	 * @throws Exception
+	 */
 	public int getReferrerCountforInstance(GKInstance instance, String ... classesToIgnore) throws Exception
 	{
 		int refCount = 0;
--- a/go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
+++ b/go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
@@ -30,7 +30,7 @@ class GoLineProcessor
 	 * @param currentGOID - The ID of the GO term currently being processed, line by line.
 	 * @return The GO Accession of the GO term currently being processed. Will be different from <code>currentGOID</code> if the ID for a new GO term is seen on this <code>line</code>
 	 */
-	static String processLine(String line, String currentGOID, Map<String, Map<String,Object>> goTerms)
+	static String processLine(String line, String currentGOID, Map<String, Map<String, Object>> goTerms)
 	{
 		String goID = currentGOID;
 		Matcher m;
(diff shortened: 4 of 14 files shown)


go-update/src/main/java/org/reactome/release/goupdate/DuplicateReporter.java
```diff
@@ -86,13 +86,21 @@ public DuplicateReporter(MySQLAdaptor adaptor)
...
 					int refCount = getReferrerCountforInstance(i, classesToIgnore);
-					referrerCounts.put(dbid,refCount);
+					referrerCounts.put(dbid, refCount);
 				}
 			}
 		}
 		return referrerCounts;
 	}
 	
+	// TODO: Move this function to release-common-lib, maybe. It's a generic utility function, could be useful somewhere else.
```
cookersjs: I can't off the top of my head think of any modules that try to get a referrer count so it might be OK here

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
```diff
@@ -162,21 +162,6 @@ static String processLine(String line, String currentGOID, Map<String, Map<Strin
...
-								case GoUpdateConstants.POSITIVELY_REGULATES:
-								{
-									addToMultivaluedAttribute(goTerms, currentGOID, line, GoUpdateConstants.RELATIONSHIP_POSITIVELY_REGULATES_REGEX, GoUpdateConstants.POSITIVELY_REGULATES);
-									break;
-								}
-								case GoUpdateConstants.NEGATIVELY_REGULATES:
-								{
-									addToMultivaluedAttribute(goTerms, currentGOID, line, GoUpdateConstants.RELATIONSHIP_NEGATIVELY_REGULATES_REGEX, GoUpdateConstants.NEGATIVELY_REGULATES);
-									break;
-								}
```
cookersjs: Since these keywords are associated with regulation, are these being removed due to data model changes?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -135,20 +135,33 @@ public void updateGOInstance(Map<String, Map<String, Object>> goTerms, Map<Strin
...
-					this.goInstance.setAttributeValue(ReactomeJavaConstants.componentOf, null);
-					this.adaptor.updateInstanceAttribute(this.goInstance, ReactomeJavaConstants.componentOf);
-					this.goInstance.setAttributeValue(ReactomeJavaConstants.name, goTerms.get(currentGOID).get(GoUpdateConstants.NAME));
-					this.adaptor.updateInstanceAttribute(this.goInstance, ReactomeJavaConstants.name);
-					this.goInstance.setAttributeValue(ReactomeJavaConstants.definition, newDefinition);
-					this.adaptor.updateInstanceAttribute(this.goInstance, ReactomeJavaConstants.definition);
+					// Changes for name
+					if (newName!=null && !newName.equals(oldName))
+					{
+						String nameUpdate = newName.equals(oldName) ? "" : "\n\tNew name:\t\""+newName+"\"\n\told name:\t\""+this.goInstance.getAttributeValue(ReactomeJavaConstants.name)+"\"";
```
cookersjs: Isn't the first condition of this ternary not possible since the preceding if statement filters against newName equaling oldName?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -135,20 +135,33 @@ public void updateGOInstance(Map<String, Map<String, Object>> goTerms, Map<Strin
...
+						nameOrDefinitionChangeStringBuilder.append("\nChange in name/definition for GO:").append(currentGOID).append(nameUpdate);
+						this.goInstance.setAttributeValue(ReactomeJavaConstants.name, newName);
+						this.adaptor.updateInstanceAttribute(this.goInstance, ReactomeJavaConstants.name);
+					}
+					// Changes for definition  
+					if (newDefinition != null && !newDefinition.equals(oldDefinition))
+					{
+						String defnUpdate = newDefinition.equals(oldDefinition) ? "" : "\n\tNew def'n:\t\""+newDefinition+"\"\n\told def'n:\t\""+this.goInstance.getAttributeValue(ReactomeJavaConstants.definition)+"\"";
```
cookersjs: Same as above comment but for definition

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -135,20 +135,33 @@ public void updateGOInstance(Map<String, Map<String, Object>> goTerms, Map<Strin
...
+						nameOrDefinitionChangeStringBuilder.append("\nChange in name/definition for GO:").append(currentGOID).append(defnUpdate);
+						this.goInstance.setAttributeValue(ReactomeJavaConstants.definition, newDefinition);
+						this.adaptor.updateInstanceAttribute(this.goInstance, ReactomeJavaConstants.definition);
+					}
+					// Now, instanceOf and componentOf are *OLNLY* valid for GO_CellularComponent
```
cookersjs: *ONLY* typo

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -238,17 +254,17 @@ public void updateReferrersDisplayNames() throws Exception
 		
 		if (instance.getSchemClass().getName().equals(ReactomeJavaConstants.GO_BiologicalProcess))
```
cookersjs: This *might* be made shorter if it was `instance.getSchemClass().isa(ReactomeJavaConstants.GO_BiologicalProcess)`, assuming that GO_BiologicalProcess/CellularComponent/MolecularFunction are always GKInstances

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -362,10 +379,27 @@ private void pointAllReferrersToOtherInstance(GKInstance replacementGOTerm) thro
...
+						referrer.setAttributeValue(attributeName, referrerAttributeValues);
+					}
+					else
+					{
+						referrer.setAttributeValue(attributeName, replacementGOTerm);
+					}
+					// The old Perl code would update referrers' displayNames if they were PhysicalEntities or CatalystActivities.
+					if (referrer.getSchemClass().isa(ReactomeJavaConstants.PhysicalEntity) || referrer.getSchemClass().isa(ReactomeJavaConstants.CatalystActivity))
+					{
+						String newReferrerDisplayName = InstanceDisplayNameGenerator.generateDisplayName(referrer);
```
cookersjs: Is InstanceDisplayNameGenerator in release-common-lib? This would be useful for Orthoinference

go-update/src/main/java/org/reactome/release/goupdate/GoTermsReconciler.java
```diff
@@ -0,0 +1,200 @@
...
+						instancesOfs = (Collection<GKInstance>) instance.getAttributeValuesList(ReactomeJavaConstants.instanceOf);
+						partOfs = (Collection<GKInstance>) instance.getAttributeValuesList(ReactomeJavaConstants.componentOf);
+						hasParts = (Collection<GKInstance>) instance.getAttributeValuesList("hasPart");
+					}
+
+					for (String k : goTerm.keySet())
+					{
+						switch (k)
+						{
+							case GoUpdateConstants.DEF:
```
cookersjs: Comments for each of these cases, I'm not really sure what it means to be a 'DEF'

go-update/src/main/java/org/reactome/release/goupdate/GoTermsReconciler.java
```diff
@@ -0,0 +1,200 @@
...
+	 * @throws InvalidAttributeException
+	 * @throws Exception
+	 */
+	private static void reconcileRelationship(String goAccession, Map<String, Object> goTerm, Collection<GKInstance> relationInstances, String relationship) throws InvalidAttributeException, Exception {
+		boolean found = false;
+		@SuppressWarnings("unchecked")
+		List<String> relationAccessionsFromFile = (List<String>) goTerm.get(relationship);
+		for (String relationAccessionFromFile: relationAccessionsFromFile)
+		{
+			for (GKInstance i : relationInstances)
```
cookersjs: variable could be named relationInstance

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
SolomonShorser-OICR: Those relationships are no longer being used in the Reactome data model.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: Ummm... yeah I'm gonna look at that. ;)

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: Yes, I think "isa" is a beter way to handle this, in general.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: `InstanceDisplayNameGenerator` is in reactome-base, I believe. It originates somewhere in the CuratorTool.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -137,57 +145,72 @@ else if (termStarted)
...
+			else // update existing instance. Including Obsolete instances, because if they can't be deleted (for some reason) they should be updated so their name/def'n indicates their obsolescence.
+			{
+				// Try to update each goInstance that has the current GO ID.
+				for (GKInstance goInst : goInstances)
 				{
-					// Try to update each goInstance that has the current GO ID.
-					for (GKInstance goInst : goInstances)
+					// Compartment is a sub-class of GO_CellularComponent - but the GO namespaces don't seem to account for that,
+					// we we'll account for that here.
+					boolean categoryOK = goInst.getSchemClass().getName().equals(currentCategory.getReactomeName())
```
cookersjs: Since boolean determination is subject to a variety of conditions, maybe there should be a method for it?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: You can see the code here: https://github.com/reactome/CuratorTool/blob/master/src/org/gk/model/InstanceDisplayNameGenerator.java

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -319,236 +255,272 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.IS_OBSOLETE) &&
...
-	protected static Map<GKSchemaAttribute, Integer> getReferrerCounts(GKInstance inst) throws Exception
+	/**
+	 * Updates the relationships of GO terms.
+	 * @param goTermsFromFile - the GO terms from the GO file.
+	 * @param allGoInstances - a map of ALL GO instances from the database.
+	 * @throws InvalidAttributeException
+	 * @throws Exception
+	 * @throws InvalidAttributeValueException
+	 */
+	private void updateRelationships(Map<String, Map<String, Object>> goTermsFromFile, Map<String, List<GKInstance>> allGoInstances) throws InvalidAttributeException, Exception, InvalidAttributeValueException
```
cookersjs: Is there any reason there are multiple Exception types? Since they all fall under Exception

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -319,236 +255,272 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.IS_OBSOLETE) &&
...
+					GoTermInstanceModifier goModifier = new GoTermInstanceModifier(this.adaptor, goInst, this.instanceEdit);
+					
+					if (goInst.getSchemClass().isa(ReactomeJavaConstants.GO_CellularComponent))
+					{
+						goModifier.updateRelationship(allGoInstances, goProps, GoUpdateConstants.IS_A, ReactomeJavaConstants.instanceOf);
+						goModifier.updateRelationship(allGoInstances, goProps, GoUpdateConstants.HAS_PART, "hasPart");
+						goModifier.updateRelationship(allGoInstances, goProps, GoUpdateConstants.PART_OF, ReactomeJavaConstants.componentOf);
+					}
+					// Update the instanace's "modififed".
+					goInst.getAttributeValuesList(ReactomeJavaConstants.modified);
```
cookersjs: I don't think this `getAttributeValuesList` is needed

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -319,236 +255,272 @@ else if (goTermsFromFile.get(goID).containsKey(GoUpdateConstants.IS_OBSOLETE) &&
...
-			Collection<GKInstance> instances = this.adaptor.fetchInstanceByAttribute( ((GONamespace)goTerm.get(GoUpdateConstants.NAMESPACE)).getReactomeName(), ReactomeJavaConstants.accession, "=", goAccession );
-			if (instances != null)
-			{
-				if (instances.size()>1)
-				{
-					reconciliationLogger.warn("GO Accession {} appears {} times in the database. It should probably only appear once.",goAccession, instances.size());
-				}
-				for (GKInstance instance : instances)
+			instancesForDeletion.addAll(goInstances);
+			attemptToDeleteObsoleteMessage.append(" Replacement Accesion: ").append(goTermsFromFile.get(goID).get(GoUpdateConstants.REPLACED_BY));
```
cookersjs: Accession typo

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
SolomonShorser-OICR: I don't like to hide more specific exceptions behind `Exception`. All three are listed because all three could be thrown by different parts of this function. I could simplify this by only throwing `Exception`, but then that could be misleading to future users of this function.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
SolomonShorser-OICR: I'm not always 100% sure with the Reactome Java API. In this case, I want to append to the existing values that are already in the database. I recall running into trouble once where I *didn't* retrieve the list prior to appending and then somehow wiped out the list. So that's probably why I did it this way.

go-update/src/main/java/org/reactome/release/goupdate/GoUpdateStep.java
```diff
@@ -50,19 +51,24 @@ public void executeStep(Properties props) throws SQLException
...
+			// ***UPDATE***
+			// URL to main GO file is now:
+			// http://current.geneontology.org/ontology/go.obo
+			// ...the ec2go file is the same.
+			
 			MySQLAdaptor adaptor = getMySQLAdaptorFromProperties(props);
 			this.loadTestModeFromProperties(props);
 			
-			long personID = new Long(props.getProperty("person.id"));
+			long personID = Long.valueOf(props.getProperty("person.id")).longValue();
```
cookersjs: Does the `.longValue()` do something that `Long.valueOf()` doesnt?

SolomonShorser-OICR: `Long.valueOf(props.getProperty("person.id")).longValue();` ...in this expression, `Long.valueOf` will return a `Long` whose value is the numeric representation of the string that was passed to it. `longValue()`, when called on a `Long` will return a `long`.

Yes, `long` and `Long` are different. `long` is a primitive, and `Long` is an Object. Primitives are passed by value, objects are passed by reference. Objects can be null, primitives _cannot_ be null. 

More on primitives here: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html
More on numeric versions of primitives here: https://docs.oracle.com/javase/tutorial/java/data/numberclasses.html

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
cookersjs: Ah true, I was wondering if that was the purpose of it. I don't know enough to prove or disprove this

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -177,10 +177,14 @@ else if (termStarted)
 				{
 					// Compartment is a sub-class of GO_CellularComponent - but the GO namespaces don't seem to account for that,
 					// we we'll account for that here.
-					boolean categoryOK = goInst.getSchemClass().getName().equals(currentCategory.getReactomeName()) 
-										|| ( (goInst.getSchemClass().getName().equals(ReactomeJavaConstants.Compartment) || goInst.getSchemClass().getName().equals(ReactomeJavaConstants.EntityCompartment) )
-												&& currentCategory.getReactomeName().equals(ReactomeJavaConstants.GO_CellularComponent) );
-					
+					boolean isCellularComponentOrSubclass = (goInst.getSchemClass().getName().equals(ReactomeJavaConstants.Compartment)
```
cookersjs: This could use the `isa()` function as well

downloadDirectory/pom.xml
```diff
@@ -108,13 +109,31 @@
...
+				<version>1.6.0</version>
+				<executions>
+					<execution>
+						<id>Build PathwayExchange</id>
+						<phase>generate-sources</phase>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+						<configuration>
+							<executable>build_pathway_exchange.sh</executable>
```
jweiser: This looks intriguing!  Does it run the shell script as a build pre-requisite?

SolomonShorser-OICR: Oops... this POM wasn't supposed to be checked in. I will remove these changes.

SolomonShorser-OICR: But yes, you can execute shell scripts as a part of a Maven build. Read up on the exec-maven-plugin to learn more.

go-update/pom.xml
```diff
@@ -117,13 +117,13 @@
 				<enabled>true</enabled>
 			</snapshots>
 		</repository>
-		<repository>
+		<!-- <repository>
```
jweiser: Why is this repository commented out?

go-update/src/main/java/org/reactome/release/goupdate/GoLineProcessor.java
```diff
@@ -30,7 +30,7 @@ private GoLineProcessor() {}
 	 * @param currentGOID - The ID of the GO term currently being processed, line by line.
 	 * @return The GO Accession of the GO term currently being processed. Will be different from <code>currentGOID</code> if the ID for a new GO term is seen on this <code>line</code>
 	 */
-	static String processLine(String line, String currentGOID, Map<String, Map<String,Object>> goTerms)
+	static String processLine(String line, String currentGOID, Map<String, Map<String, Object>> goTerms)
```
jweiser: Should the param annotations for goTerms and currentGOID be switched so they are in the same order as the method signature?  Also, it'd be good for the description to elaborate a bit about what the map of maps hold in terms of their keys and values.

go-update/pom.xml
SolomonShorser-OICR: I think that repository was causing maven to get an "Unauthorized" error when it tried to contact it.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsReconciler.java
jweiser: DEF is GO's shorthand, I think, for definition.  Maybe the value that the constant holds could be as is, but the constant name could be "DEFINITION"?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -359,13 +363,76 @@ else if (GoTermsUpdater.getReferrerCountsExcludingGOEntities(this.goInstance).is
...
 	}
+
+	/*
+	 * Clears reference attributes that point TO *this* goInstance FROM other GO entities. To be used when an instance is being deleted.
+	 */
+	private void clearAttributesFromReferringGOEntities() throws Exception
+	{
+		Map<GKSchemaAttribute, Integer> goReferrerCounts = GoTermsUpdater.getReferrerCountsFilteredByClass(this.goInstance, GoTermsUpdater.isNotGOEntity.negate());
+		// If there are any referrers that are GO entities...
+		if (goReferrerCounts.size() > 0)
```
jweiser: I don't think this check is necessary.  If the size of `goReferrerCounts` is zero, the call to the keySet method will return an empty set on line 376 and the for loop won't be entered.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
```diff
@@ -58,7 +58,7 @@
 	// this can be static, since there's only one "GO" ReferenceDatabase object in the database.
 	private static GKInstance goRefDB;
 	
-	private static Predicate<GKInstance> isNotGOEntity = i -> !i.getSchemClass().isa(ReactomeJavaConstants.GO_MolecularFunction)
+	static Predicate<GKInstance> isNotGOEntity = i -> !i.getSchemClass().isa(ReactomeJavaConstants.GO_MolecularFunction)
```
jweiser: Just curious why "private" was removed here?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -499,5 +561,14 @@ public void updateRelationship(Map<String, List<GKInstance>> allGoInstances, Map
...
-
+	
+	private static String abbreviate(String s, int maxLength)
+	{
+		return s.substring(0,Math.min(s.length(), maxLength)) + ( s.length() > maxLength ? "..." : "" );
+	}
+	
+	private static String abbreviate(String s)
+	{
+		return abbreviate(s, 47);
```
jweiser: Why 47?  This looks like a magic number - could it be given a named constant?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -359,13 +363,76 @@ else if (GoTermsUpdater.getReferrerCountsExcludingGOEntities(this.goInstance).is
...
+							this.adaptor.updateInstanceAttribute(attribReferrer, attrib.getName());
+						}
+						// now that the references to *this* GO Instance have been removed, record this operation by adding a "modified" InstanceEdit.
+						attribReferrer.getAttributeValuesList(ReactomeJavaConstants.modified);
+						attribReferrer.addAttributeValue(ReactomeJavaConstants.modified, this.instanceEdit);
+						this.adaptor.updateInstanceAttribute(attribReferrer, ReactomeJavaConstants.modified);
+					}
+					catch (Exception  e)
+					{
+						logger.error("Error trying to clear {} attribute on \"{}\", referring to \"{}\" (which is to be deleted).", attrib.getName(), abbreviate(attribReferrer.toString(), 55), abbreviate(this.goInstance.toString(), 55));
```
jweiser: Why 55 for the max length?

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
```diff
@@ -359,13 +363,76 @@ else if (GoTermsUpdater.getReferrerCountsExcludingGOEntities(this.goInstance).is
...
+		{
+			for (GKSchemaAttribute attrib : goReferrerCounts.keySet())
+			{
+				// set the referring attributes to NULL so that we don't end up with "dangling pointers" in the database.
+				Collection<GKInstance> attribReferrers = (Collection<GKInstance>) this.goInstance.getReferers(attrib);
+				for (GKInstance attribReferrer : attribReferrers)
+				{
+					try
+					{
+						updatedGOTermLogger.info("CLEARING the attribute {} on \"{}\" because it refers to \"{}\", which is flagged for deletion.", attrib.getName(), abbreviate(attribReferrer.toString(), 55), abbreviate(this.goInstance.toString(), 55));
```
jweiser: Why 55 for the max length?

go-update/src/main/java/org/reactome/release/goupdate/GoTermsUpdater.java
SolomonShorser-OICR: I wanted to access the predicate in another class, so I made it package-scoped.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: Originally, I wanted to limit to 50 characters, for readability. Because there is a lot of "header" such as "[GO_BiologicalProcess:1234] The blah blah blah..." it seemed that 50 characters gave good results. 47 is to make sure that _exactly_ 50 characters would be used because the last three characters would be: "..." ;)

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: Hmm probably.

go-update/src/main/java/org/reactome/release/goupdate/GoTermInstanceModifier.java
SolomonShorser-OICR: I think that was when I was still trying to determine optimal length for abbreviated strings.

go-update/src/main/java/org/reactome/release/goupdate/GoTermsReconciler.java
SolomonShorser-OICR: I'd prefer to keep it consistent with GO's terminology, in this case.

jweiser: Maybe a comment to clarify the abbreviation then?

SolomonShorser-OICR: Javadoc has been added to the appropriate constant to explain the meaning of DEF.

PR #55, opened 2019-05-07
Feature/europe pmc updates

jweiser:
updates the description of the Reactome project and the URLs we are exporting to Europe PMC (content page instead of pathway browser)

### DIFF
--- a/data-exporter/pom.xml
+++ b/data-exporter/pom.xml
@@ -5,7 +5,7 @@
 	<groupId>org.reactome.release</groupId>
 	<artifactId>data-exporter</artifactId>
 	<packaging>jar</packaging>
-	<version>1.1.0-SNAPSHOT</version>
+	<version>1.1.0</version>
 	<name>data-exporter</name>
 
 	<properties>
--- a/data-exporter/pom.xml
+++ b/data-exporter/pom.xml
@@ -5,7 +5,7 @@
 	<groupId>org.reactome.release</groupId>
 	<artifactId>data-exporter</artifactId>
 	<packaging>jar</packaging>
-	<version>1.0-SNAPSHOT</version>
+	<version>1.1.0-SNAPSHOT</version>
 	<name>data-exporter</name>
 
 	<properties>
--- a/data-exporter/src/main/java/org/reactome/release/dataexport/EuropePMC.java
+++ b/data-exporter/src/main/java/org/reactome/release/dataexport/EuropePMC.java
@@ -124,10 +124,8 @@ public class EuropePMC {
 		nodeNameToValue.put("id", Integer.toString(reactomeProviderID));
 		nodeNameToValue.put("resourceName", "Reactome");
 		nodeNameToValue.put("description",
-			"Reactome is a free, open-source, curated and peer-reviewed pathway database. " +
-			"Our goal is to provide intuitive bioinformatics tools for the visualization, " +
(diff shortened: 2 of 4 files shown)


data-exporter/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<groupId>org.reactome.release</groupId>
 	<artifactId>data-exporter</artifactId>
 	<packaging>jar</packaging>
-	<version>1.0-SNAPSHOT</version>
+	<version>1.1.0-SNAPSHOT</version>
```
cookersjs: We should probably start moving away from the string `SNAPSHOT` being in our versioning, all my new repos have it as well

data-exporter/src/main/java/org/reactome/release/dataexport/EuropePMC.java
```diff
@@ -124,10 +124,8 @@ private String getEuropePMCProfileXML() {
 		nodeNameToValue.put("id", Integer.toString(reactomeProviderID));
 		nodeNameToValue.put("resourceName", "Reactome");
 		nodeNameToValue.put("description",
-			"Reactome is a free, open-source, curated and peer-reviewed pathway database. " +
-			"Our goal is to provide intuitive bioinformatics tools for the visualization, " +
-			"interpretation and analysis of pathway knowledge to support basic research, genome analysis, " +
-			"modeling, systems biology and education.");
+			"Reactome is a free, open-source, curated and peer-reviewed pathway database."
```
cookersjs: Did we lose this the mission statement?

data-exporter/src/test/java/org/reactome/release/dataexport/EuropePMCTest.java
```diff
@@ -47,7 +47,7 @@ public void singleEuropePMCLinkFromMockGraphDB() {
 	public void correctEuropePMCLinkXMLElement() throws ParserConfigurationException {
 		final String PATHWAY_DISPLAY_NAME = "p53-Dependent G1 DNA Damage Response";
 		final String PATHWAY_STABLE_ID = "R-HSA-69563";
-		final String PATHWAY_URL = "https://www.reactome.org/PathwayBrowser/#/" + PATHWAY_STABLE_ID;
+		final String PATHWAY_URL = "https://reactome.org/content/detail/" + PATHWAY_STABLE_ID;
```
cookersjs: Could this become `ReactomeURLConstants.CONTENT_DETAIL_URL + PATHWAY_STABLE_ID`?

data-exporter/pom.xml
jweiser: Yeah, I'll remove it seeing is this is a finished minor update.  I had to refresh my memory on when to use SNAPSHOT and when not to use it.

https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it

data-exporter/src/main/java/org/reactome/release/dataexport/EuropePMC.java
jweiser: No, we still have this full mission statement but we decided to go with a shorter version for use in Europe PMC.

data-exporter/src/test/java/org/reactome/release/dataexport/EuropePMCTest.java
jweiser: Yeah, I think it can.  I guess I'm just used to literals in tests!

PR #56, opened 2019-05-14
Fix to prevent modified attribute being overwritten in UpdateStableIds

cookersjs:
The 'modified' attribute of StableIdentifier instances was getting overwritten since the collection was not loaded. This just calls `getAttributeValuesList("modified")` to load the attributes.


cookersjs:
> Looks good! Try to keep the pull requests smaller if possible :p.

Actually I think I could have not declared a variable so this may actually be an inefficient PR

PR #57, opened 2019-05-21
Feature/orthoinference stable id fix

cookersjs:
UPDATE: I changed the base branch of this to merge into OrthoDiagrams. OrthoDiagrams is pending due to a need for proper unit testing. I'd like to merge this branch into it before I merge the orthodiagrams one into develop. This PR should be much, much smaller now

This fixes the fact that stableIdentifier versions were not being incremented properly. This is because stable identifiers were being stored with the newly incremented value as the key, instead of using the original stable identifier. 

Example: R-MMU-12345 is used 4 times in 4 different (but likely related) PhysicalEntitys. The old (incorrect) way would produce [R-MMU-12345, R-12345-2, R-MMU-12345-2, R-MMU-12345-2]. The correct way would be for it to be 12345, 12345-2, 12345-3, 12345-4. This PR fixes that

### DIFF
--- a/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
@@ -40,14 +40,14 @@ public class StableIdentifierGenerator {
         // For now, Human is hard-coded as the source species, so we replace the stableIdentifier source species based on that assumption
         String sourceIdentifier = (String) stableIdentifierInst.getAttributeValue(identifier);
         String targetIdentifier = sourceIdentifier.replace("HSA", speciesAbbreviation);
-
         // Paralogs will have the same base stable identifier, but we want to denote when that happens.
         // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
         int paralogCount = Optional.ofNullable(seenOrthoIds.get(targetIdentifier)).orElse(0) + 1;
+        String targetIdentifierUnmodified = targetIdentifier;
         if (paralogCount > 1) {
             targetIdentifier += "-" + paralogCount;
         }
-        seenOrthoIds.put(targetIdentifier, paralogCount);
+        seenOrthoIds.put(targetIdentifierUnmodified, paralogCount);
 
         // Check that the stable identifier instance does not already exist in DB
         // TODO: Performance check
--- a/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
@@ -43,11 +43,11 @@ public class StableIdentifierGenerator {
         // Paralogs will have the same base stable identifier, but we want to denote when that happens.
         // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
         int paralogCount = Optional.ofNullable(seenOrthoIds.get(targetIdentifier)).orElse(0) + 1;
-        String targetIdentifierUnmodified = targetIdentifier;
+        seenOrthoIds.put(targetIdentifier, paralogCount);
         if (paralogCount > 1) {
             targetIdentifier += "-" + paralogCount;
         }
-        seenOrthoIds.put(targetIdentifierUnmodified, paralogCount);
+
 
         // Check that the stable identifier instance does not already exist in DB
         // TODO: Performance check


orthoinference/README.md
```diff
@@ -2,29 +2,33 @@
...
+During inference, the attribute (input/output/catalyst/regulation) is broken down into its individual <b>PhysicalEntity</b> (PE) instances. These PE instances are each run through the <a href="https://github.com/reactome/data-release-pipeline/blob/develop/orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java">createOrthoEntity</a> method. This method infers according to PE type: <i>GenomeEncodedentity/EntityWithAccessionedSequence, Complex/Polymer, EntitySet or SimpleEntity</i>. In cases where the PE itself contains multiple <i>GKInstance</i> attributes (eg: Complexes <i>hasComponent</i> attribute, EntitySets <i>hasMember</i> attribute), these too are run through the createOrthoEntity method and inferred. Through this system, PE instances will be recursively broken down until they reach the <b>EntityWithAccessionedSequence</b> (EWAS) level and are inferred in the <a href="https://github.com/reactome/data-release-pipeline/blob/develop/orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java">inferEWAS</a> method.
```
jweiser: For completeness, do you think Polymers should be mentioned for the PE types that contain multiple GKInstances?

orthoinference/README.md
```diff
@@ -2,29 +2,33 @@
...
+- <a href="https://github.com/reactome/data-release-pipeline/tree/develop/orthopairs">Orthopair</a> file generation must have been completed.
+- The `slice_current` database will need to be dumped and restored to a new `release_current` database.
 - Set the `config.properties` file
-- Locally install a build of <a href="https://github.com/reactome/data-release-pipeline/tree/develop/release-common-lib">release-common-lib</a>, following the instructions at the link
-- `normal_event_skip_list.txt` needs to be placed in `src/main/resources/` folder
+- Locally install a build of <a href="https://github.com/reactome/data-release-pipeline/tree/develop/release-common-lib">release-common-lib</a>, following the instructions at the link.
+- `normal_event_skip_list.txt` needs to be placed in `src/main/resources/` folder.
```
jweiser: These two would be good to add to the release notes for things we want to automate.  I think the `setup_orthoinference.pl` script does them, but it is implemented in Perl and should probably be ported.

orthoinference/README.md
```diff
@@ -62,8 +66,9 @@ Typically there are a number of instances that inference will be skipped for. Th
   Once all <a href="https://github.com/reactome/data-release-pipeline/tree/develop/orthoinference#-preparing-orthoinference-">prerequisites</a> have been completed, running the <a href="https://github.com/reactome/data-release-pipeline/blob/develop/orthoinference/runOrthoinference.sh">runOrthoinference.sh</a> script will begin the process. This bash script performs a git pull to update the repo with any changes that may have happened between releases. It then builds the orthoinference jar file with all dependencies and then executes it for each species that will be projected to.
   
   <b>Note</b>: To run orthoinference on particular species, modify the 'allSpecies' array in <a href="https://github.com/reactome/data-release-pipeline/blob/develop/orthoinference/runOrthoinference.sh">runOrthoinference.sh</a> so that it only contains the species you wish to project too. Alternatively, if the jar file has been built and only one species needs to be inferred, run the following command:<br> 
-`java -Xmx4096m -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar speciesCode`
+`java -jar target/orthoinference-0.0.1-SNAPSHOT-jar-with-dependencies.jar speciesCode`
```
jweiser: Just curious, was the max memory flag causing an issue?

orthoinference/README.md
```diff
@@ -100,28 +104,22 @@ Next, we want to make sure that the new `test_reactome database` can be imported
...
+-  The <b>WebELVTool</b> found in the <a href="https://github.com/reactome/CuratorTool">CuratorTool</a> repo can be used to check the validity of the `test_reactome` database. The WebELVTool jar is built by running an <b>ant</b> build on the <a href="https://github.com/reactome/CuratorTool/blob/master/ant/WebELVDiagram.xml">WebELVDiagram.xml</a> file. The jar file will appear in the `WebELVTool/` folder above the `CuratorTool/` folder. To run the program, navigate to the directory with the jar file and run:<br> `java -jar webELVTool.jar (host) (release_current) (mysql_username) (mysql_password) (mysql_port) (reactome_author_ID)`<br>The output should be many lines of 'Predicting pathway' or 'Working on pathway'. If the script runs successfully, then it is implied `test_reactome` is appropriately populated.
```
jweiser: Should the mentions of `test_reactome` be swapped for `release_current`?

orthoinference/src/main/java/org/reactome/orthoinference/StableIdentifierGenerator.java
```diff
@@ -40,14 +40,14 @@ public GKInstance generateOrthologousStableId(GKInstance inferredInst, GKInstanc
...
-
         // Paralogs will have the same base stable identifier, but we want to denote when that happens.
         // We pull the value from `seenOrthoIds`, increment it and then add it to the stable identifier name (eg: R-MMU-123456-2)
         int paralogCount = Optional.ofNullable(seenOrthoIds.get(targetIdentifier)).orElse(0) + 1;
+        String targetIdentifierUnmodified = targetIdentifier;
         if (paralogCount > 1) {
             targetIdentifier += "-" + paralogCount;
         }
-        seenOrthoIds.put(targetIdentifier, paralogCount);
+        seenOrthoIds.put(targetIdentifierUnmodified, paralogCount);
```
jweiser: Maybe this line can be moved to just after the initialization of the `targetIdentifier` on line 42.  That way, it will get stored before it is possibly modified by the paralogCount logic and you won't need the `targetIdentifierUnmodified` temp variable.  One less thing to understand and keep track of.

cookersjs: Good point, I'll make that change now

PR #58, opened 2019-05-22
Swapped pins and pprj file-generating logic

cookersjs:
This will merge into the pending goa-prepare branch, which is waiting on completed unit tests. I'd like to merge this first before merging the goa-prepare branch into develop.

This is a logic swap. This class normally creates 3 files: pins, pprj, and pont. The pins and pprj files were being populated with the wrong file type, so the code for each was swapped. It has been tested and confirmed to work.

### DIFF
--- a/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/FetchTestReactomeOntologyFiles.java
+++ b/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/FetchTestReactomeOntologyFiles.java
@@ -87,9 +87,6 @@ public class FetchTestReactomeOntologyFiles {
 					if (line.contains(".pont")) {
 						line = line.replaceAll("[a-zA-Z0-9]+.pont", pontFilename);
 					}
-					if (line.contains(".pins")) {
-						line = line.replaceAll("[a-zA-Z0-9]+.pins", pprjFilename);
-					}
 					Files.write(Paths.get(pinsFilename), line.getBytes(), StandardOpenOption.APPEND);
 				}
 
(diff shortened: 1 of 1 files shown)


downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/FetchTestReactomeOntologyFiles.java
```diff
@@ -65,32 +65,32 @@ public static void execute(MySQLAdaptor dba, String releaseNumber) throws SQLExc
...
 						line = "\n";
-						pprjSwitch = false;
+						pinsSwitch = false;
 					}
 					if (line.contains(".pont")) {
 						line = line.replaceAll("[a-zA-Z0-9]+.pont", pontFilename);
 					}
 					if (line.contains(".pins")) {
-						line = line.replaceAll("[a-zA-Z0-9]+.pins", pinsFilename);
+						line = line.replaceAll("[a-zA-Z0-9]+.pins", pprjFilename);
```
jweiser: Why is *.pins replaced with the `pprjFilename` value (instead of the `pinsFilesname` value) but *.pont is replaced with the `pontFilename` value?

SolomonShorser-OICR: I thought the pont file was created OK, just the pins and pprj had each others' names.

jweiser: Yeah, so I'm just confused why the pont file line is replaced with the pont value but the pins file line is replaced by the pprj value?  If they are getting inverted, maybe changing which variable gets assigned to what value is better instead of changing the value in the replacement?  It looks like it is mis-matched at the moment in this line, but I could be mis-understanding.

cookersjs: You are right that this is confusing -- I think this is related to why this happened in the first place. On the OICR server, if you look at one of the perl-generated pins/pprj (release 66 or so) files, it showed the pprj and pins file contents opposite to how it was being produced by the java code. I think I originally designated the pins/pprj files based on parsing the object (so I assumed when the line contains pins that it was for the pins file). This is making me wonder if the perl way of doing it was actually populating the wrong files? I'll look into it

cookersjs: OK so it turns out that that condition is never hit, so I have deleted it

cookersjs: For further reading on the pprj, pins and pont files: https://protegewiki.stanford.edu/wiki/PrF_UG_files_protege_files

PR #74, opened 2019-08-01
Feature/protege export batch

SolomonShorser-OICR:
New code to run the Perl protege exporter on top-level pathways, and then collect the results in a single tar file.

### DIFF
--- a/downloadDirectory/README.md
+++ b/downloadDirectory/README.md
@@ -186,6 +186,16 @@ This step copies the `gene_association.reactome` file that is produced during th
 
 This step copies the `models2pathways.tsv` file that is produced during the <a href="https://github.com/reactome/Release/tree/master/scripts/release/biomodels">Biomodels</a> step. Comparing the file to previous releases is sufficient for this step.
 
+<h4>Protege exporter</h4>
+This step will run the protege export code in `GKB::WebUtils`. The Perl code will create an archive file containing a pins, pont, and pprj file, for a given pathway. This code is designed to only export TOP-LEVEL pathways, as defined as pathways associated with the FrontPageItem in the database.
+
+This step takes the following configuration options:
+ - protegeexporter.pathToWrapperScript - This is the _absolute_ path to the directory containing the Perl wrapper script `run_protege_exporter.pl`. The script should be located in `src/main/resources` for _this_ project.
+ - protegeexporter.parallelism - The number of concurrent protege export jobs to run. Try to keep this smaller than the number of available cores (MySQL and your operating system should get 1 core each, at least).
+ - protegeexporter.extraIncludes - If you need to specify additional include paths for Perl, use this option. this should be a comma-separate string formatted as: `-I/alt/path/to/perl/libs,-I/other/alt/path/to/libs`.
+ - protegeexporter.filterIds - If you want to filter to only export specific pathways, specify them here as a comma-separated list of DB_IDs.
+ - protegeexporter.filterSpecies - If you want to filter to only export pathways of a specific species, you can specify a comma-separated list here. Normally, you would just set this to `Homo sapiens`.
+
 <h4>CreateReactome2BioSystems</h4>
 
 This step creates a zip file containing an <i>NCBI BioSystems</i>-formatted `xml` file for each of Reactome's <b>primary model organisms</b>. More information on NCBI BioSystems is found at the <a href="https://www.ncbi.nlm.nih.gov/biosystems/">website</a>.
--- a/downloadDirectory/README.md
+++ b/downloadDirectory/README.md
@@ -189,9 +189,9 @@ This step copies the `models2pathways.tsv` file that is produced during the <a h
 <h4>Protege exporter</h4>
 This step will run the protege export code in `GKB::WebUtils`. The Perl code will create an archive file containing a pins, pont, and pprj file, for a given pathway. This code is designed to only export TOP-LEVEL pathways, as defined as pathways associated with the FrontPageItem in the database.
 
-This step takes the following configuration options:
+This step takes the following configuration options, specified in `config.properties`:
  - protegeexporter.pathToWrapperScript - This is the _absolute_ path to the directory containing the Perl wrapper script `run_protege_exporter.pl`. The script should be located in `src/main/resources` for _this_ project.
- - protegeexporter.parallelism - The number of concurrent protege export jobs to run. Try to keep this smaller than the number of available cores (MySQL and your operating system should get 1 core each, at least).
+ - protegeexporter.parallelism - The number of concurrent protege export jobs to run. Try to keep this smaller than the number of available cores (MySQL and your operating system should get 1 core each, at least). If you do not specify anything for this value, then parallelism will be the default value used by the `ForkJoinPool` class, which is usually the number of cores minus 1.
  - protegeexporter.extraIncludes - If you need to specify additional include paths for Perl, use this option. this should be a comma-separate string formatted as: `-I/alt/path/to/perl/libs,-I/other/alt/path/to/libs`.
  - protegeexporter.filterIds - If you want to filter to only export specific pathways, specify them here as a comma-separated list of DB_IDs.
  - protegeexporter.filterSpecies - If you want to filter to only export pathways of a specific species, you can specify a comma-separated list here. Normally, you would just set this to `Homo sapiens`.
--- a/downloadDirectory/pom.xml
+++ b/downloadDirectory/pom.xml
@@ -11,6 +11,8 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<spring.version>4.2.4.RELEASE</spring.version>
+		<biopax.version>5.1.0</biopax.version>
 	</properties>
 
 	<dependencies>
@@ -38,23 +40,23 @@
 		<dependency>
 			<groupId>org.biopax.paxtools</groupId>
 			<artifactId>paxtools-core</artifactId>
-			<version>5.1.0</version>
+			<version>${biopax.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.biopax.paxtools</groupId>
 			<artifactId>paxtools-trove</artifactId>
-			<version>5.1.0</version>
+			<version>${biopax.version}</version>
 			<scope>runtime</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-context</artifactId>
-			<version>4.2.4.RELEASE</version>
+			<version>${spring.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-instrument</artifactId>
-			<version>4.2.4.RELEASE</version>
+			<version>${spring.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
@@ -103,6 +105,11 @@
 			<groupId>com.googlecode.json-simple</groupId>
 			<artifactId>json-simple</artifactId>
 			<version>1.1.1</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-compress</artifactId>
+			<version>1.18</version>
 		</dependency>
 	</dependencies>
 
--- a/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
+++ b/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
@@ -333,14 +333,12 @@ public class ProtegeExporter
 			tarOutStream.closeArchiveEntry();
 			tarOutStream.flush();
 		}
-		catch (FileNotFoundException e)
-		{
-			e.printStackTrace();
-		}
 		catch (IOException e)
 		{
+			logger.error("Error while adding to tar: {}", e.getMessage());
 			e.printStackTrace();
 		}
+		
 	}
 	
 	/**
--- a/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
+++ b/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
@@ -1,15 +1,25 @@
 package org.reactome.release.downloadDirectory;
 
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.IOException;
-import java.lang.ProcessBuilder.Redirect;
+import java.io.InputStream;
+import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.ForkJoinPool;
+import java.util.stream.Collectors;
 
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.gk.model.GKInstance;
@@ -24,21 +34,57 @@ import org.gk.persistence.MySQLAdaptor;
  */
 public class ProtegeExporter
 {
+	private static final String PROTEGE_ARCHIVE_PATH = "/tmp/protege_files.tar";
+	private static final String PROTEGE_FILES_DIR = PROTEGE_ARCHIVE_PATH.replace(".tar", "/");
 	private static final Logger logger = LogManager.getLogger();
-	
 	private String releaseDirectory;
-	
 	private int parallelism = ForkJoinPool.getCommonPoolParallelism();
-	
 	private List<String> extraIncludes = new ArrayList<>();
-	
 	private String pathToWrapperScript = "";
+	private String downloadDirectory;
+	private Set<Long> pathwayIdsToProcess = new HashSet<>();
+	
+	public ProtegeExporter()
+	{
+		// Default no-arg consturctor.
+	}
+	
+	public ProtegeExporter(Properties props, String releaseDir, String downloadDir)
+	{
+		this.setReleaseDirectory(releaseDir);
+		String propsPrefix = "protegeexporter";
+		int degP = Integer.parseInt(props.getProperty(propsPrefix + ".parallelism"));
+		this.setParallelism(degP);
+		String extraIncludeStr = props.getProperty(propsPrefix+".extraIncludes");
+		if (extraIncludeStr != null && !extraIncludeStr.trim().isEmpty())
+		{
+			List<String> extraIncs = Arrays.asList(extraIncludeStr.split(","));
+			this.setExtraIncludes(extraIncs);
+		}
+		String pathToWrapper = props.getProperty(propsPrefix+".pathToWrapperScript");
+		this.setPathToWrapperScript(pathToWrapper);
+		this.setDownloadDirectory(downloadDir);
+		
+		String filterIds = props.getProperty(propsPrefix+".filterIds");
+		if (filterIds != null && !filterIds.trim().isEmpty())
+		{
+			List<String> ids = Arrays.asList(filterIds.split(","));
+			this.setPathwayIdsToProcess( ids.stream().map(Long::valueOf).collect(Collectors.toSet()) );
+		}
+	}
 	
 	public void execute(MySQLAdaptor dba)
 	{
+		// Create the protege files.
+		createProtegeFiles(dba);
+		// Now put all of the files into an archive.
+		tarProtegeFiles();
+	}
+
+	private void createProtegeFiles(MySQLAdaptor dba)
+	{
 		try
 		{
-			 
 			@SuppressWarnings("unchecked")
 			Collection<GKInstance> frontPages =  dba.fetchInstancesByClass(ReactomeJavaConstants.FrontPage);
 			// There should only ever be one FrontPage object, but the database returns a list, so let's just go with that...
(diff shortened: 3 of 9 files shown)


downloadDirectory/README.md
```diff
@@ -186,6 +186,16 @@ This step copies the `gene_association.reactome` file that is produced during th
 
 This step copies the `models2pathways.tsv` file that is produced during the <a href="https://github.com/reactome/Release/tree/master/scripts/release/biomodels">Biomodels</a> step. Comparing the file to previous releases is sufficient for this step.
 
+<h4>Protege exporter</h4>
+This step will run the protege export code in `GKB::WebUtils`. The Perl code will create an archive file containing a pins, pont, and pprj file, for a given pathway. This code is designed to only export TOP-LEVEL pathways, as defined as pathways associated with the FrontPageItem in the database.
+
+This step takes the following configuration options:
+ - protegeexporter.pathToWrapperScript - This is the _absolute_ path to the directory containing the Perl wrapper script `run_protege_exporter.pl`. The script should be located in `src/main/resources` for _this_ project.
+ - protegeexporter.parallelism - The number of concurrent protege export jobs to run. Try to keep this smaller than the number of available cores (MySQL and your operating system should get 1 core each, at least).
```
jweiser: Just to clarify, does this mean the number of cores should always be set to at least 2?

downloadDirectory/pom.xml
```diff
@@ -104,6 +106,11 @@
 			<artifactId>json-simple</artifactId>
 			<version>1.1.1</version>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-compress</artifactId>
+			<version>1.18</version>
```
jweiser: I like the extraction of the versions to properties!  I know it only appears once, but could this one be put in that section too so all dependency version are together?

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
```diff
@@ -0,0 +1,67 @@
...
+
+@SuppressWarnings("static-method")
+public class TestProtegeExporter
+{
+	private static final String RELEASE_NUM = "70";
+	private static String host;
+	private static String name;
+	private static String user;
+	private static String password;
+	static final String PATH_TO_JAVA_ROOT = "/home/sshorser/workspaces/reactome/ReleaseNG/data-release-pipeline/downloadDirectory/";
```
jweiser: Can the `PATH_TO_JAVA_ROOT` be relative so the test can run when the repository is cloned on other machines?

downloadDirectory/src/main/resources/run_protege_exporter.pl
```diff
@@ -0,0 +1,9 @@
+#! /usr/bin/perl
```
jweiser: Best not to have a space between the `#!` and the path to the perl binary.

jweiser: `use strict;` and `use warnings;` should also be added

downloadDirectory/src/main/resources/run_protege_exporter.pl
```diff
@@ -0,0 +1,9 @@
...
+
+use GKB::WebUtils;
+
+my $wu = GKB::WebUtils->new_from_cgi();
+my $id = @ARGV[0];
+my $file_name = @ARGV[1];
```
jweiser: you can add `|| die "no id/file_name specified"` to lines 6 and 7 to test the arguments are given

jweiser: To access individual element values from the @ARGV array, it needs to be `$ARGV[0]` and `$ARGV[1]`

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
```diff
@@ -0,0 +1,67 @@
...
+		Properties props = new Properties();
+		try(FileInputStream fis = new FileInputStream("src/test/resources/db.properties"))
+		{
+			props.load(fis);
+			TestProtegeExporter.host = props.getProperty("db.host");
+			TestProtegeExporter.name = props.getProperty("db.name");
+			TestProtegeExporter.user = props.getProperty("db.user");
+			TestProtegeExporter.password = props.getProperty("db.password");
+			String release = RELEASE_NUM;
+			Files.createDirectories(Paths.get(release+"/"));
```
jweiser: Can there be a corresponding "After" to clean up this created directory from a run of the tests?

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
```diff
@@ -0,0 +1,67 @@
...
+		testExporter.setPathToWrapperScript(PATH_TO_JAVA_ROOT + "src/main/resources/");
+		testExporter.setExtraIncludes(Arrays.asList("-I/home/ubuntu/perl5/lib/perl5/","-I/home/sshorser/perl5/lib/perl5/"));
+		testExporter.setParallelism(4);
+		testExporter.setPathwayIdsToProcess(new HashSet<>(Arrays.asList(1670466L,8963743L,870392L,1500931L,5205647L)));
+		testExporter.setDownloadDirectory(RELEASE_NUM);
+		testExporter.setSpeciesToProcess(new HashSet<>(Arrays.asList("Mycobacterium tuberculosis")));
+		testExporter.execute(adaptor);
+		final String pathToFinalTar = PATH_TO_JAVA_ROOT + RELEASE_NUM + "/protege_files.tar";
+		assertTrue(Files.exists(Paths.get(pathToFinalTar)));
+		assertTrue(Files.size(Paths.get(pathToFinalTar)) > 0);
```
jweiser: Can any tests or asserts be done on the name or contents of the files produced?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+			tarOutStream.closeArchiveEntry();
+			tarOutStream.flush();
+		}
+		catch (FileNotFoundException e)
+		{
+			e.printStackTrace();
+		}
+		catch (IOException e)
+		{
+			e.printStackTrace();
```
jweiser: Since both are printing a stack trace, could the exceptions be combined into one catch block?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+			tarOutStream.putArchiveEntry(entry);
+			
+			final int len = 1024;
+			byte[] buff = new byte[len];
+			long bytesRead = is.read(buff, 0, len);
+			while (bytesRead > 0)
+			{
+				// If bytesRead + len > entrySize then an exception will be thrown, so always take min of len,bytesRead. 
+				tarOutStream.write(buff, 0, (int) Math.min(len, bytesRead));
+				bytesRead = is.read(buff, 0, len);
```
jweiser: Could this be done in the while condition itself so it is not repeated from line 313:
```
long bytesRead;
while ((bytesRead = is.read(buff, 0, len)) > 0) {
    tarOutStream.write(buff, 0, (int) Math.min(len, bytesRead));
}

File: downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java [hunk 10]
```diff
@@ -0,0 +1,404 @@
...
+		}
+		
+		// Now that we're finished creating the tar file, move it to the download directory. Overwrite existing.
+		try
+		{
+			Files.move(Paths.get(PROTEGE_ARCHIVE_PATH), Paths.get(this.downloadDirectory + "/protege_files.tar"), StandardCopyOption.REPLACE_EXISTING);
+		}
+		catch (IOException e)
+		{
+			logger.error("An error occurred while trying to move {} to the download directory ({}). You may need to move it manually.", PROTEGE_ARCHIVE_PATH, this.downloadDirectory);
```
jweiser: Could the exception object be passed to the logger.error statement?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+				if (protegeFile.toFile().isFile() && protegeFile.getFileName().toString().matches("Reactome_pathway_.*\\.tar\\.gz"))
+				{
+					logger.info("Adding {} to protege_files.tar", protegeFile.toString());
+					addFileToTar(tarOutStream, protegeFile);
+				}
+			});
+			tarOutStream.finish();
+		}
+		catch (IOException e)
+		{
```
jweiser: Should there be a logger.error statement here like there is below?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
jweiser: Can this be logged, also?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+			{
+				String pathwaySpecies = ((GKInstance) pathway.getAttributeValue("species")).getDisplayName();
+				if (!this.speciesToProcess.contains(pathwaySpecies))
+				{
+					processPathway = false;
+				}
+			}
+			catch (Exception e)
+			{
+				e.printStackTrace();
```
jweiser: If trying to obtain the pathwaySpecies throws an exception, should `processPathway` be set to false in the catch block?  I would guess that if the pathwaySpecies couldn't be obtained, the pathway shouldn't be processed with the `speciesFilterInEffect` being true?

jweiser: Could the exception condition be logged, also?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+		
+		String filterIds = props.getProperty(propsPrefix+".filterIds");
+		if (filterIds != null && !filterIds.trim().isEmpty())
+		{
+			List<String> ids = Arrays.asList(filterIds.split(","));
+			this.setPathwayIdsToProcess( ids.stream().map(Long::valueOf).collect(Collectors.toSet()) );
+		}
+		
+		String filterSpecies = props.getProperty(propsPrefix+".filterSpecies");
+		if (filterSpecies != null && !filterSpecies.trim().isEmpty())
```
jweiser: Could this be a helper function since it is used three times?  Something like:
```
private boolean hasNonWhiteSpace(String str) {
    return str != null && !str.trim().isEmpty();
}
```
There's probably a better method name :)

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+				{
+					// parallelStream should use the degree of parallelism set on the "pool" object.
+					pathways.parallelStream().forEach(pathway -> 
+					{
+						boolean processPathway = shouldPathwayBeProcessed(idFilterInEffect, speciesFilterInEffect, pathway);
+						
+						if (processPathway)
+						{
+							logger.info("Running protegeexport script for Pathway: {}", pathway.toString());
+							// include a normalized display_name in the output to make it easiser to identify the contents of an archive.
```
jweiser: Typo: easier

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,404 @@
...
+								else
+								{
+									// Should a non-zero exit throw an exception and interrupt the whole process? I am leaning to "no" - if a *single* pathway fails to export,
+									// I think I'd want the rest of them to continue, and then we can re-run for just the failed pathway.
+									logger.error("Non-zero exit code from protegeexporter script: {}. Check logs for more details.", exitCode);
+								}
+							}
+							catch (IOException e)
+							{
+								e.printStackTrace();
```
jweiser: Could any exceptions triggered in the code base be logged as well?

downloadDirectory/README.md
SolomonShorser-OICR: It's probably a good idea to do that, if you can. If you don't set anything, then it defaults to whatever ForkJoinPool defaults to on that system (usually: number_of_cores - 1)

downloadDirectory/pom.xml
SolomonShorser-OICR: That would suggest ALL version numbers should be replaced with properties, which I think would create needless bloat in the pom file, because there would be quite a few property values for dependency versions that are only used once within the file.

downloadDirectory/src/main/resources/run_protege_exporter.pl
SolomonShorser-OICR: I dunno... `@ARGV[0]` seems to work too. What's the difference, why should I change it?

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
SolomonShorser-OICR: I was going to, but I didn't because I also wanted to be able to inspect the file manually after the code finished.

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
SolomonShorser-OICR: In theory... yes.

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
SolomonShorser-OICR: This might be valid Java, but it's not a _normal_ pattern, and I would call it a code smell. I find it weird, unexpected, and confusion.

downloadDirectory/src/main/resources/run_protege_exporter.pl
SolomonShorser-OICR: Weird... it only complained about `@ARGV[0]` _after_ I changed to the line from `my $id = @ARGV[0];` to `my $id = @ARGV[0] || die "No ID specified specified.\n";`

Perl is weird. ;)

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
SolomonShorser-OICR: Hmm I can look into that...

downloadDirectory/README.md
jweiser: Got it!

downloadDirectory/src/main/resources/run_protege_exporter.pl
jweiser: The "$" symbol is used any time you are referring to a single value even it is from an array.  Whereas the "@" symbol is used when you refer to a list of values.  If you were taking an array slice, for example.  You can take a look at the section "Variable Names and Sigils" here if you're interested: http://modernperlbooks.com/books/modern_perl/chapter_03.html

It is allowed to use the "@" symbol, but it isn't good form.  With warnings enabled, you'll get this:
```
Scalar value @ARGV[0] better written as $ARGV[0] at test.pl line 6.
Scalar value @ARGV[1] better written as $ARGV[1] at test.pl line 7.
```

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
jweiser: Could it be made real :)?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
jweiser: Yeah, the while expression is a bit verbose.  If there's a way to consolidate the calls to `is.read(buff,0,len)` so it happens once in the code instead of twice, though, that'd be good.

downloadDirectory/pom.xml
jweiser: Ah okay, nvm.  I didn't see the whole POM file when I made the comment and didn't realize there were as many dependencies as there are.

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
SolomonShorser-OICR: Maybe... It's not trivial.

SolomonShorser-OICR: But not impossible. If it can be made to work (relatively) quickly, I'll try to add it.

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
jweiser: Okay, I see.  Is there any issue if the test is re-run with the directory and files already existing?  Once the code is finished development, though, the test code should probably clean up after itself?  Any checks done manually could be added as tests in their own right.

SolomonShorser-OICR: I might consider building a "cleanup" option that can be triggered by users who want that...

downloadDirectory/src/test/java/downloadDirectory/TestProtegeExporter.java
```diff
@@ -52,16 +64,51 @@ public void testProtegeExporterIT() throws SQLException, IOException
 		ProtegeExporter testExporter = new ProtegeExporter();
 		
 		testExporter.setReleaseDirectory("/home/sshorser/workspaces/reactome/Release");
-		testExporter.setPathToWrapperScript(PATH_TO_JAVA_ROOT + "src/main/resources/");
+		testExporter.setPathToWrapperScript(pathToJavaRoot + "src/main/resources/");
 		testExporter.setExtraIncludes(Arrays.asList("-I/home/ubuntu/perl5/lib/perl5/","-I/home/sshorser/perl5/lib/perl5/"));
 		testExporter.setParallelism(4);
-		testExporter.setPathwayIdsToProcess(new HashSet<>(Arrays.asList(1670466L,8963743L,870392L,1500931L,5205647L)));
+		HashSet<Long> pathwayIds = new HashSet<>(Arrays.asList(1670466L,8963743L,870392L,1500931L,5205647L));
```
jweiser: Could this be declared as Set<Long>?

SolomonShorser-OICR: Yes, but automatic refactoring chose to be specific and declare it a `HashSet`.

jweiser: Okay, I see.  It is better practice to declare the interface and assign the implementation, though.

SolomonShorser-OICR: Yes, I know that, it will be changed.

jweiser: Okay, thanks!

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
SolomonShorser-OICR: It _could_ be but I don't think it's worth it in this case.

jweiser: Okay, just a thought.

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+ * and the system was not too overloaded. Less than that seemed to complete slower.
+ * @author sshorser
+ *
+ */
+public class ProtegeExporter
+{
+	// The code in GKB::WebUtils always writes protege files to /tmp/ and it doesn't look like that is configurable,
+	// so we'll work in /tmp as well.
+	private static final String PROTEGE_ARCHIVE_PATH = "/tmp/protege_files.tar";
+	private static final String PROTEGE_FILES_DIR = PROTEGE_ARCHIVE_PATH.replace(".tar", "/");
```
cookersjs: There is a /tmp/protege_files.tar file and a /tmp/protege_files/ directory? I suppose it is a working directory? I'll probably find out below

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+	private static final Logger logger = LogManager.getLogger();
+	private String releaseDirectory;
+	private int parallelism = ForkJoinPool.getCommonPoolParallelism();
+	private List<String> extraIncludes = new ArrayList<>();
+	private String pathToWrapperScript = "";
+	private String downloadDirectory;
```
cookersjs: These string variables are initialized differently

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+	private Set<Long> pathwayIdsToProcess = new HashSet<>();
+	private Set<String> speciesToProcess = new HashSet<>();
+	
+	public ProtegeExporter()
```
cookersjs: Could this be removed?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+	}
+	
+	public ProtegeExporter(Properties props, String releaseDir, String downloadDir)
+	{
+		this.setReleaseDirectory(releaseDir);
+		String propsPrefix = "protegeexporter";
+		// Use the value of CommonPoolParallelism if nothing is in the properties file.
+		int degP = Integer.parseInt(props.getProperty(propsPrefix + ".parallelism", String.valueOf(ForkJoinPool.getCommonPoolParallelism())));
+		this.setParallelism(degP);
+		String extraIncludeStr = props.getProperty(propsPrefix+".extraIncludes");
```
cookersjs: Could there be a comment briefly explaining this property?

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+		if (extraIncludeStr != null && !extraIncludeStr.trim().isEmpty())
+		{
+			List<String> extraIncs = Arrays.asList(extraIncludeStr.split(","));
+			this.setExtraIncludes(extraIncs);
+		}
+		String pathToWrapper = props.getProperty(propsPrefix+".pathToWrapperScript");
```
cookersjs: Comment for each of the props? I know they are explained in the readme, but I like small explainer comments in the code

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
```diff
@@ -0,0 +1,414 @@
...
+								if (exitCode == 0)
+								{
+									Files.createDirectories(Paths.get("/tmp/protege_files"));
+									// Move the file to the protege_files, overwrite existing.
+									Files.move(Paths.get("/tmp/"+fileName+".tar.gz"), Paths.get(PROTEGE_FILES_DIR+fileName+".tar.gz"), StandardCopyOption.REPLACE_EXISTING);
+								}
+								else
+								{
+									// Should a non-zero exit throw an exception and interrupt the whole process? I am leaning to "no" - if a *single* pathway fails to export,
+									// I think I'd want the rest of them to continue, and then we can re-run for just the failed pathway.
```
cookersjs: I agree, if any fail we can just re-run the script with a filter set to just try and generate the failed ones

downloadDirectory/src/main/resources/config.properties
```diff
@@ -9,3 +9,11 @@ absoluteReleaseDirectoryPath=/usr/local/gkb/scripts/release/
 releaseDownloadDirectoryPath=/usr/local/gkb/scripts/release/download_directory/
```
cookersjs: These two default paths need to be updated -- though they might already be in an unmerged branch

downloadDirectory/src/test/resources/db.properties
```diff
@@ -0,0 +1,5 @@
+db.user=root
+db.password=root
+db.host=localhost
+db.port=3306
+db.name=release_current_before_addlinks_69
```
cookersjs: Should this db.name property have a hard-coded release number?

cookersjs:
My comments were mostly aesthetic, so I am going to approve

downloadDirectory/src/test/resources/db.properties
SolomonShorser-OICR: That's a file used for testing. I would expect that other developers would update as necessary.

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
SolomonShorser-OICR: No, the test class creates an instance of ProtegeExporter without arguments (it passes them in based on what comes from the test properties file and some test values it defines itself).

In general, if NO constructor is defined, then a class will have a default no-arg constructor. But if you define a constructor that takes arguments, then the class will NOT have a no-arg constructor unless you create one.

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
SolomonShorser-OICR: There are comments on all of the setters for the properties, including extraIncludes. It says:
<pre>If, for some reason, there are libraries/modules that are not in Perl's "@INC",
add them here as list, be sure to prefix each with "-I", as <code>Arrays.asList("-I/home/MY_USER/perl5/lib/perl5/","-I/other/path/to_perl_libs/")</code></pre>

downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java
SolomonShorser-OICR: I have written detailed documentation for the setter methods. I don't want to re-document the same thing too many times in multiple places, because if something changes then there are more areas of documentation that need to be kept in sync.

jweiser: You could add a comment "See 'method name'" if it's already documented elsewhere.

PR #75, opened 2019-08-06
Updates release-common-lib pom.xml

jweiser:
release-common-lib dependency on org.reactome.base.reactome-base update to refer to RELEASE version for SSL database connection


cookersjs:
Looks good

PR #76, opened 2019-08-12
Feature/release fixes justin 1

cookersjs:
Fixes related to Release that were assigned to me based on https://docs.google.com/spreadsheets/d/1SXzS4WBnFQHo8vby6kFPk2uqN2oPo7T1RnAdnYBmyRI/edit?usp=sharing

I recommend checking it out by commit, since they are modular fixes being made.


orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -117,7 +124,8 @@
 						GKInstance firstPsiModInst = (GKInstance) modifiedResidueInst.getAttributeValue(psiMod);
 						if (phosFlag && firstPsiModInst.getAttributeValue(name).toString().contains("phospho")) {
 							String phosphoName = "phospho-" + infEWASInst.getAttributeValue(name);
-							List<GKInstance> ewasNames = (ArrayList<GKInstance>) infEWASInst.getAttributeValuesList(name);
+							List<String> ewasNames = (ArrayList<String>) infEWASInst.getAttributeValuesList(name);
+							String originalName = ewasNames.get(0);
```
jweiser: This line and the next could be combined since the remove method will return the element that is being removed:

`String originalName = ewasNames.remove(0)`

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -67,7 +67,7 @@ public static void inferEvents(Properties props, String pathToConfig, String spe
 		int port = Integer.valueOf(props.getProperty("port"));
 		
 		dbAdaptor = new MySQLAdaptor(host, database, username, password, port);
-		setDbAdaptors(dbAdaptor);
+		setDbAdaptors();
```
jweiser: It may be better to keep the setting of the dbAdaptor in the `setDbAdaptors` method (for both this and other classes) by passing in the created adaptor from the line before.  That way everything to do with setting the adaptor is in one place.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -236,21 +235,20 @@ private static void outputReport(String species) throws IOException
...
-		File reportFile = new File(reportFilename);
-		reportFile.createNewFile();
+		createNewFile(reportFilename);
 		String results = "hsap to " + species + ":\t" + inferredCount + " out of " + eligibleCount + " eligible reactions (" + String.format("%.2f", percentInferred) + "%)\n";
 		Files.write(Paths.get(reportFilename), results.getBytes(), StandardOpenOption.APPEND);
 	}
 	
 	// Statically store the adaptor variable in each class
-	private static void setDbAdaptors(MySQLAdaptor dbAdaptor2) 
+	private static void setDbAdaptors()
```
jweiser: Adding a check in this method to ensure the dbAdaptor being used isn't null when the method is called would be great.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -234,6 +252,7 @@ private static boolean checkValidSpeciesProtein(String homologueId) {
 				referenceDNAInstances.add(alternateRefDNAInst);
 			}
 		}
+		logger.info("Total ReferenceDNASequence instance(s) created: " + referenceDNAInstances.size());
```
SolomonShorser-OICR: I like the idea of info messages giving information on what's been created, and how many. :+1:

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
SolomonShorser-OICR: Creating multiple adaptors is useful _if_ you plan to have multiple concurrent connections to the database.

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
```diff
@@ -25,10 +24,15 @@ public static void main(String[] args) throws Exception {
...
+
 		Properties props = new Properties();
 		props.load(new FileInputStream(pathToConfig));
-        EventsInferrer.inferEvents(props, pathToConfig, speciesCode);
+        EventsInferrer.inferEvents(props, speciesCode);
+
+        // Link report file to report_ortho_inference.txt in website_files_update directory
+        String pathtoWebsiteFilesUpdateWithReportFile = props.getProperty("pathToWebsiteFilesUpdateFolder") + "report_ortho_inference.txt";
+        String reportFilename = "report_ortho_inference_test_reactome_" + props.getProperty("releaseNumber") + ".txt";
+        Runtime.getRuntime().exec("ln " + reportFilename + " " + pathtoWebsiteFilesUpdateWithReportFile);
```
SolomonShorser-OICR: It is possible to create file links from within Java: https://docs.oracle.com/javase/tutorial/essential/io/links.html#symLink

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -122,55 +122,66 @@ public static void inferReaction(GKInstance reactionInst) throws InvalidAttribut
...
 	@SuppressWarnings("unchecked")
-	private static boolean inferReactionInputsOrOutputs(GKInstance reactionInst, GKInstance infReactionInst, String attribute) throws InvalidAttributeException, Exception
+	private static boolean inferReactionInputsOrOutputs(GKInstance reactionInst, GKInstance infReactionInst, String attribute) throws Exception
 	{
-		List<GKInstance> infAttributeInstances = new ArrayList<GKInstance>();
-		for (GKInstance attributeInst : (Collection<GKInstance>) reactionInst.getAttributeValuesList(attribute))
+		List<GKInstance> infAttributeInstances = new ArrayList<>();
+		Collection<GKInstance> attributeInstances = (Collection<GKInstance>) reactionInst.getAttributeValuesList(attribute);
+		logger.info("Total " + attribute + " instances: " + attributeInstances.size());
+		logger.info(attribute.substring(0,1).toUpperCase() + attribute.substring(1) + " instances: " + attributeInstances);
```
SolomonShorser-OICR: So, `attribute.substring(0,1).toUpperCase() + attribute.substring(1)` is an abbreviation of attribute made of it's first two letters? Is it possible that this could get confusing if multiple attributes start with the same two letters, such as `goBiologicalProcess` and `goMolecularFunction`? Although I admit I'm not sure _which_ attributes will come through here.

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	// This generates the inferred Pathway of an inferred RlE. It iterates, inferring parent Pathways until reaching the TopLevelPathway.
+	// Inferred Reactions are not added to the Pathway at this step.
+	private static void createInferredPathwayHierarchy(GKInstance sourceEventInst) throws Exception
+	{
+		List<GKInstance> sourcePathwayReferralInstances = (ArrayList<GKInstance>) sourceEventInst.getReferers(hasEvent);
+
+		if (sourcePathwayReferralInstances != null && sourcePathwayReferralInstances.size() > 0)
```
jweiser: Rather than a guard clause for a null or empty list, this kind of approach can help reduce nesting and make the code a bit clearer:

`https://stackoverflow.com/a/2250056/2295778`

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -0,0 +1,263 @@
...
+		{
+			for (GKInstance sourcePathwayReferralInst : sourcePathwayReferralInstances)
+			{
+				logger.info("Generating inferred Pathway: " + sourcePathwayReferralInst);
+				if (inferredEventIdenticals.get(sourcePathwayReferralInst) == null)
```
jweiser: Could this logic be extracted to a method called something like `inferPathway`?  This would make this method easier to read and allow for unit testing of inferring a single pathway.

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -179,9 +190,15 @@ private static boolean inferReactionCatalysts(GKInstance reactionInst, GKInstanc
...
 				
-				List<GKInstance> activeUnits = new ArrayList<GKInstance>();
-				for (GKInstance activeUnitInst : (Collection<GKInstance>) catalystInst.getAttributeValuesList(activeUnit))
+				List<GKInstance> activeUnits = new ArrayList<>();
+				Collection<GKInstance> activeUnitInstances = (Collection<GKInstance>) catalystInst.getAttributeValuesList(activeUnit);
+				logger.info("Total active unit instances: " + activeUnitInstances);
+				if (activeUnitInstances.size() > 0) {
+					logger.info("Active unit instance(s): " + activeUnitInstances);
+				}
+				for (GKInstance activeUnitInst : activeUnitInstances)
```
SolomonShorser-OICR: The if-statement above checks that `activeUnitInstances` has elements, so this loop that depends on `activeUnitInstances`  could be moved to be within that if-statement.

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -192,35 +209,50 @@ private static boolean inferReactionCatalysts(GKInstance reactionInst, GKInstanc
...
 	{
-		List<GKInstance> inferredRegulations = new ArrayList<GKInstance>();
-		for (GKInstance regulatedInst : (Collection<GKInstance>) reactionInst.getAttributeValuesList("regulatedBy"))
+		List<GKInstance> inferredRegulations = new ArrayList<>();
+		Collection<GKInstance> regulationInstances = (Collection<GKInstance>) reactionInst.getAttributeValuesList("regulatedBy");
+		logger.info("Total RegulatedBy instances: " + regulationInstances.size());
+		if (regulationInstances.size() > 0) {
+			logger.info("Regulation instances: " + regulationInstances);
+		}
+		for (GKInstance regulationInst : regulationInstances)
```
SolomonShorser-OICR: Again, there is an if-statement checking that the iterated structure has elements, so the loop on that iterated structure (`regulationInstances` in this case) probably should be moved to be within that if-statement.

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
SolomonShorser-OICR: From the same thread, but further down, a more modern Java 8 approach: https://stackoverflow.com/a/45735160

jweiser: Looks good!  I am still inclined to extract 

`Optional.ofNullable(someList).orElse(Collections.emptyList()))` into a helper function like `safeList` we could re-use.

Having to ensure lists and collections we retrieve aren't null comes up a lot in our code and personally, I find deep nesting with loops and conditionals can be difficult to read and understand.

Adding a helper method to release-common-lib would alleviate that!

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -0,0 +1,263 @@
...
+		{
+			if (humanPathwayInst.getSchemClass().isValidAttribute(hasEvent)) {
+				if (!seenInferredPathway.contains(humanPathwayInst.getDBID())) {
+					// Collect inferred Events associated with source Event
+					List<GKInstance> inferredEventInstances = new ArrayList<>();
+					for (GKInstance eventInst : (Collection<GKInstance>) humanPathwayInst.getAttributeValuesList(hasEvent)) {
+						if (inferredEventIdenticals.get(eventInst) != null) {
+							inferredEventInstances.add(inferredEventIdenticals.get(eventInst));
+						}
+					}
```
jweiser: The for loop above could be its own method:
```
private List<GKInstance> getInferredEventInstances(GKInstance humanPathwayInst) {
    	// Collect inferred Events associated with source Event
	List<GKInstance> inferredEventInstances = new ArrayList<>();
	for (GKInstance eventInst : (Collection<GKInstance>) humanPathwayInst.getAttributeValuesList(hasEvent)) {
		if (inferredEventIdenticals.get(eventInst) != null) {
			inferredEventInstances.add(inferredEventIdenticals.get(eventInst));
		}
	}
}
```

Extracting related portions of a larger method into their own small methods helps to clarify what the function is doing overall by being more declarative than imperative.  It also helps find where the code that needs to be examined is by having a name so that section can be jumped to quickly from the main function rather than having to consider the entire function.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: If the database can't be connected to the MySQLAdaptor class throws a `MySQLNonTransientConnectionException`, so I think adding additional guards would be redundant

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
cookersjs: Kk I've reverted it

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
cookersjs: I just did that uppercase modification since the attributes `input` and `output` are lowercase, but I wanted the logging to look cleaner. They are the only attribute strings that are used here

SolomonShorser-OICR: ok.

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
cookersjs: The only thing about the `Optional` method is that it gets rid of the `else` condition, which for this particular case has a function. Since this is 'PathwaysInferrer', what is happening at this point in the code is that it starts at some non-TopLevelPathway, infers it, and then infers all the way up the Pathway hierarchy until there are no more referrers (_ipso facto_ a TopLevelPathway). By just iterating on the list, the code loses the chance to log that the Pathway Hierarchy has been completely inferred.

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -189,28 +194,31 @@ private static GKInstance createInfComplexPolymer(GKInstance complexInst, boolea
...
 					return nullInst;
 				}
 			}
-			
+			logger.info("Complex protein counts. Total: " + complexTotalProteinCounts + "  Inferrable: " + complexInferrableProteinCounts);
 			GKInstance infComplexInst = InstanceUtilities.createNewInferredGKInstance(complexInst);
 			infComplexInst.addAttributeValue(summation, complexSummationInst);
 			infComplexInst.addAttributeValue(name, complexInst.getAttributeValue(name));
-			ArrayList<GKInstance> infComponentInstances = new ArrayList<GKInstance>();
+			ArrayList<GKInstance> infComponentInstances = new ArrayList<>();
```
jweiser: One wayward ArrayList that can be changed for List :).

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -26,84 +26,86 @@
...
 /** The heart of the OrthoInference process. This function takes PhysicalEntity (PE) instances and will infer those that are EWAS', Complexes/Polymers, or EntitySets.
 	 The function's arguments are an incoming PE instance and an override attribute. Instances that are comprised of PE's will often recursively call this createOrthoEntity function
 	 on constituent PE's with the override attribute set to 'true'. This ensures that these PE's are inferred, despite the fact that they might not pass some filter criteria.
 	 This is often handled using 'mock' instances (i.e. 'ghost instances' from Perl script), which allow a PE to be inferred without having to commit a 'real' instance to the DB.
 */
-	public static GKInstance createOrthoEntity(GKInstance entityInst, boolean override) throws InvalidAttributeException, Exception
+	public static GKInstance createOrthoEntity(GKInstance entityInst, boolean override) throws Exception
 	{
-		logger.info("\tGenerating orthologous instance for " + entityInst);
+		logger.info("Attempting PE inference: " + entityInst);
```
jweiser: This method has pretty deep nesting.  It would be easier to read if some of the outer conditions were made into guard clauses.

```
if (!entityInst.getSchemClass().isValidAttribute(species)) {
	// This used to have a conditional statement based on the returned value of the 'check_intracellular' function.
	// That function doesn't exist anymore (only seemed to apply to the 'mtub' species, which hasn't been inferred for a while).
	// Since the instance is species-agnostic, just returns the original instance.
	logger.info("Could not find valid species attribute, returning original instance: " + entityInst);
	return entityInst;
}

if (orthologousEntityIdenticals.get(entityInst) != null) {
	logger.info("Inferred PE instance already exists");
	return orthologousEntityIdenticals.get(entityInst);
}
```

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -189,28 +194,31 @@ private static GKInstance createInfComplexPolymer(GKInstance complexInst, boolea
...
+			ArrayList<GKInstance> infComponentInstances = new ArrayList<>();
 			// Inference handling is different depending on if it is a Complex or a Polymer. Complexes will infer all 'components' while Polymers will infer all 'repeatedUnits'.
+			// TODO: Log the ratio of inferred complex/polyer from human?
 			if (complexInst.getSchemClass().isa(Complex))
 			{
-				logger.info("\tInferring Complex components...");
-				for (Object componentInst : complexInst.getAttributeValuesList(hasComponent))
+				Collection<GKInstance> componentInstances = complexInst.getAttributeValuesList(hasComponent);
+				logger.info("Complex components: " + componentInstances);
+				for (Object componentInst : componentInstances)
```
jweiser: This declaration can be a GKInstance instead of an Object so there's no need for a cast.

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -189,28 +194,31 @@ private static GKInstance createInfComplexPolymer(GKInstance complexInst, boolea
...
 					infComponentInstances.add(createOrthoEntity((GKInstance) componentInst, true));
 				}
 				infComplexInst.addAttributeValue(hasComponent, infComponentInstances);
 			} else  if (complexInst.getSchemClass().isa(Polymer))
 			{
-				logger.info("\tInferring Polymer repeatedUnits...");
-				for (Object repeatedUnitInst : complexInst.getAttributeValuesList(repeatedUnit))
+				Collection<GKInstance> repeatedUnitInstances = complexInst.getAttributeValuesList(repeatedUnit);
+				logger.info("Polymer repeated units: " + repeatedUnitInstances);
+				for (Object repeatedUnitInst : repeatedUnitInstances)
```
jweiser: Same as above -- Object can be replaced with GKInstance

jweiser: The similarity in branches could mean a common method could be extracted, too.

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -255,22 +265,31 @@ private static GKInstance createInfEntitySet(GKInstance entitySetInst, boolean o
 		if (inferredEntitySetIdenticals.get(entitySetInst) == null)
 		{
 			// Equivalent to infer_members function in infer_events.pl
-			HashSet<String> existingMemberInstances = new HashSet<String>();
-			ArrayList<GKInstance> membersList = new ArrayList<GKInstance>();
-			logger.info("\tInferring EntitySet member instances...");
-			for (GKInstance memberInst : (Collection<GKInstance>) entitySetInst.getAttributeValuesList(hasMember))
+			HashSet<String> existingMemberInstances = new HashSet<>();
+			ArrayList<GKInstance> infMembersList = new ArrayList<>();
```
jweiser: These two declarations could be "Set" and "List", respectively.

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -280,43 +299,46 @@ private static GKInstance createInfEntitySet(GKInstance entitySetInst, boolean o
...
 					{
 						existingCandidateInstances.add(infCandidateInst.getAttributeValue(name).toString());
-						candidatesList.add(infCandidateInst);
+						infCandidatesList.add(infCandidateInst);
 					}
 				}
+				logger.info("Total number of inferred candidates: " + infCandidatesList.size() + "/" + candidateInstances.size());
 				// Handling of CandidateSets
-				if (candidatesList.size() > 0)
+				if (infCandidatesList.size() > 0)
```
jweiser: Could the deep level of nesting here be changed for guard clauses or using else-if branches?

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -0,0 +1,273 @@
...
+		dba.updateInstanceAttribute(sourcePathwayReferralInst, orthologousEvent);
+
+		//TODO: At this point, sourcePathwayReferralInst is always a Pathway. Perhaps move to its own data structure? Holdout from Perl...
+		updatedInferrableHumanEvents.add(sourcePathwayReferralInst);
+	}
+
+	// This populates the hasEvent slot of all inferred Pathways that were just generated with corresponding inferred reactions
+	private static void addInferredEventsToInferredPathways() throws Exception {
+		Set<Long> seenInferredPathway = new HashSet<>();
+		for (GKInstance humanPathwayInst : updatedInferrableHumanEvents)
```
jweiser: Another area of deep nesting here that could be refactored

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
SolomonShorser-OICR: The general term is "programming to an interface (instead of implementation)": https://stackoverflow.com/questions/1992384/program-to-an-interface-what-does-it-mean

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -17,13 +17,12 @@
...
 import static org.gk.model.ReactomeJavaConstants.*;
 import org.gk.persistence.MySQLAdaptor;
-import org.gk.schema.InvalidAttributeException;
 
 public class SkipInstanceChecker {
 	
 	private static final Logger logger = LogManager.getLogger();
 	private static MySQLAdaptor dba;
-	static Set<String> skipList = new HashSet<String>();
+	static Set<String> skipList = new HashSet<>();
```
jweiser: Just curious if this declaration could be made "private" for consistency?

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
cookersjs: Ahh, one squeaked through!

SolomonShorser-OICR: LOL it's really not _that_ big of a deal. ;)

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
jweiser: Sure, that's fair.  It may be a good, though, to put a check in just in case the method is ever called in some way a null does make it in.  A little bit of paranoid redundancy here could be helpful before setting all adaptors and so this code is protected if there are ever changes to MySQLAdaptor class.

cookersjs:
@SolomonShorser-OICR @jweiser I've made changed according to your suggestions. If you could review the last 2 commits and let me know if it is acceptable to merge, that would be great.

orthoinference/src/main/java/org/reactome/orthoinference/OrthologousEntityGenerator.java
```diff
@@ -43,71 +43,70 @@ public static GKInstance createOrthoEntity(GKInstance entityInst, boolean overri
...
-			return entityInst;
+			logger.warn("Unknown PhysicalEntity class: " + entityInst.getClass());
+		}
+		if (override)
+		{
+			return infEntityInst;
 		}
+		orthologousEntityIdenticals.put(entityInst, infEntityInst);
+			logger.info("PE inference completed: " + entityInst);
+			return orthologousEntityIdenticals.get(entityInst);
```
jweiser: Since `infEntityInst` was put into the`orthologousEntityIdenticals` map, it might be better just to return `infEntityInst` for clarity and to avoid a lookup.

PR #82, opened 2019-08-23
Feature/update release common lib pom

jweiser:
All pom files in data-release-pipeline updated to use release-common-lib v1.0.3 which has been updated to include:

- the reactome-base dependency at v1.2.21
- a new CollectionUtils class with methods to generate "safe" sets and lists (i.e. null values will be transformed into empty sets and lists) so null conditional checks aren't necessary before iterating.


SolomonShorser-OICR:
You've updated the version numbers of a number of projects. Do you have plans to create new version tags for the projects?

jweiser:
> You've updated the version numbers of a number of projects. Do you have plans to create new version tags for the projects?

Yes, creating new tags makes sense.  I updated the version numbers for the projects since one of their dependencies is updated, which looks to call for a patch update in semantic versioning when there is no API change according to some quick reading.

Actually, I initially updated release-common-lib's version with a patch (i.e. 1.0.2 to 1.0.3) but I think I will update it to 1.1.0 since I added some new methods which means the API is updated.

SolomonShorser-OICR:
Yes, if you've added new code then updating to 1.1.0 makes sense.

jweiser:
> Yes, if you've added new code then updating to 1.1.0 makes sense.

Okay, done.

SolomonShorser-OICR:
Also: Does using reactome-base 1.2.21 cause issues with any of these projects? What has changed in this version of reactome-base?

jweiser:
> Also: Does using reactome-base 1.2.21 cause issues with any of these projects? What has changed in this version of reactome-base?

I believe the major change is supporting SSL connection for MySQL.  We've tested most of the projects recently because reactome-base v1.2.21 was needed for the project's code to run.

PR #94, opened 2019-10-04
Feature/jenkins

cookersjs:
Note: This merge has a lot of commits due to the fact that Jenkins clones/pulls repositorys for each run, so I was often making small changes to Jenkinsfiles for a commit. Look at the entire file rather than individual commits!

The Jenkins update. Most java-written release steps now have a Jenkinsfile in their directory that dictates how it will be automatically executed by the Jenkins script runner. This update also changes how config variables are named in the scripts, since after this it is possible to use a 'master' config file that contains all configurable variables within it, meaning things like database names needed to be changed from just the generic 'username' to, for example, 'release.database.user'.

This branch also has renamed some of the projects to match the convention we wanted, from camelCase to kebab-case, which will probably have a few conflicts before merge.

Thanks!


Jenkinsfile-release-config
```diff
@@ -0,0 +1,44 @@
...
+				script{
+				       	def userInput = input(
+          				id: 'userInput', message: "Have the release-specific Jenkins environmental variables been updated for this release?",
+          				parameters: [
+                      					[$class: 'TextParameterDefinition', defaultValue: '', description: '', name: 'response']
+                      				])
+					
+					if (!userInput.toLowerCase().startsWith("y")) {
+						error("Please update release-specific environmental variables found at 'Jenkins -> Manage Jenkins -> Configure System -> Global Properties'.")
+					}
```
jweiser: I think Jenkins allows the creation of functions.  Extracting out the logic would be useful here since it is very similar logic to the check above for the config file.

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
```diff
@@ -27,10 +27,17 @@ public static void main(String[] args) throws Exception {
...
 		if (args.length > 0) {
 			pathToConfig = args[0];
+			if (args.length > 1) {
+				pathToStepConfig = args[1];
+			} else {
+				pathToStepConfig = "src/main/resources/stepsToRun.config";
+			}
 		} else {
 			pathToConfig = "src/main/resources/config.properties";
+			pathToStepConfig = "src/main/resources/stepsToRun.config";
```
jweiser: It looks like if there is more than one argument, `pathToStepConfig` will be an empty String.  Is that okay?

orthoinference/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
+import groovy.json.JsonSlurper
+
+pipeline{
+	agent any
+	
+	stages{
+		stage('Check if upstream builds succeeded'){
```
jweiser: Could this read "Check if Orthopairs and UpdateStableIds builds succeeded"

orthoinference/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+			steps{
+				script{
+					def orthopairsStatusUrl = httpRequest authentication: 'jenkinsKey', url: "${env.JENKINS_JOB_URL}/job/${env.RELEASE_NUMBER}/job/Orthopairs/lastBuild/api/json"
+					def orthopairsStatusJson = new JsonSlurper().parseText(orthopairsStatusUrl.getContent())
+					if(orthopairsStatusJson['result'] != "SUCCESS"){
+						error("Most recent Orthopairs build status: " + orthopairsStatusJson['result'])
+					}
+			    	}
```
jweiser: If possible, a utility function for checking a previous build would be great with parameters of what to check and any error message text to append to a common message.

orthoinference/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+					}
+				}	
+		    	}
+	    	}
+		stage('Setup: Backup release_current'){
+			steps{
+				script{
+					dir('orthoinference'){
+						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
+							sh "mysqldump -u$user -p$pass ${env.RELEASE_CURRENT} > ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_orthoinference.dump"
```
jweiser: A similar question about the password on the command-line to the AddLinks Jenkins PR.  Is it okay to do and is it necessary/doable to provide it in a different way?

orthoinference/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+		stage('Main: Infer mmus'){
+			steps {
+				script{
+					dir('orthoinference'){
+						withCredentials([file(credentialsId: 'Config', variable: 'FILE')]){
+							sh "java -Xmx${env.JAVA_MEM_MAX}m -jar target/orthoinference-${env.ORTHOINFERENCE_VERSION}-jar-with-dependencies.jar $FILE mmus"
+						}
+					}
+				}
+			}
```
jweiser: I remember you mentioned that it was necessary to repeat the inference stage for each species, but I was just wondering if using a groovy function is possible in the Jenkins file to consolidate the common parts?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -75,14 +75,15 @@ public static void inferEvents(Properties props, String species) throws Exceptio
 		setDbAdaptors(dbAdaptor);
 
 		releaseVersion = props.getProperty("releaseNumber");
-		String pathToOrthopairs = props.getProperty("pathToOrthopairs");
+		String pathToOrthopairs = props.getProperty("pathToOrthopairs") + releaseVersion + "/";
```
jweiser: For creating a path string, it's usually preferable to use a library to handle the directory separator:

https://stackoverflow.com/questions/16919501/create-a-path-from-string-in-java7

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
```diff
@@ -58,10 +65,10 @@ public static void main(String[] args) throws Exception {
 			releaseDir.mkdir();
 		}
 
-		String pathToSpeciesConfig = props.getProperty("speciesConfigPath");
+		String pathToSpeciesConfig = "src/main/resources/Species.json";
```
jweiser: Using `Paths.get("src/main/resources/Species.json").toString()` might be better just in case so the system takes care of the directory separators for you.

https://stackoverflow.com/questions/16919501/create-a-path-from-string-in-java7

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
jweiser: Similar to the `Paths.get(...).toString()` comment above so the trailing slash is not hard-coded

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
```diff
@@ -30,13 +29,15 @@ public static void main(String[] args) throws Exception {
...
-        Files.createLink(pathtoWebsiteFilesUpdateWithReportFile, reportFilename);
+		EventsInferrer.inferEvents(props, speciesCode);
+
+		// Link report file to report_ortho_inference.txt in website_files_update directory
+		/*
+		Path pathtoWebsiteFilesUpdateWithReportFile = Paths.get(props.getProperty("pathToWebsiteFilesUpdateFolder") + "report_ortho_inference.txt");
+		Path reportFilename = Paths.get("report_ortho_inference_test_reactome_" + props.getProperty("releaseNumber") + ".txt");
+		Files.deleteIfExists(pathtoWebsiteFilesUpdateWithReportFile);
+		Files.createLink(pathtoWebsiteFilesUpdateWithReportFile, reportFilename);
+		*/
```
jweiser: Can this be removed?

orthopairs/Jenkinsfile
```diff
@@ -0,0 +1,41 @@
+import groovy.json.JsonSlurper
+
+pipeline{
+    agent any
+
+	stages{
+		stage('Check upstream builds succeeded'){
```
jweiser: Could this read "Check release config confirmation build"?

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
cookersjs: No, good catch. I'll make the path in the final 'else' the default string.

orthoinference/Jenkinsfile
cookersjs: I had this reply in the addlinks version of this PR:

> I am sure its doable, but this is probably more easily tackled when we actually deploy this on the release server, in case there are any differences in how the Jenkins user is utilized there than on my computer. For now, Jenkins is very good for making sure credentialed values show up as asterisks. Definitely can consider the use of the `my.cnf` file once its in place, though I don't know if having to switch between the local release server and curator server are going to make things difficult, _a la_ the `pass2production.sh` issue we had for release 70

orthoinference/Jenkinsfile
cookersjs: Yea I tried to do this a couple of weeks ago. I need to take another stab at it for sure

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -48,12 +48,18 @@ public static void main( String[] args ) throws IOException, ParseException {
...
+                pathToConfig = args[0];
+                sourceMappingSpecies = args[1];
+            } else {
+                pathToConfig = args[0];
+                sourceMappingSpecies = "hsap";
+            }
         } else {
             pathToConfig = "src/main/resources/config.properties";
             sourceMappingSpecies = "hsap";
         }
```
jweiser: This could be replaced with:
```
String pathToConfig = args.length > 0 ? args[0] : Paths.get("src/main/resources/config.properties").toString()`
String sourceMappingSpecies = args.length > 1 ? args[1] : "hsap";
```

orthopairs/updateOrthopairsConfig.sh
```diff
@@ -0,0 +1,28 @@
+
```
jweiser: Could the following be added?
```
#!/bin/bash
set -e

cd $DIR
```

scripts/confirmJenkinsConfigurations.sh
```diff
@@ -0,0 +1,43 @@
...
+while read line; do
+  if [[ $line =~ releaseNumber= ]] ; then
+    configFileReleaseNumber=${line#*=}
+  fi
+done < $configFile
+
+if [[ $userInputReleaseNumber != $environmentalReleaseNumber ]] ; then #|| [[ $userInputReleaseNumber != $configFileReleaseNumber ]]; then
+  echo "User input release number ($userInputReleaseNumber) does not match Jenkins environmental release number ($environmentalReleaseNumber)."
+  echo "Please update release-specific environmental variables found at 'Jenkins -> Manage Jenkins -> Configure System -> Global Properties'."
+  exit 2
```
jweiser: Just curious: why an exit value of 2?

scripts/confirmJenkinsConfigurations.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo "This script is used during the SetupReleaseConfig step in Jenkins. It confirms correct configurations that will be used by Jenkins during Reactome's data release."
+    echo ""
+    echo "Usage: 'bash confirmJenkinsConfigurations.sh configFilepath jenkinsEnvironmentReleaseNumber jenkinsEnvironmentPreviousReleaseNumber userInputReleaseNumber'"
+    exit 1
+fi
+
+configFile=$1
+environmentalReleaseNumber=$2
+environmentalPrevReleaseNumber=$3
+userInputReleaseNumber=$4
```
jweiser: Was there a reason to use ordered rather than named argument for this script?

scripts/confirmJenkinsConfigurations.sh
```diff
@@ -0,0 +1,43 @@
+set -e
```
jweiser: Could the following be added?
```
#!/bin/bash

cd $DIR
```

scripts/confirmJenkinsConfigurations.sh
```diff
@@ -0,0 +1,43 @@
...
+environmentalPrevReleaseNumber=$3
+userInputReleaseNumber=$4
+
+configFileReleaseNumber=
+
+while read line; do
+  if [[ $line =~ releaseNumber= ]] ; then
+    configFileReleaseNumber=${line#*=}
+  fi
+done < $configFile
```
jweiser: Could you add some comments to detail what the loop is iterating over and what is being done inside to decide what value `configFileReleaseNumber` receives?

update-dois/Jenkinsfile
```diff
@@ -0,0 +1,120 @@
+import groovy.json.JsonSlurper
+
+pipeline {
+	agent any
+
+	stages {
+		stage('Check if upstream builds succeeded'){
```
jweiser: Could this read "Check if UpdateStableIds build succeeded"?

update-dois/Jenkinsfile
```diff
@@ -0,0 +1,120 @@
...
+	    	}
+		stage('Setup: Back up DBs'){
+			steps{
+				script{
+					dir('update-dois'){
+						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
+							sh "mysqldump -u$user -p$pass ${env.RELEASE_CURRENT} > ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_update_dois.dump"
+							sh "gzip -f ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_update_dois.dump"
+							sh "mysqldump -u$user -p$pass -h${env.CURATOR_SERVER} ${env.GK_CENTRAL} > ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_before_update_dois.dump"
+							sh "gzip -f ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_before_update_dois.dump"
```
jweiser: Since it's re-used a few times, maybe `${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_before_update_dois.dump` could be assigned to its own local variable if that's possible.

update-dois/Jenkinsfile
```diff
@@ -0,0 +1,120 @@
...
+		}	
+		stage('Post: Backup DBs'){
+			steps{
+				script{
+					dir('update-dois'){
+						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
+							sh "mysqldump -u$user -p$pass ${env.RELEASE_CURRENT} > ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_after_update_dois.dump"
+							sh "gzip -f ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_after_update_dois.dump"
+							sh "mysqldump -u$user -p$pass -h${env.CURATOR_SERVER} ${env.GK_CENTRAL} > ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_after_update_dois.dump"
+							sh "gzip -f ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_after_update_dois.dump"
```
jweiser: Similar to the above - a local variable for the dump file?

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
cookersjs: Thanks! I'd been having trouble thinking of a succinct way to write these and I was just spacing on the ternary operators

scripts/confirmJenkinsConfigurations.sh
cookersjs: I'm not especially familiar with exit numbers actually, all I needed to happen was for Jenkins to see that it exited due to incorrect behaviour (see: error). Since it was an error of my own description, I was just looking for basic 'exit with errors' behaviour, and this worked

scripts/confirmJenkinsConfigurations.sh
cookersjs: No reason, I can add that in

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -12,55 +13,62 @@
...
+	  
+    boolean testMode = true;
     if (args.length > 0 && !args[0].equals(""))
     {
       pathToConfig = args[0];
-      pathToReport = args[1];
+      if (args.length > 1 && !args[1].equals("")) {
+        pathToReport = args[1];
+        testMode = false;
+      }
```
jweiser: A ternary for pathToConfig used at the time of declaration might be more succinct and easier to follow than a nested if:
```
String pathToConfig = args.length > 0 && !args[0].isEmpty() ? args[0] : Paths.get("src/main/resources/config.properties").toString();

String pathToReport = Paths.get("src/main/resources/UpdateDOIs.report").toString(); 
boolean testMode = true;
if (args.length > 1 && !args[1].isEmpty()) {
    pathToReport = args[1];
    testMode = false;
}
```

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -12,55 +13,62 @@
...
+      releaseNumber = props.getProperty("releaseNumber");
+      //if (props.getProperty("testMode") != null) {
+      //  testMode = Boolean.valueOf(props.getProperty("testMode"));
+      //}
 
       // Set up db connections.
-      dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, port);
-      dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, port);
+      dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, portTR);
+      dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, portGK);
```
jweiser: Since the purpose of most of these variables is to create a MySQLAdaptor on lines 64 and 65, it might be worth creating a private method to do the extraction from the `props` object and pass in the value "release" or "curator" to extract the correct value from the Jenkins prop file for the calls to create a release and curator dba, respectively.

```
# Where databaseType is checked to be either "release" or "curator", respectively
# If you're feeling very ambitious, you could make databaseType into an Enum with values "release" and "curator" instead of a String so only permissible values are able to be passed.
private MySQLAdaptor createDBAdaptor(Properties props, String databaseType)  {

}
```

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -31,10 +34,16 @@ public static void setAdaptors(MySQLAdaptor adaptorTR, MySQLAdaptor adaptorGK) {
...
-	public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pathToReport, boolean testMode) {
+	public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pathToReport, String releaseNumber, boolean testMode) throws IOException {
 
+		String doisListFilename = "doisToBeUpdated-v" + releaseNumber + ".txt";
+		File doisListFile = new File(doisListFilename);
 		if (testMode) {
 			logger.info("Test mode is active. Outputting DOIs that can be updated");
+			if (doisListFile.exists()) {
+				doisListFile.delete();
+			}
```
jweiser: You could replace with the Files.deleteIfExists(...) method: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#deleteIfExists(java.nio.file.Path)

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -31,10 +34,16 @@ public static void setAdaptors(MySQLAdaptor adaptorTR, MySQLAdaptor adaptorGK) {
...
+			doisListFile.createNewFile();
```
jweiser: I think using the `Files` class is the preferred way to handle files for Java 8+.  To create a file, you can use this method `Files.createFile(Path path)`:
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createFile-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-

update-stable-ids/Jenkinsfile
```diff
@@ -0,0 +1,105 @@
+import groovy.json.JsonSlurper
+
+pipeline {
+	agent any
+	
+	stages {
+		stage('Check upstream builds succeeded'){
```
jweiser: Similar to other sections - "Check upstream ConfimReleaseConfig build succeeded"?

update-stable-ids/Jenkinsfile
```diff
@@ -0,0 +1,105 @@
...
+				}
+			}
+		}
+		stage('Setup: Back up gk_central'){
+			steps{
+				script{
+					dir('update-stable-ids'){
+						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
+							sh "mysqldump -u$user -p$pass -h${env.CURATOR_SERVER} ${env.GK_CENTRAL} > ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_before_st_id.dump"
+							sh "gzip -f ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_before_st_id.dump"
```
jweiser: Could a local variable for the dump file be used here so any changes to the name only need to be done in one spot?

update-stable-ids/Jenkinsfile
```diff
@@ -0,0 +1,105 @@
...
+		}
+		stage('Post: Backup DBs'){
+			steps{
+				script{
+					dir('update-stable-ids'){
+						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
+							sh "mysqldump -u$user -p$pass -h${env.CURATOR_SERVER} ${env.GK_CENTRAL} > ${env.GK_CENTRAL}_${env.RELEASE_NUMBER}_after_st_id.dump"
+							sh "gzip -f gk_central_${env.RELEASE_NUMBER}_after_st_id.dump"
+							sh "mysqldump -u$user -p$pass ${env.SLICE_CURRENT} > ${env.SLICE_CURRENT}_${env.RELEASE_NUMBER}_after_st_id.dump"
+							sh "gzip -f ${env.SLICE_CURRENT}_${env.RELEASE_NUMBER}_after_st_id.dump"
```
jweiser: Similar to before - local variable for dump file name?

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
```diff
@@ -27,33 +27,34 @@ public static void main( String[] args ) throws Exception
...
        MySQLAdaptor dbaPrevSlice = new MySQLAdaptor(sliceHost, prevSliceDatabase, sliceUsername, slicePassword, port);
        
-       String gkCentralUsername = props.getProperty("gkCentralUsername");
-       String gkCentralPassword = props.getProperty("gkCentralPassword");
-       String gkCentralHost = props.getProperty("gkCentralHost");
-       String gkCentralDatabase = props.getProperty("gkCentralDatabase");
+       String gkCentralUsername = props.getProperty("curator.database.user");
+       String gkCentralPassword = props.getProperty("curator.database.password");
+       String gkCentralHost = props.getProperty("curator.database.host");
+       String gkCentralDatabase = props.getProperty("gk_central.name");
```
jweiser: For consistency, maybe this could be "curator.database.name"?

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
jweiser: Actually, looking over this again, I think I'm wrong.  Both `pathToConfig` and `pathToStepConfig` should be set.  It might be a good place to use ternaries for the varaibles initialization, though:

```
String pathToConfig = args.length > 0 ? args[0] : Paths.get("src/ /main/resources/config.properties").toString();
String pathToStepConfig = args.length > 1 ? args[1] : Paths.get("src/main/resources/stepsToRun.config").toString();
```

orthoinference/Jenkinsfile
jweiser: Sounds good - we can work on it in a new PR after trying it on a server.

scripts/confirmJenkinsConfigurations.sh
jweiser: I'm not too familiar with exit codes either, but this looks like a good, quick summary:
http://www.tldp.org/LDP/abs/html/exitcodes.html

Jenkinsfile-release-config
```diff
@@ -0,0 +1,44 @@
+pipeline{
+	agent any
+	
+	stages{
+		stage('Confirm configurations'){
+			steps{
+				script{
+          				def userInput = input(
+          				id: 'userInput', message: "Has the Config file credentials been updated for this release?",
+          				parameters: [
```
SolomonShorser-OICR: It looks like this line (and many others) use a mix of tabs and spaces. As this is all new code, could we have the tabs/spaces made consistent within the file? Either all tabs, or all spaces, I'm not _too_ picky. But mixing tabs and spaces can cause weird & annoying formatting issues later.

download-directory/Jenkinsfile
```diff
@@ -0,0 +1,65 @@
...
+				}
+			}
+		}
+        	stage('Setup: Build DownloadDirectory archive'){
+            		steps{
+				script{
+                    			dir('download-directory'){
+						sh "mvn clean package -DskipTests"
+			    			sh "rm -rf downloadDirectory"
+						sh "unzip -o target/downloadDirectory-distr.zip"
```
SolomonShorser-OICR: What is `downloadDirectory-distr.zip`, and why does it get unzipped?

download-directory/Jenkinsfile
```diff
@@ -0,0 +1,65 @@
...
+                			}
+          			}
+            		}
+        	}
+		stage('Main: Run DownloadDirectory'){
+			steps{
+				script{
+					dir('download-directory'){
+						withCredentials([file(credentialsId: 'Config', variable: 'ConfigFile')]){
+							withCredentials([file(credentialsId: 'stepsToRun', variable: 'StepsToRun')]){
```
SolomonShorser-OICR: How does `StepsToRun` get configured?

update-dois/Jenkinsfile
```diff
@@ -0,0 +1,120 @@
...
+							sh "java -jar target/update-dois-${env.UPDATE_DOIS_VERSION}-jar-with-dependencies.jar $ConfigFile"
+						}
+					}
+				}
+			}
+		}
+		stage('Send email of updateable DOIs to curator'){
+			steps{
+				script{
+					emailext (
```
SolomonShorser-OICR: Have you gotten email from Jenkins to work?

download-directory/Jenkinsfile
cookersjs: This is how download-directory gets packaged. Since it depends on a lot of files stored in `src/main/resources`, it basically packages all of them, with the downloadDirectory jar into an archive. This could be one of those things that I've done in an elaborate and unnecessary way, but this is how download directory has always worked.

download-directory/Jenkinsfile
cookersjs: In Jenkins, I have it so that it is stored as a secret file. This way if you still want to run specific steps in download directory, its still possible. All you have to do is upload the file with the ones you want to skip commented out

update-dois/Jenkinsfile
cookersjs: On my workstation, yes, but it involves opening another port. I figure this is something we'll need to discuss moving forward, but as a proof-of-concept, yes it works

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
cookersjs: Yes, I just meant to keep it here as a reminder to implement the symlink code in Jenkins/bash. I'd like to keep it here and then maybe create a small PR in the future for that

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
cookersjs: Yea, to be honest I think updateDOIs is in need of its own refactoring, both to refresh variable naming and just to take a hard look at how its doing what its doing. Since it was so early in my rewrite days, the codes not very efficiently done nor is it apparent to the user why its doing what its doing. When I have time to do that, I'll create a seperate PR for stuff like that.

 That being said, if I can still create that function in this PR

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
cookersjs: That would work for this particular case, but since release has multiple DBs it wouldn't be as clean. I could do something like 'release.current.database.name' if that seems better

download-directory/Jenkinsfile
SolomonShorser-OICR: Right. Because the BioPax step needs to run with that "javaagent" argument, and that requires everything to be bundled up in a zip file.

download-directory/Jenkinsfile
SolomonShorser-OICR: Ok. If there's a README for the Jenkins usage, this should be in it.

update-dois/Jenkinsfile
SolomonShorser-OICR: I think Robin can give you the config details to get this working on the server.

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
jweiser: Sounds good.  Could you create an GitHub issue for that just so it stays on the task list radar?

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
jweiser: Great, thanks!  Could you create a GitHub issue for this separate PR, too?

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
jweiser: For sure, I think it's okay to keep "slice_current.name", "release_current.name", etc. for release databases since there are many of them, but for the curator database it should only be one being used and since it's possible not to be using gk_central, so "curator.database.name" or "curator_database.name" might be better?

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
cookersjs: Done

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
cookersjs: Done

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
cookersjs: Fair point, I can make that change

orthoinference/Jenkinsfile
cookersjs: This has been done now, although in a bit of a janky way

Jenkinsfile-release-config
cookersjs: Yes it allows functions, though looking at it just now I think the amount of boilerplate surrounding this would be adding as much if not more code. If this module expands to add more questions of exactly the same structure I would want to do it, but for simplicity's sake I'd prefer to leave it as is.

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
cookersjs: Thanks, I used these but I didn't use the `Paths.get().toString()`, I just used the string variable

orthoinference/Jenkinsfile
cookersjs: Done

scripts/confirmJenkinsConfigurations.sh
cookersjs: Thanks, I changed all exit codes to 1

scripts/confirmJenkinsConfigurations.sh
cookersjs: Done

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
cookersjs: Thanks, implemented

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
cookersjs: Done

cookersjs:
@SolomonShorser-OICR @jweiser I've made all changes requested, if you could review a final time and let me know if there are anymore required, that would be great!

orthoinference/Jenkinsfile
```diff
@@ -1,214 +1,79 @@
...
+		}
+		// This stage backs up the release current database before it is modified.
 		stage('Setup: Backup release_current'){
 			steps{
 				script{
 					dir('orthoinference'){
 						withCredentials([usernamePassword(credentialsId: 'mySQLUsernamePassword', passwordVariable: 'pass', usernameVariable: 'user')]){
-							sh "mysqldump -u$user -p$pass ${env.RELEASE_CURRENT} > ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_orthoinference.dump"
-							sh "gzip -f ${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_orthoinference.dump"
+							def release_current_before_orthoinference_dump = "${env.RELEASE_CURRENT}_${env.RELEASE_NUMBER}_before_orthoinference.dump"
```
jweiser: Nice :)

orthoinference/Jenkinsfile
```diff
@@ -1,214 +1,79 @@
...
+							sh "mysqldump -u$user -p$pass ${env.RELEASE_CURRENT} > $release_current_before_orthoinference_dump"
+							sh "gzip -f $release_current_before_orthoinference_dump"
 						}
 					}
 				}
 			}
-		}			
+		}
+		// This stage builds the jar file using maven. It also runs the Main orthoinference process as a 'sub-stage'.
+		// This was due to a restriction in iterating over a list of species names. To iterate, you need to first have a 'stage > steps > script hierarchy.
```
jweiser: Being very nitpicky :), is there a missing single closing quote?

orthoinference/Jenkinsfile
```diff
@@ -1,214 +1,79 @@
...
 						sh "mvn clean compile assembly:single"
 					}
 				}
-			}
-		}
-		stage('Main: Infer mmus'){
-			steps {
+				// This script block executes the main orthoinference code one species at a time.
+				// It takes all Human Reaction instances in the database and attempts to project each Reaction to each species by
+				// stripping them down to the reactio's constituent proteins, checks if the protein homolog exists for that species, and infers it in Reactome's data model.
```
jweiser: Typo - "reaction's"?

orthoinference/Jenkinsfile
SolomonShorser-OICR: _SUUUUPER_ nitpicky! ;)

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -17,18 +18,13 @@ public static void main( String[] args ) throws IOException {
 
 	 // Default locations of properties and pre-set report files
 	 // Will override if arguments are provided
-	 // TODO: Lots of room for error here if only 1 argument provided, or if the report file isn't updated
-    String pathToConfig = "src/main/resources/config.properties";
-    String pathToReport = "src/main/resources/UpdateDOIs.report";
-	  
+    String pathToConfig = args.length > 0 && !args[0].isEmpty() ? args[0] : Paths.get("src/main/resources/config.properties").toString();
+
+    String pathToReport = Paths.get("src/main/resources/UpdateDOIs.report").toString();
```
jweiser: Separation of the path into parameters can be done here, too

update-dois/Jenkinsfile
SolomonShorser-OICR: I'm curious: Is there a reason that sending email is its own stage? Logically, it makes sense to me to include it in the 'Main: UpdateDOIs Test Run' stage, but maybe there's something subtle that I'm missing...

update-dois/Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+							sh "gzip -f $release_current_after_update_dois_dump"
+							sh "mysqldump -u$user -p$pass -h${env.CURATOR_SERVER} ${env.GK_CENTRAL} > $central_before_after_dois_dump"
+							sh "gzip -f $central_before_after_dois_dump"
+						}
+					}
+				}
+			}
+		}
+		// This stage archives all logs and database backups produced by UpdateDOIs.
+		stage('Archive logs and backups'){
```
SolomonShorser-OICR: Also: Why are backing up the database and archive the logs and backup files two separate stages? To me it would make sense to group them as two steps of a single "Post-UpdateDOI" stage. Is there an advantage to breaking everything up such that each stage only contains one step?

update-stable-ids/Jenkinsfile
```diff
@@ -0,0 +1,122 @@
...
+							sh "gzip -f $slice_test_snapshot_dump"
+							sh "mysql -u$user -p$pass -e \'drop database if exists ${env.SLICE_CURRENT}; create database ${env.SLICE_CURRENT}\'"
+							sh "zcat  ${env.SLICE_TEST}_${env.RELEASE_NUMBER}_snapshot.dump.gz 2>&1 | mysql -u$user -p$pass ${env.SLICE_CURRENT}"
+						}
+					}
+				}
+			}
+		}
+		// This stage backs up the gk_central database before it is modified.
+		stage('Setup: Back up gk_central'){
```
SolomonShorser-OICR: Is the database still named "gk_central" on the server, or has it be renamed to "central"? I think that keeping the "gk_" here could cause future confusion with other developers, if they don't know that "gk_" is a historic prefix, not used anywhere else anymore.  BUT... if the database is actually still using the name "gk_central" in MySQL, then it's probably OK to leave this as-is.

update-dois/Jenkinsfile
cookersjs: No there isn't a particular advantage, aside from allowing for easier identification of failures (but that is sort of trivial between DB backup and archiving). I was just trying to be explicit for stage naming/theming. I can have a single 'post' step that includes backups and archiving if wanted

update-stable-ids/Jenkinsfile
cookersjs: This is actually gk_central that's found on the curator server. On release we've changed it to be named 'central', but not on curator. I could generalize it to be named something like 'Back up curator database' if we think that would be better for others

SolomonShorser-OICR: No, that's OK. I didn't realize it was still "gk_central" on the Curator server.

update-dois/Jenkinsfile
SolomonShorser-OICR: Being able to easily identify failures is a Good Thing, so in that case, maybe it's better to have them separate.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -255,14 +256,14 @@ private static void setDbAdaptors(MySQLAdaptor dbAdaptor)
...
-		
+
 	}
 
 	// Read the species-specific orthopair 'mapping' file, and create a HashMap with the contents
 	private static Map<String, String[]> readHomologueMappingFile(String toSpecies, String fromSpecies, String pathToOrthopairs) throws IOException
 	{
 		String orthopairsFileName = fromSpecies + "_" + toSpecies + "_mapping.txt";
-		String orthopairsFilePath = pathToOrthopairs + orthopairsFileName;
+		String orthopairsFilePath = pathToOrthopairs + "/" + orthopairsFileName;
```
jweiser: Whenever possible, try to use `Path.get()` to avoid hard-coding the file separator as "/"

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -12,55 +14,57 @@
...
 	 // Will override if arguments are provided
-	 // TODO: Lots of room for error here if only 1 argument provided, or if the report file isn't updated
-    String pathToConfig = "src/main/resources/config.properties";
-    String pathToReport = "src/main/resources/UpdateDOIs.report";
-    if (args.length > 0 && !args[0].equals(""))
-    {
-      pathToConfig = args[0];
+    String pathToConfig = args.length > 0 && !args[0].isEmpty() ? args[0] : Paths.get("src/main/resources/config.properties").toString();
+
+    String pathToReport = Paths.get("src/main/resources/UpdateDOIs.report").toString();
```
jweiser: Be sure to separate each part of the directory path into individual parameters passed to `Paths.get()`:

E.g. `Paths.get("src", "main", "resources", "config.properties")`

That way `Paths.get()` defers to the OS for the path separator and it doesn't need hard-coding.

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -12,55 +14,57 @@
...
-      dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, port);
+      dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, portTR);
+      dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, portGK);
     } catch (Exception e) {
       e.printStackTrace();
     }
       UpdateDOIs.setAdaptors(dbaTestReactome, dbaGkCentral);
       logger.info("Starting UpdateDOIs");
-      UpdateDOIs.findAndUpdateDOIs(authorIdTR, authorIdGK, pathToReport, testMode);
+      UpdateDOIs.findAndUpdateDOIs(authorIdTR, authorIdGK, pathToReport, releaseNumber, testMode);
```
jweiser: If the catch block is triggered, the state of the variables being set from the property file is likely inconsistent.  Maybe the program should be terminated if that occurs instead of printing a stack trace and continuing?

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -49,10 +56,13 @@ public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pa
 			instanceEditGK = UpdateDOIs.createInstanceEdit(UpdateDOIs.dbaGkCentral, authorIdGK, creatorFile);
 		}
 		// Gets the updated report file if it was provided for this release
-		Map<String, Map<String,String>> expectedUpdatedDOIs = UpdateDOIs.getExpectedUpdatedDOIs(pathToReport);
+		Map<String, Map<String,String>> expectedUpdatedDOIs = new HashMap<>();
+		File reportFile = new File(pathToReport);
+		if (reportFile.exists()) {
```
jweiser: You could use the `Files.exists()` method instead:

`if (Files.exists(Paths.get(pathToReport)))`

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#exists-java.nio.file.Path-java.nio.file.LinkOption...-

download-directory/Jenkinsfile
```diff
@@ -0,0 +1,78 @@
...
+	agent any
+
+	stages {
+		// This stage checks that an upstream project, AddLinks-Insertion, was run successfully for its last build.
+		stage('Check AddLinks-Insertion build succeeded'){
+			steps{
+				script{
+					def currentRelease = (pwd() =~ /Releases\/(\d+)\//)[0][1];
+					// This queries the Jenkins API to confirm that the most recent build of AddLinks-Insertion was successful.
+					def addLinksInsertionStatusUrl = httpRequest authentication: 'jenkinsKey', validResponseCodes: '200:404', url: "${env.JENKINS_JOB_URL}/job/$currentRelease/job/AddLinks-Insertion/lastBuild/api/json"
```
jweiser: These valid response codes look to be repeated in the PR.  Can they be consolidated to a single place?  Also, would other 2XX and 3XX codes be considered valid or only 200?

download-directory/Jenkinsfile
```diff
@@ -0,0 +1,78 @@
...
+					dir('download-directory'){
+						sh "mkdir -p archive/${env.RELEASE_NUMBER}/logs"
+						sh "mv --backup=numbered biopax*validator.zip archive/${env.RELEASE_NUMBER}/"
+						sh "gzip logs/*"
+						sh "mv logs/* archive/${env.RELEASE_NUMBER}/logs/"
+					}
+				}
+			}
+		}
+	}
```
jweiser: Often the procedure for the archiving of files as a post-step test is the same with just different files and starting directory.  Could a common function be created for the full pipeline and just overridden/tweaked by specific steps when needed?

download-directory/Jenkinsfile
```diff
@@ -0,0 +1,78 @@
...
+// It requires that the AddLinks-Insertion step has been run successfully before it can be run.
+pipeline {
+	agent any
+
+	stages {
+		// This stage checks that an upstream project, AddLinks-Insertion, was run successfully for its last build.
+		stage('Check AddLinks-Insertion build succeeded'){
+			steps{
+				script{
+					def currentRelease = (pwd() =~ /Releases\/(\d+)\//)[0][1];
```
jweiser: Does the value here differ from the environment variable `${env.RELEASE_NUMBER}`?

orthoinference/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+	def statusUrl = httpRequest authentication: 'jenkinsKey', validResponseCodes: '200:404', url: "${env.JENKINS_JOB_URL}/job/$currentRelease/job/$stepName/lastBuild/api/json"
+	if (statusUrl.getStatus() == 404) {
+		error("$stepName has not yet been run. Please complete a successful build.")
+	} else {
+		def statusJson = new JsonSlurper().parseText(statusUrl.getContent())
+		if(statusJson['result'] != "SUCCESS"){
+			error("Most recent $stepName build status: " + statusJson['result'] + ". Please complete a successful build.")
+		}
+	}
+}
```
jweiser: This is good!  

Can it be moved to a higher level so every step can access it?  It would help remove duplicated logic between other steps.

orthoinference/src/main/java/org/reactome/orthoinference/EWASInferrer.java
```diff
@@ -273,7 +273,7 @@ public static void setInstanceEdit(GKInstance instanceEditCopy)
 	public static void readENSGMappingFile(String toSpecies, String pathToOrthopairs) throws IOException
 	{
 		String mappingFileName = toSpecies + "_gene_protein_mapping.txt";
-		String mappingFilePath = pathToOrthopairs + mappingFileName;
+		String mappingFilePath = pathToOrthopairs + "/" + mappingFileName;
```
jweiser: See https://github.com/reactome/data-release-pipeline/pull/94/files#r339370886

Jenkinsfile-release-config
```diff
@@ -0,0 +1,54 @@
...
+		// After the user inputs, it executes a bash script that checks the values to confirm that they all match.
+		stage('Confirm configurations'){
+			steps{
+				// This asks for confirmation the configuration file has been updated for this release.
+				// This is a locally-scoped secret file accessed at Jenkins -> Releases -> ## (release number) -> Credentials -> Config
+				script{
+					def userInput = input(
+					id: 'userInput', message: "Has the Config file credentials been updated for this release?",
+					parameters: [
+												[$class: 'TextParameterDefinition', defaultValue: '', description: '', name: 'response']
```
jweiser: Just wondering why the description value is empty?

download-directory/Jenkinsfile
cookersjs: This is actually what was causing me a headache one day a few weeks ago. I was using the environmental variable, but there was a blind spot created. If the user hasn't updated the release number to match the actual current release, what would end up happening is that it would end up checking the build status for the step of the _previous_ release instead of the current one. 

By doing it based on the current directory, which the user would have had to create themselves to run release, it removes the chance of a person falsely running a step based on the success of the step in the previous release.

Jenkinsfile-release-config
cookersjs: Description variable doesn't ever get used, I think I was just copying examples from stack overflow and I left it like that. Chances are it can get removed, so I'll try that out

jweiser: Great!  Filling it in or removing it altogether would be nice.

download-directory/Jenkinsfile
jweiser: Ah okay, I see.  Perhaps the first step in the pipeline and/or a pre-step check common to all steps could compare the value obtained in the `currentRelease` variable, compare it to the environment variable, and throw an error if they are not the same?

That way the environment variable can always be used and it won't be a gotcha for any other places the environment variable is used.

cookersjs:
> For future work: Could the commit descriptions be a little more, well... descriptive? Everything for Jenkins is just `Update Jenkinsfile`. It's possible to have multi-line commit messages to say more than a few brief words, if it's necessary.
> 
> I find that I see a long list of "Update Jenkins" commits and I don't know which one to look at first.

@SolomonShorser-OICR Yea thats fair, I tried to make the commit descriptions more descriptive when it was a notable change. The one thing is that this was mostly completed in github since Jenkins clones the repository fresh each time. This meant that sometimes it would be veery small changes, like adding in a single character in order to get it to work, so I just used the default title a lot of the time (Update jenkinsfile). That being said, thats a reasonable request, so I'll try to make titles more descriptive in the futre

update-dois/Jenkinsfile
cookersjs: So the UpdateDOIs Test and Email are seperated because the email stage also has a User Input section, asking if the emailed contents look correct. This allows the user to answer yes/no without requiring them to re-run the UpdateDOIs code again. This is sort of trivial, considering that UpdateDOIs takes only 30 seconds, but when we apply this to things like orthoinference it will be better to allow the email stage to fail without needing to re-run orthoinference.

orthoinference/Jenkinsfile
cookersjs: You're good -- I bet you've knocked off a Where's Waldo or two in your day

download-directory/Jenkinsfile
cookersjs: The validResponseCodes are actually a property used by this httpRequest library, and it is actually meant to indicate the range of possible response codes that will be accepted. The default behaviour allowed for, I think, included all 2XX and 3XX ones, but not the 404 (why, is beyond me), which is why I needed to actually specify the acceptable range

cookersjs: Regarding consolidation, I believe they can be incorporated into an environmental variable

download-directory/Jenkinsfile
cookersjs: As discussed, for now we'll make an issue and then after release look into having globally available methods written in groovy

download-directory/Jenkinsfile
cookersjs: How it currently works is similar to that: Basically, each step has this check based on the working directory that should funnel you to the 'ConfirmReleaseConfigs' step, the now-first step of the release pipline when running it in Jenkins. ConfirmReleaseConfigs solicits your input for the release number, and then evaluates the environment's and the master configuration file's release number, confirming that all release number values are equal.

download-directory/Jenkinsfile
cookersjs: These have been changed to environmental variables

orthoinference/Jenkinsfile
```diff
@@ -31,7 +31,7 @@ pipeline{
 			}
 		}
 		// This stage builds the jar file using maven. It also runs the Main orthoinference process as a 'sub-stage'.
-		// This was due to a restriction in iterating over a list of species names. To iterate, you need to first have a 'stage > steps > script hierarchy.
+		// This was due to a restriction in iterating over a list of species names. To iterate, you need to first have a 'stage > steps > script' hierarchy.
```
jweiser: Thank you for humouring my nitpickiness!

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
```diff
@@ -15,7 +15,7 @@
 	
 	public static void main(String[] args) throws Exception {
 
-		String pathToConfig = "src/main/resources/config.properties";
+		String pathToConfig = Paths.get("src","main","resources","config.properties").toString();
```
jweiser: No need to change it, but whitespace between arguments would help readability.

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -61,6 +61,7 @@ public static void main( String[] args ) throws IOException {
       dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, portGK);
     } catch (Exception e) {
       e.printStackTrace();
+      System.exit(0);
```
jweiser: For an irregular program termination, a non-zero System exit might be better.  `System.exit(1)` is okay here or perhaps even just remove the try block and add the exception to the main method's `throws` list.

jweiser: That's true, but the effect should be the same by adding to the `throws` list in terms of an irregular exit and a stack trace being printed.

SolomonShorser-OICR: If the errors are not recoverable, then it's probably OK to just let them get thrown from `main`. On the other hand, in this case catching them and calling `System.exit(1)` is probably OK too, but there should be a more descriptive message, perhaps.

(In general, I don't like catching `Exception` - it's too broad. More specific exception types should be handled separately, when possible, but I know that many of the core Reactome libraries only throw `Exception` so there's not much we can do about that).

download-directory/Jenkinsfile
jweiser: Ah, okay I see.  Per our discussion, we could remove the release and previous release numbers from the environment variables and have them be determined through a function (and later on, possibly through a configuration local to the release run).

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
jweiser: In this case, I think it's a bit cleaner to add `SQLException` to the `throws` list since the functionality doesn't change and it removes some code and manual handling.

jweiser:
After transitioning the release and previous release variables (https://github.com/reactome/data-release-pipeline/pull/94#discussion_r340223390), I'm okay with merging!

download-directory/src/main/resources/config.properties
```diff
@@ -1,9 +1,11 @@
...
-host=localhost
-port=3306
-release=releaseNumber
+release.database.user=releaseDatabaseUsername
+release.database.password=releaseDatabasePassword
+release_current.name=release_current
+release.database.host=releaseDatabaseHost
+release.database.port=3306
+
+releaseNumber=
```
jweiser: Is there a mechanism to sync the release number obtained by Jenkins to be the same as here?

orthoinference/src/main/java/org/reactome/orthoinference/Main.java
```diff
@@ -15,7 +15,7 @@
 	
 	public static void main(String[] args) throws Exception {
 
-		String pathToConfig = Paths.get("src","main","resources","config.properties").toString();
+		String pathToConfig = Paths.get("src", "main", "resources", "config.properties").toString();
```
jweiser: :+1:

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
+releaseNumber=
+
+pathToSpeciesConfig=
+pathToOrthopairs=
+pathToWebsiteFilesUpdateFolder=
+pathToSkipList=
```
jweiser: Could this be more specific as to which skip list?

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
...
+dateOfRelease=yyyy-mm-dd
+
+pathToSpeciesConfig=
```
jweiser: Is this a duplicate of line 3?

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
...
+pantherFileURL=ftp://ftp.pantherdb.org/ortholog/current_release/
```
jweiser: Maybe more precise to name it as `pantherCurrentFileFolderURL`?

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
...
+orphanet.user=
+cosmic.user=
+brenda.user=
+orphanet.password=
+cosmic.password=
+dockblaster.password=
+dockblaster.user=
+
+absoluteReleaseDirectoryPath=
+releaseDownloadDirectoryPath=
```
jweiser: Based on the Slack conversation yesterday, can/should this go in the code and make use of `absoluteReleaseDirectoryPath` for the path to prefix the download directory path?

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
...
+protegeexporter.pathToWrapperScript=
+# 5 is for an 8-core system (need to leave some resources for MySQL and the OS).
+# If you have a different number of cores, you can change this value, but try to 
+# leave at least 2 or 3 cors available for other processes (such as MySQL and OS).
```
jweiser: Type - cores?

reactome-jenkins-release-config-template.properties
```diff
@@ -0,0 +1,54 @@
...
+dockblaster.user=
+
+absoluteReleaseDirectoryPath=
+releaseDownloadDirectoryPath=
+protegeexporter.pathToWrapperScript=
+# 5 is for an 8-core system (need to leave some resources for MySQL and the OS).
+# If you have a different number of cores, you can change this value, but try to 
+# leave at least 2 or 3 cors available for other processes (such as MySQL and OS). 
+protegeexporter.parallelism=5
+#protegeexporter.extraIncludes=-I/home/ubuntu/perl5/lib/perl5/,-I/home/$USER/perl5/lib/perl5/
```
jweiser: Why is this one commented out?

update-dois/Jenkinsfile
```diff
@@ -66,25 +69,27 @@ pipeline {
...
 				script{
 					emailext (
 						body: "This is an automated message. Please review the attached file of Pathway DOIs to be updated and confirm they are correct with the developer running release. Thanks!",
 						to: '$DEFAULT_RECIPIENTS',
-						subject: "UpdateDOIs List for v${env.RELEASE_NUMBER}",
-						attachmentsPattern: "**/doisToBeUpdated-v${env.RELEASE_NUMBER}.txt"
-					)
+						subject: "UpdateDOIs List for v${currentRelease}",
+						attachmentsPattern: "**///doisToBeUpdated-v${currentRelease}.txt" <- This pattern (**/) messses up multi-line comments. When uncommenting, remove 2/3 backslashes before 'doisToBeUpdated'
+		/*			)
```
jweiser: It looks like the `/*` opened on line 72 should still be in effect here?  Is this one needed?

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -59,13 +58,12 @@ public static void main( String[] args ) throws IOException {
       // Set up db connections.
       dbaTestReactome = new MySQLAdaptor(hostTR, databaseTR, userTR, passwordTR, portTR);
       dbaGkCentral = new MySQLAdaptor(hostGK, databaseGK, userGK, passwordGK, portGK);
-    } catch (Exception e) {
-      e.printStackTrace();
-      System.exit(1);
+    } catch (SQLException e) {
+      throw new SQLException("Unable to create MySQL adaptor. Check properties file.");
```
jweiser: Since the SQLException is thrown in the method signature, the try/catch block can be removed entirely.

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -47,8 +47,7 @@ public static void main( String[] args ) throws IOException {
       String hostGK = props.getProperty("curator.database.host");
       String databaseTR = props.getProperty("release_current.name");
       String databaseGK = props.getProperty("curator.database.name");
-      authorIdTR = Integer.valueOf(props.getProperty("personId"));
-      authorIdGK = Integer.valueOf(props.getProperty("personId"));
+      personId = Integer.valueOf(props.getProperty("personId"));
```
jweiser: The `personId` variable can be declared here instead of line 33 once the `try` block is removed.  

It also may be better parsed using `Long.parseLong(props.getProperty("personId"))` so a `long` value is returned rather than an `int` (though, Java will cast for you implicitly).

update-dois/src/main/java/org/reactome/release/updateDOIs/Main.java
```diff
@@ -47,8 +47,7 @@ public static void main( String[] args ) throws IOException {
...
       int portTR = Integer.valueOf(props.getProperty("release.database.port"));
       int portGK = Integer.valueOf(props.getProperty("curator.database.port"));
```
jweiser: It's preferable to use `Integer.parseInt(...)` for these lines since the values are being stored as primitive `int` values:

https://stackoverflow.com/a/7355056/2295778

update-dois/src/main/resources/config.properties
```diff
@@ -1,12 +1,14 @@
...
+release_current.name=release_current
+release.database.port=3306
+
+curator.database.user=centralDatabaseUsername
+curator.database.password=centralDatabasePassword
+curator.database.host=centralDatabaseHost
+curator.database.name=central
+curator.database.port=3306
+
+personId=8985221
```
jweiser: A comment that the `personId` should be changed to the db id of the Person instance for the one running the release would be good.

SolomonShorser-OICR: This is probably something that Jenkins should prompt the user for.

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
```diff
@@ -35,7 +35,7 @@ public static void setAdaptors(MySQLAdaptor adaptorTR, MySQLAdaptor adaptorGK) {
 	}
 
 	@SuppressWarnings("unchecked")
-	public static void findAndUpdateDOIs(long authorIdTR, long authorIdGK, String pathToReport, String releaseNumber, boolean testMode) throws IOException {
+	public static void findAndUpdateDOIs(long personId, String pathToReport, String releaseNumber, boolean testMode) throws IOException {
```
jweiser: Should `releaseNumber` be a typed as an `int`?

jweiser: `pathToReport` might be better typed as a `Path` object

download-directory/src/main/resources/config.properties
cookersjs: That is one of the things handled by the ConfirmReleaseConfigs steps

update-dois/src/main/resources/config.properties
cookersjs: Actually, the personId attribute should be left blank, that one just slipped through. I'll add comments

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
cookersjs: If you follow how releaseNumber is used, its actually just to generate the 'doisToBeUpdated-vXX' file that is meant to be sent to curators automatically, so keeping it as a String makes sense. Maybe a better way of doing it would be to generate the path and then pass that as an argument?

update-stable-ids/src/main/java/org/reactome/release/updateStableIds/Main.java
```diff
@@ -36,7 +36,7 @@ public static void main( String[] args ) throws Exception
        int releaseNumber = Integer.valueOf(props.getProperty("releaseNumber"));
        int prevReleaseNumber = releaseNumber - 1;
        
-       logger.info("Creating DB adaptors for test_slice_" + releaseNumber + ", test_slice_" + prevReleaseNumber + "_snapshot and gk_central");
+       logger.info("Creating DB adaptors");
```
jweiser: This would be a good log statement to include in a small private method that wraps the logic of extracting the db connection info from the properties file and then calling `new MySQLAdaptor`:

```
private createMySQLAdaptor(Properties props, String userProperty, String passwordProperty, String hostProperty, String databaseNameProperty, int portProperty) throw SQLException {
    String userName = props.getProperty(useProperty);
    String password = props.getProperty(passwordProperty);
    String host = props.getProperty(hostProperty);
    String databaseName = props.getProperty(databaseNameProperty);
    int port = Integer.parseInt(props.getProperty(portProperty));
    
    logger.info(
        "Creating DB adaptor for " + databaseName + " at " + host + " on port " + port + " for user " 
        + userName
    );

    return new MySQLAdaptor(host, databaseName, userName, password, port);
}
```

reactome-jenkins-release-config-template.properties
jweiser: Haha, don't call me out on my typos in talking about typos :).

update-dois/Jenkinsfile
jweiser: Ah okay, I didn't notice that.  Carry on!

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
jweiser: Okay, I see.  I'd still be inclined to make it an `int`.  Java will handle it properly when concatenated to a String and it makes it ensures a number is passed as the Reactome version.

reactome-jenkins-release-config-template.properties
SolomonShorser-OICR: :P

update-dois/src/main/java/org/reactome/release/updateDOIs/UpdateDOIs.java
SolomonShorser-OICR: Or rename the variable to something that suggests more than just a number?

reactome-jenkins-release-config-template.properties
cookersjs: I've made that change now

reactome-jenkins-release-config-template.properties
cookersjs: @SolomonShorser-OICR ?

SolomonShorser-OICR: See: https://github.com/reactome/data-release-pipeline/blob/ffb51885a7b2109bc1859e784673c0c345b6204e/downloadDirectory/src/main/java/org/reactome/release/downloadDirectory/ProtegeExporter.java#L74

(also search for it here: https://github.com/reactome/data-release-pipeline/blob/ffb51885a7b2109bc1859e784673c0c345b6204e/downloadDirectory/README.md)

That option was added primarily for testing, since my local Perl installation needed to be added to the Perl "-I" includes explicitly. It could be used for other people for testing, or on the server if necessary Perl files are not in the correct place.

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
```diff
@@ -41,8 +41,7 @@ public static void main(String[] args) throws Exception {
 		int port = Integer.valueOf(props.getProperty("release.database.port"));
 		String releaseNumber = props.getProperty("releaseNumber");
 		String releaseDirAbsolute = props.getProperty("absoluteReleaseDirectoryPath");
-		String releaseDownloadDir = props.getProperty("releaseDownloadDirectoryPath");
-		String releaseDownloadDirWithNumber = releaseDownloadDir + releaseNumber;
+		String releaseDownloadDirWithNumber = Paths.get("download_directory", releaseNumber).toString();
```
jweiser: Is this path supposed to be relative or does it need to be prefixed by the value in `releaseDirAbsolute`?

orthopairs/src/main/java/org/reactome/release/orthopairs/Main.java
```diff
@@ -54,7 +54,7 @@ public static void main( String[] args ) throws IOException, ParseException {
         // Load all config properties
         String releaseNumber = props.get("releaseNumber").toString();
         String pathToSpeciesConfig = props.get("pathToSpeciesConfig").toString();
-        String pantherFilepath = props.get("pantherFileURL").toString();
+        String pantherFilepath = props.get("pantherCurrentFilFoldereURL").toString();
```
jweiser: Typo - the "e" for file got put at the end of "Folder"

reactome-jenkins-release-config-template.properties
jweiser: Could a brief comment be included in the file to highlight this?

update-dois/src/test/java/org/reactome/release/updateDOIs/TestUpdateDOIs.java
```diff
@@ -49,7 +53,7 @@ public void testfindNewDOIsAndUpdate() throws Exception
 		Mockito.when(mockAdaptor.fetchInstanceByAttribute("Pathway", "doi", "NOT REGEXP", "^10.3180")).thenReturn(testResults);
 		Mockito.when(mockAdaptor.fetchInstanceByAttribute("Pathway", "DB_ID", "=", "67890")).thenReturn(testResults);
 
-		check.findAndUpdateDOIs(12345L, "reportPath", "70", true);
+		check.findAndUpdateDOIs(12345L, MOCK_REPORT_PATH, MOCK_RELEASE_NUMBER, true);
```
jweiser: Extracting `12345L` into a named constant would also be good for consistency, context, and re-use in the two times below.

download-directory/src/main/java/org/reactome/release/downloadDirectory/Main.java
cookersjs: Whoops, great catch

PR #95, opened 2019-10-04
Feature/chebi update fix no author

SolomonShorser-OICR:
Some ReferenceMolecules do not have a "created" attribute value so there is now a problem when we try to display the name of the person who created a ReferenceMolecule. When there is nothing to display in that case, "AUTHOR UNKNOWN" will be displayed.

### DIFF
--- a/chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
+++ b/chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
@@ -48,7 +48,7 @@ public class ChebiUpdater
 	private long personID;
 	private boolean useCache;
 	private Comparator<GKInstance> personComparator;
-	
+
 	/**
 	 * Create a ChebiUpdater
 	 * @param adaptor - The database adaptor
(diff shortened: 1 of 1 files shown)


chebi-update/src/main/java/org/reactome/release/chebiupdate/ChebiUpdater.java
```diff
@@ -416,12 +416,18 @@ private void updateReferenceEntities(GKInstance molecule, String chebiName, GKIn
...
+	 * If the instance does not have a "created" attribute, then NULL will be returned.
 	 * @throws InvalidAttributeException
 	 * @throws Exception
 	 */
 	private static GKInstance getCreator(GKInstance inst) throws InvalidAttributeException, Exception
 	{
 		GKInstance createdInstanceEdit = (GKInstance) inst.getAttributeValue(ReactomeJavaConstants.created);
+		if (createdInstanceEdit == null)
+		{
+			logger.error("Instance {} does not have a value for \"created\" attribute!", inst.toString());
```
cookersjs: Should this be a warn instead of an error?

SolomonShorser-OICR: Yes, probably.

PR #96, opened 2019-10-07
Feature/biomodels

cookersjs:
Finished version of the Biomodels data insertion step, rewritten in Java (credit to Joel) with README and unit tests.


biomodels/src/main/java/org/reactome/release/BioModelsUtilities.java
```diff
@@ -0,0 +1,105 @@
...
+            biomodelsReferenceDatabase.setDisplayName(bioModelsDatabaseName);
+
+            //TODO Check if both accessUrl and Url should be using https
+            biomodelsReferenceDatabase.addAttributeValue(
+                    ReactomeJavaConstants.accessUrl, "http://www.ebi.ac.uk/biomodels-main/publ-model.do?mid=###ID###"
+            );
+            biomodelsReferenceDatabase.addAttributeValue(
+                    ReactomeJavaConstants.url, "https://www.ebi.ac.uk/biomodels/"
+            );
+            dba.storeInstance(biomodelsReferenceDatabase);
```
SolomonShorser-OICR: I would recommend including an InstanceEdit object in the "created" attribute of the new ReferenceDatabase object.

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
```diff
@@ -0,0 +1,65 @@
...
+
+        try(BufferedReader br = new BufferedReader(new FileReader(tsvFile))) {
+            String line;
+            while ((line = br.readLine()) != null) {
+                String[] fields = line.split("\t");
+                String biomodelsId = fields[0];
+                String pathwayStableId = fields[1];
+
+                if (!matchBioModelsPattern(biomodelsId) || !matchStableIdPattern(pathwayStableId)) {
+                    logger.warn("Line has improperly formatted BioModel ID or Stable ID -- skipping");
```
SolomonShorser-OICR: Maybe test the conditions separately so the user knows exactly which one was the problem?

biomodels/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <version>1.6.5</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.reactome.server</groupId>
+            <artifactId>reactome-core</artifactId>
+            <version>1.1.2</version>
+        </dependency>
```
jweiser: I didn't realize it at the time, but it's probably good to replace this with the release-common-lib dependency

biomodels/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
+  </component>
+</project>
```
SolomonShorser-OICR: I think this is an IntelliJ-specific file. Should it be in github? Same for `biomodels/.idea/compiler.xml` (https://github.com/reactome/data-release-pipeline/pull/96/files#diff-09e35e8ccefda43d436317c68f659dd4)

biomodels/src/main/java/org/reactome/release/BioModelsUtilities.java
cookersjs: Good catch, thanks

biomodels/.idea/vcs.xml
cookersjs: Hmm, past Justin must have included this accidentally.

biomodels/README.md
```diff
@@ -0,0 +1,62 @@
+# BioModels Data Insertion Release Step
+
+The BioModels step is run following the <a href="https://github.com/reactome/analysis-core">analysis-core</a> and before UpdateGKCurrent (link needed) in Reactome's quarterly data release. It consists of two parts: (1) running the <a href="https://github.com/reactome/biomodels-mapper">biomodels-mapper</a> to generate `models2pathways.tsv` and than (2) running the code found in this repository to perform database updates. This is a rewrite of <a href="https://github.com/reactome/Release/blob/master/scripts/release/biomodels/add_links_to_biomodels_resource.pl">add_links_to_biomodels_resource.pl</a>
```
jweiser: Typo - "and then (2) running ..."?

biomodels/README.md
```diff
@@ -0,0 +1,62 @@
...
+BIOMD0000000287 R-HSA-69563     6.066213418698396E-6    https://reactome.org/PathwayBrowser/#/R-HSA-69563       p53-Dependent G1 DNA Damage Response    IEA     Homo sapiens
+BIOMD0000000287 R-HSA-2559586   6.066213418698396E-6    https://reactome.org/PathwayBrowser/#/R-HSA-2559586     DNA Damage/Telomere Stress Induced Senescence   IEA     Homo sapiens
+```
+
+For the purposes of this module, the first two columns are the most important. The first column contains BioModels identifiers, and the second column contains Reactome Pathway identifiers that have have been connected to them by biomodels-mapper. For each line in `models2pathways.tsv`, the corresponding Pathway instance in Reactome's relationship database will receive a CrossReference that points to the biomodel identifier. If the biomodel identifier does not already have a database instance, it will be created before being slotted into the Pathway instance as an instance of type `DatabaseIdentifier`. All biomodels identifier instances are connected to the sole 'BioModels Database' `ReferenceDatabase` object.
```
jweiser: For the biomodels-mapper mentioned at the end of the second sentence, it would be great to have a link to its GitHub repository: https://github.com/reactome/biomodels-mapper

biomodels/README.md
```diff
@@ -0,0 +1,62 @@
...
+Once this has been completed and saved somewhere on the server, you should be able to run the biomodels insertion step.
+  
+## Compiling & Running
+
+This is a Java application which requries a Java 8+ environment. You will need maven and a full JDK to compile.
+
+To compile the application, run this command:
+
+```
+$ mvn clean compile package
```
jweiser: It should be enough to run `mvn clean package`, I think.  Since the `package` life-cycle phase occurs after the compile phase, it should first do a compile followed by a package.

biomodels/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+If this is successful, you should see a JAR file in the `target` directory, with a name like `biomodels-VERSION_NUMBER-jar-with-dependencies.jar`. This is the file you will run.
+
+For example, for version "1.0-SNAPSHOT", execute this command to run the program:
+```
+$ java -jar target/biomodels-1.0-SNAPSHOT-jar-with-dependencies.jar ./config.properties path/to/models2pathways.tsv
+```
+
+If no properties file is specified as the first argument, the program will look in src/main/resources/ for config.properties. Similarly for models2pathways.tsv, unless otherwise specified it will assume it exists in src/main/resources.
```
jweiser: It's a little ambiguous here what will happen if the models2pathways.tsv file path is provided as the first and only argument.  Could the docs be updated to specify what happens and/or if this shouldn't be done?

As a separate GitHub issue or PR, maybe we can look into standardizing our data-release-pipeline jar artifacts to use named arguments.  It might even be a method or class we can create (for parsing arguments) in release-common-lib so we can have a standard that can be re-used.

biomodels/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Validating BioModels Insertion </h3>
+
+The insertion step generally takes a few minutes. Once it has completed, the simplest way to verify that it ran correctly is through Reactome's <a href="https://reactome.org/download-data/reactome-curator-tool">Curator Tool</a>. Once you have loaded up the database, first confirm that a 'BioModels Database' instance exists in the `ReferenceDatabase` class. This should have been created during the insertion step, if it didn't exist already. By right-clicking on the BioModels Database instance, you are able to 'Display Referrers'. This should bring up a list of `DatabaseIdentifier` objects that look like 'BioModels Database:BIOMD########'. Confirm that there are no duplicates in this list and compare the count in this release to the previous release. The numbers should be similar, if not a bit higher for the most recent one. That is everything that needs to be done for validation! Next up is the UpdateGKCurrent (link needed) step.
```
jweiser: It might be good to make the last sentence be its own paragraph so it stands out what the next step is.

biomodels/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.11.0</version>
+        </dependency>
+    </dependencies>
+
+    <repositories>
```
jweiser: I think it should be okay to remove the EBI repositories from here.  I think all the dependencies we need for this project are either local like release-common-lib or in Maven Central.

biomodels/src/main/java/org/reactome/release/BioModelsUtilities.java
```diff
@@ -0,0 +1,105 @@
...
+            biomodelsReferenceDatabase.addAttributeValue(ReactomeJavaConstants.created, instanceEdit);
+            String bioModelsDatabaseName = "BioModels Database";
+            biomodelsReferenceDatabase.addAttributeValue(
+                    ReactomeJavaConstants.name, Arrays.asList(bioModelsDatabaseName, "BioModels")
+            );
+            biomodelsReferenceDatabase.setDisplayName(bioModelsDatabaseName);
+
+            //TODO Check if both accessUrl and Url should be using https
+            biomodelsReferenceDatabase.addAttributeValue(
+                    ReactomeJavaConstants.accessUrl, "http://www.ebi.ac.uk/biomodels-main/publ-model.do?mid=###ID###"
```
jweiser: I just got an e-mail from Peter and Lisa and they found the BioModels access URL is now outdated.  It looks like this is the new one: https://www.ebi.ac.uk/biomodels/###ID### (e.g. https://www.ebi.ac.uk/biomodels/BIOMD0000000007)

biomodels/src/main/java/org/reactome/release/Main.java
```diff
@@ -0,0 +1,212 @@
...
+
+        for (GKInstance pathway: getPathwaysWithBiomodelsIds(dba, pathwayStableIdToBiomodelsIds.keySet())) {
+            logger.info("Adding BioModels ids to pathway " + pathway.getExtendedDisplayName());
+            List<String> biomodelsIds = pathwayStableIdToBiomodelsIds.get(getStableIdentifier(pathway));
+            List<GKInstance> biomodelsDatabaseIdentifiers =
+                createBiomodelsDatabaseIdentifiers(biomodelsIds, instanceEdit, dba);
+
+            try {
+                pathway.addAttributeValue(ReactomeJavaConstants.crossReference, biomodelsDatabaseIdentifiers);
+                pathway.addAttributeValue(ReactomeJavaConstants.modified, instanceEdit);
```
jweiser: Will this overwrite any existing modified instance edits?  I forget if it's necessary to load the existing attribute values first or not.

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private ModelsTSVParser() { }
+
+    /**
+     * Parses the contents of the models2pathways.tsv file, returning a Map of ReactomePathwayIds=[BioModelsIdentifiers,...]
+     * @param tsvFile -- String, path/to/models2pathways.tsv
+     * @return Map of ReactomePathwayIds and a List of BioModelsIdentifiers
+     */
+    public static Map<String, List<String>> parse(String tsvFile) {
+        Map<String, List<String>> pathwayToBiomodelsIds = new HashMap<>();
```
jweiser: To ensure there are no duplicates, a Set of Strings for the map's value might be better:
`Map<String, Set<String>> pathwayToBiomodelsIds = new HashMap<>()`

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
```diff
@@ -0,0 +1,65 @@
...
+                String biomodelsId = fields[0];
+                String pathwayStableId = fields[1];
+
+                if (!matchBioModelsPattern(biomodelsId) || !matchStableIdPattern(pathwayStableId)) {
+                    logger.warn("Line has improperly formatted BioModel ID or Stable ID -- skipping");
+                    continue;
+                }
+
+                if (pathwayToBiomodelsIds.get(pathwayStableId) == null) {
+                    List<String> biomodelsIds = new ArrayList<>();
```
jweiser: Once changed the mapping is changed to have a set as a value, this could be `Set<String> biomodels = new LinkedHashSet();` so that order is still preserved.

biomodels/src/test/java/org/reactome/release/BioModelsUtilitiesTester.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    public void bioModelsReferenceDatabaseDoesNotExistInDBReturnsNull() throws Exception {
+
+        Collection<GKInstance> collectionWithoutMockDatabaseInstance = new ArrayList<>();
+
+        Mockito.when(mockAdaptor.fetchInstanceByAttribute("ReferenceDatabase", "name", "=", "BioModels")).thenReturn(collectionWithoutMockDatabaseInstance);
+        GKInstance returnedDbInstance = BioModelsUtilities.retrieveBioModelsDatabaseInstance(mockAdaptor);
+        assertThat(returnedDbInstance, is(equalTo(null)));
```
jweiser: An alternative is `assertThat(returnedDbInstance, is(nullValue())`: https://stackoverflow.com/a/18987738/2295778

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
+package org.reactome.release;
+
+import org.junit.Test;
+
+import java.util.*;
+
+import static org.junit.Assert.*;
```
jweiser: Could the tests in this class be converted to their Hamcrest equivalents?

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    public void emptyMapFromUnknownFileName() {
+        Map<String, List<String>> pathwayToBioModelsIds =  ModelsTSVParser.parse("filedoesnotexist.tsv");
+        assertTrue("Unknown file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+    }
+
+    @Test
+    public void emptyMapFromImproperFile() {
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/improper_models2pathways.tsv");
```
jweiser: It might be good to set a resource directory class-level constant:
```
private final String RESOURCE_DIR = Paths.get("src", "test", "resources").toFile().getAbsolutePath();
```

You can then do a variable in each test method for the file you need:
```
String filePath = Paths.get(RESOURCE_DIR).toFile().getAbsolutePath();
```

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+        assertTrue("Improperly formatted file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+    }
+
+    @Test
+    public void populatedMapFromPartiallyCorrectFile() {
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/partiallycorrect_models2pathways.tsv");
```
jweiser: Similar to: https://github.com/reactome/data-release-pipeline/pull/96/files#r333326084

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+        assertEquals(pathwayToBioModelsIds.keySet().size(), 3);
+        assertTrue(pathwayToBioModelsIds.get("R-MMU-1169091").contains("BIOMD0000000226"));
+    }
+
+    @Test
+    public void testModels2PathwaysCorrectlyParsed() {
+        Map<String, List<String>> pathwayToBioModelsIds =
+            ModelsTSVParser.parse("src/test/resources/models2pathways.tsv");
```
jweiser: Similar to: https://github.com/reactome/data-release-pipeline/pull/96/files#r333326084

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+    @Test
+    public void emptyMapFromImproperFile() {
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/improper_models2pathways.tsv");
+        assertTrue("Improperly formatted file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+    }
+
+    @Test
+    public void populatedMapFromPartiallyCorrectFile() {
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/partiallycorrect_models2pathways.tsv");
+        assertEquals(pathwayToBioModelsIds.keySet().size(), 3);
```
jweiser: A named constant would be great for the value of `3` (e.g. correctPathwayCount)

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+        assertTrue(pathwayToBioModelsIds.get("R-MMU-1169091").contains("BIOMD0000000226"));
+    }
+
+    @Test
+    public void testModels2PathwaysCorrectlyParsed() {
+        Map<String, List<String>> pathwayToBioModelsIds =
+            ModelsTSVParser.parse("src/test/resources/models2pathways.tsv");
+        assertEquals(pathwayToBioModelsIds.keySet().size(), 745); // Number of pathways in test input file
```
jweiser: A named constant in place of the comment would be good

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+
+        long bioModelsIdsCount =
+            pathwayToBioModelsIds.keySet().stream().map(
+                pathway -> pathwayToBioModelsIds.get(pathway)
+            ).flatMap(
+                Collection::stream
+            ).distinct().count();
+        assertEquals(bioModelsIdsCount, 153); // Number of distinct BioModels Ids in test input file
```
jweiser: A named constant would be good here, too

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    public void populatedMapFromPartiallyCorrectFile() {
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/partiallycorrect_models2pathways.tsv");
+        assertEquals(pathwayToBioModelsIds.keySet().size(), 3);
+        assertTrue(pathwayToBioModelsIds.get("R-MMU-1169091").contains("BIOMD0000000226"));
+    }
+
+    @Test
+    public void testModels2PathwaysCorrectlyParsed() {
```
jweiser: Could this test be divided up into smaller test since there are many test conditions?  It's usually best practice to have one assert statement per test whenever possible.  Common setup logic could be extracted to its own private method if needed.

biomodels/README.md
cookersjs: I linked to it earlier in the README but couldn't hurt to link again

jweiser: Ah right - I didn't notice since the review shows the HTML instead of the formatted link.  It's probably okay not to link it again.  Maybe just add in brackets "see link above"?

biomodels/README.md
cookersjs: Created an issue for implementing arguments: https://github.com/reactome/data-release-pipeline/issues/101

biomodels/src/main/java/org/reactome/release/BioModelsUtilities.java
cookersjs: @SolomonShorser-OICR this is actually done at line 77

biomodels/README.md
cookersjs: After removing the EBI dependencies and installing release-common-lib locally (with the most up to date version that is seen on develop), it only produces a jar file without the dependencies included. I'm not entirely sure why, maybe this is all we need or I'm missing something? Maybe we can look at it tomorrow if you have time

cookersjs: That being said, there was no difference in `mvn clean package` and `mvn clean compile package`, so this is still correct

jweiser: Rhis might be a target that needs ‘mvn clean compile assembly:single’

biomodels/src/main/java/org/reactome/release/Main.java
cookersjs: You were correct, they needed to be loaded first. I think maybe we should conduct a thorough review of the new code to make sure this kind of thing isn't happening anywhere

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
cookersjs: This will require the code to get changed a bit more than what is suggested, so before I go ahead and make these changes, I'd like to point out that I have a commit for preventing duplicate identifiers being created: https://github.com/reactome/data-release-pipeline/pull/96/commits/d40d0f61065111aee17caf2c2fb3cf6d11a5c522

This uses the HashMap structure that is common in orthoinference to prevent duplicate instances from being created. If your overall goal is to prevent duplicate biomodels _instances_, then that has been handled in a different way. Otherwise, I will go ahead and make the suggested change

biomodels/README.md
cookersjs: Yes that worked

biomodels/README.md
cookersjs: I also updated the language of the README

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
cookersjs: See above comment: https://github.com/reactome/data-release-pipeline/pull/96#discussion_r338246134 -- I already have a mechanism in place for screening duplicate biomodel identifier instance, if that is what this suggested change is for

SolomonShorser-OICR:
I saw the IntelliJ files were removed :)
Could you please add something to a .gitignore so they don't get accidentally added again?

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
jweiser: I think this code change is still important.  The GKInstance created from an identifier won't be duplicated with the commit https://github.com/reactome/data-release-pipeline/commit/d40d0f61065111aee17caf2c2fb3cf6d11a5c522 but the List of String identifiers iterated over in the method `createBiomodelsDatabaseIdentifiers` could have the same String 2 or more times.  This would result in the unique GKInstance representing the BioModels identifier being added multiple times to the list returned from that method.  This could cause problems for the list of BioModels identifier instances in a pathway.  Using a Set of Strings prevents that.  IntelliJ should guide you wherever any changes are needed to change the data structure used from a list to a set after the type is changed on line 25.

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
jweiser: See https://github.com/reactome/data-release-pipeline/pull/96#discussion_r338725804

cookersjs:
@SolomonShorser-OICR @jweiser I believe all requested changes have been made for the PR

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
-    private final int correctNumberBioModelsIdsInTestFile = 153;
+    private final long correctNumberBioModelsIdsInTestFile = 153L;
+    private final String RESOURCE_DIR = Paths.get("src", "test", "resources").toFile().getAbsolutePath();
+
     @Test
-    public void emptyMapFromEmptyFileName() {
+    public void emptyFilenameReturnsEmptyMap() {
         Map<String, List<String>> pathwayToBioModelsIds =  ModelsTSVParser.parse("");
-        assertTrue("Empty file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+        assertThat("Empty file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty(), is(true));
```
jweiser: A more standard way to write this may be:

`assertThat("Empty file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds, is(anEmptyMap()));`

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
     }
 
     @Test
-    public void emptyMapFromUnknownFileName() {
+    public void unknownFileNameReturnsEmptyMap() {
+        String filepath = Paths.get(RESOURCE_DIR, "filedoesnotexist.tsv").toFile().getAbsolutePath();
         Map<String, List<String>> pathwayToBioModelsIds =  ModelsTSVParser.parse("filedoesnotexist.tsv");
-        assertTrue("Unknown file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+        assertThat("Unknown file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty(), is(true));
```
jweiser: See https://github.com/reactome/data-release-pipeline/pull/96/files/3f52d04f54bfc64f3838bdbbee390c80a851f32e..84ef2a5cc83900db50e239d7834951eeea7e8b48#r339374370

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
     }
 
     @Test
-    public void emptyMapFromImproperFile() {
-        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/improper_models2pathways.tsv");
-        assertTrue("Improperly formatted file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty());
+    public void improperlyFormattedFileReturnsEmptyMap() {
+        String filepath = Paths.get(RESOURCE_DIR, "improper_models2pathways.tsv").toFile().getAbsolutePath();
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse(filepath);
+        assertThat("Improperly formatted file produces empty pathwayToBioModelsIds map", pathwayToBioModelsIds.isEmpty(), is(true));
```
jweiser: https://github.com/reactome/data-release-pipeline/pull/96/files/3f52d04f54bfc64f3838bdbbee390c80a851f32e..84ef2a5cc83900db50e239d7834951eeea7e8b48#r339374370

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
     @Test
-    public void populatedMapFromPartiallyCorrectFile() {
-        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse("src/test/resources/partiallycorrect_models2pathways.tsv");
-        assertEquals(pathwayToBioModelsIds.keySet().size(), correctPathwayCount);
-        assertTrue(pathwayToBioModelsIds.get("R-MMU-1169091").contains("BIOMD0000000226"));
+    public void partiallyCorrectFileReturnsPopulatedMap() {
+        String filepath = Paths.get(RESOURCE_DIR, "partiallycorrect_models2pathways.tsv").toFile().getAbsolutePath();
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse(filepath);
+        assertThat(pathwayToBioModelsIds.keySet().size(), is(equalTo(correctPathwayCount)));
+        assertThat(pathwayToBioModelsIds.get("R-MMU-1169091").contains("BIOMD0000000226"), is(equalTo(true)));
```
jweiser: Try to avoid testing against a literal boolean value if possible.  Hamcrest usually has matchers for the common scenarios to test like here to see if a value is contained in a list.  In this case, the assert can be written as:

`assertThat(pathwayToBioModelsIds.get("R-MMU-1169091"), contains("BIOMD0000000226"));`

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
+        assertThat(bioModelsIdsCount, is(equalTo(correctNumberBioModelsIdsInTestFile))); // Number of distinct BioModels Ids in test input file
+    }
 
+    @Test
+    public void testFileContainsCorrectMapping() {
+        String filepath = Paths.get(RESOURCE_DIR, "models2pathways.tsv").toFile().getAbsolutePath();
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse(filepath);
         // Tests pathway contains correct BioModels id
-        assertTrue(pathwayToBioModelsIds.get("R-HSA-199418").contains("BIOMD0000000175"));
+        assertThat(pathwayToBioModelsIds.get("R-HSA-199418").contains("BIOMD0000000175"), is(equalTo(true)));
```
jweiser: See https://github.com/reactome/data-release-pipeline/pull/96/files/3f52d04f54bfc64f3838bdbbee390c80a851f32e..84ef2a5cc83900db50e239d7834951eeea7e8b48#r339374612

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
+    }
 
+    @Test
+    public void testFileProducesMappingWithoutNullKeysOrValues() {
+        String filepath = Paths.get(RESOURCE_DIR, "models2pathways.tsv").toFile().getAbsolutePath();
+        Map<String, List<String>> pathwayToBioModelsIds = ModelsTSVParser.parse(filepath);
         // Tests for null in Map keys and value
-        assertTrue("pathwayToBioModelsIds map does not contain any null keys", !pathwayToBioModelsIds.containsKey(null));
-        assertTrue("pathwayToBioModelsIds map does not contain any null values", !pathwayToBioModelsIds.containsValue(null));
+        assertThat("pathwayToBioModelsIds map does not contain any null keys", pathwayToBioModelsIds.containsKey(null), is(equalTo(false)));
```
jweiser: This can (theoretically -- I haven't tested :) be:

`assertThat("pathwayToBioModelsIds map does not contain any null keys", pathwayToBioModelsIds, not(hasKey(nullValue())));`

biomodels/src/test/java/org/reactome/release/ModelsTSVParserTester.java
```diff
@@ -2,68 +2,95 @@
...
+        assertThat("pathwayToBioModelsIds map does not contain any null values", pathwayToBioModelsIds.containsValue(null), is(equalTo(false)));
```
jweiser: This can (theoretically -- I haven't tested :) be:

`assertThat("pathwayToBioModelsIds map does not contain any null keys", pathwayToBioModelsIds, not(hasValue(nullValue())));`

biomodels/src/main/java/org/reactome/release/Main.java
jweiser: Could a Unit Test be added for this?  

You could extract the `try` block to a method that takes a pathway instance, the list of biomodelsDatabaseIdentifiers, and the instance edit as the three parameters.

Then test by sending a mock pathway instance that already has one modified instance edit which finishes with two modified instance edits afterwards?

biomodels/src/test/java/org/reactome/release/BioModelsUtilitiesTester.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void bioModelsReferenceDatabaseExistsInDBReturnsInstance() throws Exception {
+
+        Collection<GKInstance> collectionWithMockDatabaseInstance = new ArrayList<>();
+        collectionWithMockDatabaseInstance.add(mockBiomodelsDatabase);
+
+        Mockito.when(mockAdaptor.fetchInstanceByAttribute("ReferenceDatabase", "name", "=", "BioModels")).thenReturn(collectionWithMockDatabaseInstance);
+        GKInstance returnedDbInstance = BioModelsUtilities.retrieveBioModelsDatabaseInstance(mockAdaptor);
+        assertThat(returnedDbInstance, is(equalTo(mockBiomodelsDatabase)));
```
jweiser: Minor style point for tests - if possible, keep an empty line between the three distinct sections of the test method body (arrange, act, assert).  It helps readability to see at a glance where the setup is, where the method being tested is called, and where the assert for the test is.

jweiser:
Changes look good.  I left some requests for some style-related points, but no blockers.

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
cookersjs: OK I've made the change. ModelsTSVParser now returns a Map<String, Set<String>>. This Set is then transformed into a List when creating the DatabaseIdentifiers for the BioModels IDs, which keeps the code mostly the same.

biomodels/src/main/java/org/reactome/release/ModelsTSVParser.java
cookersjs: Added in the LinkedHashSet as well

Issue #83, opened 2019-08-26
Orthoinference report needs a header to identify the Reactome version


jweiser:
The `report_ortho_inference.txt` file under `/usr/local/reactomes/Reactome/production/Release/scripts/release/website_files_update` needs a header such as:

`# Reactome Release Version 70`

This would make sure anyone viewing the statistics knows which version they apply to.

Issue #31, opened 2019-03-11
Orthoinference needs to be interruption proof


cookersjs:
During last release, it came up that Orthoinference might be interruption-proof for ReactionlikeEvents, but not for Pathways. This needs to be checked out so that in the future we don't need to worry about re-running it

PR #104, opened 2019-10-17
Feature/orthoinference 71 updates

cookersjs:
This PR addresses a couple of things that had been on the backburner for Orthoinference:

1) _Most_ PhysicalEntity instances are now screened before any inference is attempted on a ReactionlikeEvent. The only cases that aren't screened ahead of time are some edge cases for EntitySets that were very difficult to isolate out ahead of time. Safeguards still exist for preventing inference, so the ones that do squeak through this new screening system will still be rejected. This will prevent a significant majority of orphan PEs.

2) Orthoinference is now interruptable. If the process is stopped in the middle of a species' inference, it can be restarted and it will still create Pathways and PathwayDiagrams for any _automatically_ inferred RlEs. **A question**: Should manually inferred ones also be receiving Pathway and PathwayDiagram instances? Its a very simple adjustment if so. This addresses #31 

3) The report file produced at the end of a species' inference now has a header that denotes the release. This addresses #83 

Thanks!

### DIFF
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -49,8 +49,8 @@ public class EventsInferrer
 	private static String releaseVersion;
 	private static GKInstance instanceEditInst;
 	private static GKInstance speciesInst;
-	private static Map<GKInstance,GKInstance> manualEventToNonHumanSource = new HashMap<>();
-	private static List<GKInstance> manualHumanEvents = new ArrayList<>();
+	private static Map<GKInstance, GKInstance> eventsAlreadyInferredMap = new HashMap<>();
+	private static List<GKInstance> eventsAlreadyInferred = new ArrayList<>();
 	private static StableIdentifierGenerator stableIdentifierGenerator;
 	private static OrthologousPathwayDiagramGenerator orthologousPathwayDiagramGenerator;
 
@@ -166,15 +166,15 @@ public class EventsInferrer
 			logger.info("Attempting RlE inference: " + reactionInst);
 			// Check if the current Reaction already exists for this species, that it is a valid instance (passes some filters), and that it doesn't have a Disease attribute.
 			// Adds to manualHumanEvents array if it passes conditions. This code block allows you to re-run the code without re-inferring instances.
-			List<GKInstance> previouslyInferredInstances = new ArrayList<GKInstance>();
+			List<GKInstance> previouslyInferredInstances = new ArrayList<>();
 			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, orthologousEvent, previouslyInferredInstances);
-			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances);
+			previouslyInferredInstances.addAll(checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances));
 			if (previouslyInferredInstances.size() > 0) {
 				GKInstance prevInfInst = previouslyInferredInstances.get(0);
 				if (prevInfInst.getAttributeValue(disease) == null) {
 					logger.info("Inferred RlE already exists, skipping inference");
-					manualEventToNonHumanSource.put(reactionInst, prevInfInst);
-					manualHumanEvents.add(reactionInst);
+					eventsAlreadyInferredMap.put(reactionInst, prevInfInst);
+					eventsAlreadyInferred.add(reactionInst);
 				} else {
 					logger.info("Disease reaction, skipping inference");
 				}
@@ -190,8 +190,11 @@ public class EventsInferrer
 				return;
 			}
 		}
-		PathwaysInferrer.setInferredEvent(ReactionInferrer.getInferredEvent());
-		PathwaysInferrer.inferPathways(ReactionInferrer.getInferrableHumanEvents());
+		// Retrieve events inferred from this run, and any that were already inferred. Combine them and then begin Pathway inference.
+		// The two methods below perform this for a map, containing the original RlE and the inferred RlE, and a List of just the inferred RlEs.
+		// The latter will be iterated through when building Pathway hierarchies, the former when information from original RlE is needed during this build.
+		PathwaysInferrer.setInferredEvent(ReactionInferrer.getInferredEvent(eventsAlreadyInferredMap));
+		PathwaysInferrer.inferPathways(ReactionInferrer.getInferrableHumanEvents(eventsAlreadyInferred));
 		orthologousPathwayDiagramGenerator.generateOrthologousPathwayDiagrams();
 		outputReport(species);
 		logger.info("Finished orthoinference of " + speciesName);
@@ -217,7 +220,7 @@ public class EventsInferrer
 	}
 
 	@SuppressWarnings("unchecked")
-	private static List<GKInstance> checkIfPreviouslyInferred(GKInstance reactionInst, String attribute, List<GKInstance> previouslyInferredInstances) throws InvalidAttributeException, Exception 
+	private static List<GKInstance> checkIfPreviouslyInferred(GKInstance reactionInst, String attribute, List<GKInstance> previouslyInferredInstances) throws Exception
 	{
 		for (GKInstance attributeInst : (Collection<GKInstance>) reactionInst.getAttributeValuesList(attribute))
 		{
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -113,6 +113,7 @@ public class EventsInferrer
 			Map<String,String[]> homologueMappings = readHomologueMappingFile(species, "hsap", pathToOrthopairs);
 			ProteinCountUtility.setHomologueMappingFile(homologueMappings);
 			EWASInferrer.setHomologueMappingFile(homologueMappings);
+			SkipInstanceChecker.setHomologueMappingFile(homologueMappings);
 		} catch (FileNotFoundException e) {
 			logger.fatal("Unable to locate " + speciesName +" mapping file: hsap_" + species + "_mapping.txt. Orthology prediction not possible.");
 			return;
@@ -168,11 +169,9 @@ public class EventsInferrer
 			List<GKInstance> previouslyInferredInstances = new ArrayList<GKInstance>();
 			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, orthologousEvent, previouslyInferredInstances);
 			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances);
-			if (previouslyInferredInstances.size() > 0)
-			{
+			if (previouslyInferredInstances.size() > 0) {
 				GKInstance prevInfInst = previouslyInferredInstances.get(0);
-				if (prevInfInst.getAttributeValue(disease) == null)
-				{
+				if (prevInfInst.getAttributeValue(disease) == null) {
 					logger.info("Inferred RlE already exists, skipping inference");
 					manualEventToNonHumanSource.put(reactionInst, prevInfInst);
 					manualHumanEvents.add(reactionInst);
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -53,6 +53,8 @@ public class EventsInferrer
 	private static List<GKInstance> eventsAlreadyInferred = new ArrayList<>();
 	private static StableIdentifierGenerator stableIdentifierGenerator;
 	private static OrthologousPathwayDiagramGenerator orthologousPathwayDiagramGenerator;
+	private static final String summationText = "This event has been computationally inferred from an event that has been demonstrated in another species.<p>The inference is based on the homology mapping from PANTHER. Briefly, reactions for which all involved PhysicalEntities (in input, output and catalyst) have a mapped orthologue/paralogue (for complexes at least 75% of components must have a mapping) are inferred to the other species. High level events are also inferred for these events to allow for easier navigation.<p><a href='/electronic_inference_compara.html' target = 'NEW'>More details and caveats of the event inference in Reactome.</a> For details on PANTHER see also: <a href='http://www.pantherdb.org/about.jsp' target='NEW'>http://www.pantherdb.org/about.jsp</a>";
+
 
 	@SuppressWarnings("unchecked")
 	public static void inferEvents(Properties props, String species) throws Exception
@@ -171,12 +173,14 @@ public class EventsInferrer
 			previouslyInferredInstances.addAll(checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances));
 			if (previouslyInferredInstances.size() > 0) {
 				GKInstance prevInfInst = previouslyInferredInstances.get(0);
-				if (prevInfInst.getAttributeValue(disease) == null) {
+				GKInstance prevInfSummationInst = (GKInstance) prevInfInst.getAttributeValue(summation);
+				String prevInfSummationText = prevInfSummationInst.getAttributeValue(text).toString();
+				if (prevInfInst.getAttributeValue(disease) == null &&  prevInfSummationText.equals(summationText)) {
 					logger.info("Inferred RlE already exists, skipping inference");
 					eventsAlreadyInferredMap.put(reactionInst, prevInfInst);
 					eventsAlreadyInferred.add(reactionInst);
 				} else {
-					logger.info("Disease reaction, skipping inference");
+					logger.info("Either a disease or manually inferred reaction, skipping inference");
 				}
 				continue;
 			}
@@ -304,7 +308,6 @@ public class EventsInferrer
 		GKInstance summationInst = new GKInstance(dbAdaptor.getSchema().getClassByName(Summation));
 		summationInst.setDbAdaptor(dbAdaptor);
 		summationInst.addAttributeValue(created, instanceEditInst);
-		String summationText = "This event has been computationally inferred from an event that has been demonstrated in another species.<p>The inference is based on the homology mapping from PANTHER. Briefly, reactions for which all involved PhysicalEntities (in input, output and catalyst) have a mapped orthologue/paralogue (for complexes at least 75% of components must have a mapping) are inferred to the other species. High level events are also inferred for these events to allow for easier navigation.<p><a href='/electronic_inference_compara.html' target = 'NEW'>More details and caveats of the event inference in Reactome.</a> For details on PANTHER see also: <a href='http://www.pantherdb.org/about.jsp' target='NEW'>http://www.pantherdb.org/about.jsp</a>";
 		summationInst.addAttributeValue(text, summationText);
 		summationInst.addAttributeValue(_displayName, summationText);
 		summationInst = InstanceUtilities.checkForIdenticalInstances(summationInst, null);
--- a/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
@@ -105,7 +105,7 @@ public class EventsInferrer
 		String inferredFilename = "inferred_" + species + "_75.txt";
 		createNewFile(eligibleFilename);
 		createNewFile(inferredFilename);
-		ReactionInferrer.setEligibleFilename(eligibleFilename);
+		SkipInstanceChecker.setEligibleFilename(eligibleFilename);
 		ReactionInferrer.setInferredFilename(inferredFilename);
 
 		stableIdentifierGenerator = new StableIdentifierGenerator(dbAdaptor, (String) speciesObject.get("abbreviation"));
@@ -239,7 +239,7 @@ public class EventsInferrer
 
 	private static void outputReport(String species) throws IOException
 	{
-		int eligibleCount = ReactionInferrer.getEligibleCount();
+		int eligibleCount = SkipInstanceChecker.getEligibleCount();
 		int inferredCount = ReactionInferrer.getInferredCount();
 		float percentInferred = (float) 100*inferredCount/eligibleCount;
 		// Create file if it doesn't exist
@@ -247,8 +247,10 @@ public class EventsInferrer
 		logger.info("Updating " + reportFilename);
 		if (!Files.exists(Paths.get(reportFilename))) {
 			createNewFile(reportFilename);
+			String reportHeader = "## Number of inferred reactions by species for Reactome Release " + releaseVersion;
+			Files.write(Paths.get(reportFilename), reportHeader.getBytes(), StandardOpenOption.APPEND);
 		}
-		String results = "hsap to " + species + ":\t" + inferredCount + " out of " + eligibleCount + " eligible reactions (" + String.format("%.2f", percentInferred) + "%)\n";
+		String results = "hsap to " + species + ":\tInferred " + inferredCount + " out of " + eligibleCount + " eligible reactions (" + String.format("%.2f", percentInferred) + "%)\n";
 		Files.write(Paths.get(reportFilename), results.getBytes(), StandardOpenOption.APPEND);
 	}
 	
--- a/orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
+++ b/orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
@@ -47,8 +47,6 @@ public class PathwaysInferrer {
 		addInferredEventsToInferredPathways();
 		logger.info("Finished populating inferred Pathways with inferred Events");
 
-		//TODO: LOG starting HERE
-
 		// Connect preceding events to RlEs, if they have any in the source species.
 		logger.info("Adding preceding events to inferred Events");
 		inferPrecedingEvents();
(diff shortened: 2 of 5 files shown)


orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -49,10 +49,12 @@
...
-	private static List<GKInstance> manualHumanEvents = new ArrayList<>();
+	private static Map<GKInstance, GKInstance> eventsAlreadyInferredMap = new HashMap<>();
+	private static List<GKInstance> eventsAlreadyInferred = new ArrayList<>();
 	private static StableIdentifierGenerator stableIdentifierGenerator;
 	private static OrthologousPathwayDiagramGenerator orthologousPathwayDiagramGenerator;
+	private static final String summationText = "This event has been computationally inferred from an event that has been demonstrated in another species.<p>The inference is based on the homology mapping from PANTHER. Briefly, reactions for which all involved PhysicalEntities (in input, output and catalyst) have a mapped orthologue/paralogue (for complexes at least 75% of components must have a mapping) are inferred to the other species. High level events are also inferred for these events to allow for easier navigation.<p><a href='/electronic_inference_compara.html' target = 'NEW'>More details and caveats of the event inference in Reactome.</a> For details on PANTHER see also: <a href='http://www.pantherdb.org/about.jsp' target='NEW'>http://www.pantherdb.org/about.jsp</a>";
```
jweiser: I'm not sure the "target = 'NEW'" attribute that is used two times is allowed in modern HTML.  For a new window or tab, 'target=_blank' is the attribute value to use:

https://www.w3schools.com/tags/att_a_target.asp

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -113,6 +115,7 @@ public static void inferEvents(Properties props, String species) throws Exceptio
 			Map<String,String[]> homologueMappings = readHomologueMappingFile(species, "hsap", pathToOrthopairs);
 			ProteinCountUtility.setHomologueMappingFile(homologueMappings);
 			EWASInferrer.setHomologueMappingFile(homologueMappings);
+			SkipInstanceChecker.setHomologueMappingFile(homologueMappings);
```
jweiser: The name `setHomologueMappingFile` is a little misleading since it is taking the homologue map data structure itself rather than the file name.  Could it be `setHomologueMap` instead?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
SolomonShorser-OICR: Could this string be split on to several lines? This is a bit wide, even for me.

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -168,40 +166,31 @@ private static void addInferredEventsToInferredPathways() throws Exception {
...
 		Set<GKInstance> seenPrecedingEvent = new HashSet<>();
-		for (GKInstance inferrableEventInst : updatedInferrableHumanEvents)
-		{
-			if (!seenPrecedingEvent.contains(inferrableEventInst))
-			{
-				if (inferrableEventInst.getAttributeValue(precedingEvent)!= null)
-				{
+		for (GKInstance inferrableEventInst : updatedInferrableHumanEvents) {
+			if (!seenPrecedingEvent.contains(inferrableEventInst)) {
+				if (inferrableEventInst.getAttributeValue(precedingEvent) != null) {
```
SolomonShorser-OICR: Could these conditions be merged into one if-statement as `!seenPrecedingEvent.contains(inferrableEventInst) && inferrableEventInst.getAttributeValue(precedingEvent) != null` ?
I don't see an else-statement pairing with the inner if-statement so it seems like it should be OK, and would make the rest of this a little more readable.

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -52,87 +51,72 @@ public static void inferReaction(GKInstance reactionInst) throws Exception
...
-							{
-								infReactionInst.addAttributeValue(releaseDate, dateOfRelease);
-							}
-							// FetchIdenticalInstances would just return the instance being inferred. Since this step is meant to always
-							// add a new inferred instance, the storeInstance method is just called here.
-							GKInstance orthoStableIdentifierInst = EventsInferrer.getStableIdentifierGenerator().generateOrthologousStableId(infReactionInst, reactionInst);
-							infReactionInst.addAttributeValue(stableIdentifier, orthoStableIdentifierInst);
-							dba.storeInstance(infReactionInst);
-							logger.info("Inferred RlE instance: " + infReactionInst);
+							return;
```
SolomonShorser-OICR: Is this an if-statement that only contains a `return`? Could this be modified to be the _negation_ of the same condition which _allows_ the rest of the code to flow? This is a void function, so it's not even expected to return anything. I find that using `return` as a form of flow-control to be confusing.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -165,19 +168,19 @@ public static void inferEvents(Properties props, String species) throws Exceptio
 			logger.info("Attempting RlE inference: " + reactionInst);
 			// Check if the current Reaction already exists for this species, that it is a valid instance (passes some filters), and that it doesn't have a Disease attribute.
 			// Adds to manualHumanEvents array if it passes conditions. This code block allows you to re-run the code without re-inferring instances.
-			List<GKInstance> previouslyInferredInstances = new ArrayList<GKInstance>();
+			List<GKInstance> previouslyInferredInstances = new ArrayList<>();
 			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, orthologousEvent, previouslyInferredInstances);
-			previouslyInferredInstances = checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances);
-			if (previouslyInferredInstances.size() > 0)
-			{
+			previouslyInferredInstances.addAll(checkIfPreviouslyInferred(reactionInst, inferredFrom, previouslyInferredInstances));
```
jweiser: Since `previouslyInferredInstances` is only added to and not queried in the method `checkIfPreviouslyInferred` passing it as a parameter isn't needed.  It's best practice not to modify parameters, but instead to create and return a different variable in the method body.  

You could do this part of the code as follows:

```
List<GKInstance> previouslyInferredInstances = checkIfPreviouslyInferred(
    reactionInst, Arrays.asList(inferredFrom, orthologousEvent)
);
```

Then the `checkIfPreviouslyInferred` method could be:

```
private static List<GKInstance> checkIfPreviouslyInferred(GKInstance reactionInst, List<String> attributes) throws Exception 
{
	List previouslyInferredInstances = new ArrayList<>();
	for (String attribute : attributes) {
		for (GKInstance attributeInst : (Collection<GKInstance>) reactionInst.getAttributeValuesList(attribute)) {
			GKInstance reactionSpeciesInst = (GKInstance) attributeInst.getAttributeValue(species);
			if (reactionSpeciesInst.getDBID().equals(speciesInst.getDBID()) && attributeInst.getAttributeValue(isChimeric) == null) {
				previouslyInferredInstances.add(attributeInst);
			}
		}
	}
	return previouslyInferredInstances;
}
```

jweiser: Also, it may be better to rename `checkIfPreviouslyInferred` to `getInstancesIfPreviouslyInferred` to make it clear that the list of previously inferred instances is what the method returns.  A method name starting with "check" may be more suitable to a void or possibly a boolean return value.

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -165,19 +168,19 @@ public static void inferEvents(Properties props, String species) throws Exceptio
...
+			if (previouslyInferredInstances.size() > 0) {
 				GKInstance prevInfInst = previouslyInferredInstances.get(0);
```
jweiser: Is a check needed to see if the list contains more than one element?  If the list has more than one previously inferred instances would that be an error?

orthoinference/src/main/java/org/reactome/orthoinference/EventsInferrer.java
```diff
@@ -165,19 +168,19 @@ public static void inferEvents(Properties props, String species) throws Exceptio
...
-				if (prevInfInst.getAttributeValue(disease) == null)
-				{
+				GKInstance prevInfSummationInst = (GKInstance) prevInfInst.getAttributeValue(summation);
+				String prevInfSummationText = prevInfSummationInst.getAttributeValue(text).toString();
+				if (prevInfInst.getAttributeValue(disease) == null &&  prevInfSummationText.equals(summationText)) {
 					logger.info("Inferred RlE already exists, skipping inference");
-					manualEventToNonHumanSource.put(reactionInst, prevInfInst);
-					manualHumanEvents.add(reactionInst);
+					eventsAlreadyInferredMap.put(reactionInst, prevInfInst);
+					eventsAlreadyInferred.add(reactionInst);
```
jweiser: Are the `eventsAlreadyInferredMap` and `eventsAlreadyInferred` data structures both needed?  Could the list populated in `eventsAlreadyInferred` be obtained from the key set of the `eventsAlreadyInferredMap`?

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
jweiser: These two conditions could be combined with `&&` if `seenPrecedingEvent.add(inferrableEventInst);` (line 198) is moved outside the inner if condition.  Since `seenPrecedingEvent` is a set, it won't cause a problem if you add an instance that is already present in the set.

jweiser: You may also want to consider refactoring the body of the if statement to its own method so that you have the operation occurring for each `inferrableEventInst` isolated.

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -168,40 +166,31 @@ private static void addInferredEventsToInferredPathways() throws Exception {
...
 						inferredPrecedingEvents.add(precedingEventInst.getDBID().toString());
 					}
 					List<GKInstance> updatedPrecedingEventInstances = new ArrayList<>();
 					// Find existing preceding events that haven't already been attached to the inferred instance
-					for (GKInstance precedingEventInst : precedingEventInstances)
-					{
-						if (!inferredPrecedingEvents.contains(precedingEventInst.getDBID().toString()))
-						{
+					for (GKInstance precedingEventInst : precedingEventInstances) {
+						if (!inferredPrecedingEvents.contains(precedingEventInst.getDBID().toString())) {
```
jweiser: Could `inferredPrecedingEvents` be typed as Set<Long> instead of Set<String> so a call to `toString()` isn't needed?

orthoinference/src/main/java/org/reactome/orthoinference/PathwaysInferrer.java
```diff
@@ -168,40 +166,31 @@ private static void addInferredEventsToInferredPathways() throws Exception {
...
 							updatedPrecedingEventInstances.add(precedingEventInst);
 						}
 					}
 					// Add preceding event to inferred instance
-					if (updatedPrecedingEventInstances != null && updatedPrecedingEventInstances.size() > 0)
-					{
+					if (updatedPrecedingEventInstances != null && updatedPrecedingEventInstances.size() > 0) {
```
jweiser: The null check shouldn't be necessary since `updatingPrecedingEventInstances` is initialized as an empty ArrayList.

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -52,87 +51,72 @@ public static void inferReaction(GKInstance reactionInst) throws Exception
...
-				String eligibleEventName = reactionInst.getAttributeValue(DB_ID).toString() + "\t" + reactionInst.getDisplayName() + "\n";	
-				// Having passed all tests/filters until now, the reaction is recorded in the 'eligible reactions' file, meaning inference is continued.
-				eligibleCount++;
-				Files.write(Paths.get(eligibleFilehandle), eligibleEventName.getBytes(), StandardOpenOption.APPEND);
-				// Attempt to infer all PhysicalEntities associated with this reaction's Input, Output, CatalystActivity and RegulatedBy attributes.
-				// Failure to successfully infer any of these attributes will end inference for this reaction.
-				logger.info("Inferring inputs...");
-				if (inferReactionInputsOrOutputs(reactionInst, infReactionInst, input))
+				logger.info("Inferring outputs...");
+				if (inferReactionInputsOrOutputs(reactionInst, infReactionInst, output))
```
jweiser: It may be clearer to use guard clauses rather than nest. 

https://stackoverflow.com/a/4887280/2295778
https://refactoring.guru/replace-nested-conditional-with-guard-clauses

```
// Attempt to infer all PhysicalEntities associated with this reaction's Input, Output, CatalystActivity and RegulatedBy attributes.
// Failure to successfully infer any of these attributes will end inference for this reaction.
logger.info("Inferring inputs...");
if (!(inferReactionInputsOrOutputs(reactionInst, infReactionInst, input)))
{
	logger.info("Input inference unsuccessful -- terminating inference for " + reactionInst);
	return;
}

logger.info("Inferring outputs...");
if (!(inferReactionInputsOrOutputs(reactionInst, infReactionInst, output)))
{
	logger.info("Output inference unsuccessful -- terminating inference for " + reactionInst);
	return;
}

logger.info("Inferring catalysts...");
if (!(inferReactionCatalysts(reactionInst, infReactionInst)))
{
	logger.info("Catalyst inference unsuccessful -- terminating inference for " + reactionInst);
	return;
}
```

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -52,87 +51,72 @@ public static void inferReaction(GKInstance reactionInst) throws Exception
...
-							String inferredEvent = infReactionInst.getAttributeValue(DB_ID).toString() + "\t" + infReactionInst.getDisplayName() + "\n";	
-							Files.write(Paths.get(inferredFilehandle), inferredEvent.getBytes(), StandardOpenOption.APPEND);
-						} else {
-							logger.info("Catalyst inference unsuccessful -- terminating inference for " + reactionInst);
 						}
+						// After successfully adding a new inferred instance to the DB, it is recorded in the 'inferred reactions' file
+						inferredCount++;
+						inferrableHumanEvents.add(reactionInst);
+						String inferredEvent = infReactionInst.getAttributeValue(DB_ID).toString() + "\t" + infReactionInst.getDisplayName() + "\n";
+						Files.write(Paths.get(inferredFilehandle), inferredEvent.getBytes(), StandardOpenOption.APPEND);
```
jweiser: `inferredFilehandle` might be better named `inferredFilePath`.  It also may be better to store it as a Path object rather than a String:

```
private static Path inferredFilePath;

// Replacing the setInferredFilename method
public static void setInferredFilePath(String inferredFilename)
{
	inferredFilePath = Paths.get(inferredFilename);
}
```

jweiser: If the reaction db_id and display name are needed, the `getExtendedDisplayName()` method could be used in place of the `inferredEvent` temp variable:

```
Files.write(
    Paths.get(inferredFilehandle), 
    infReactionInst.getExtendedDisplayName().getBytes(), 
    StandardOpenOption.APPEND
);
```

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
```diff
@@ -293,20 +272,17 @@ public static void setSummationInstance(GKInstance summationInstCopy)
...
+		inferredEvent.putAll(eventsAlreadyInferredMap);
 		return inferredEvent;
 	}
 	
-	public static List<GKInstance> getInferrableHumanEvents()
+	public static List<GKInstance> getInferrableHumanEvents(List<GKInstance> eventsAlreadyInferred)
 	{
+		inferrableHumanEvents.addAll(eventsAlreadyInferred);
 		return inferrableHumanEvents;
 	}
```
jweiser: See https://github.com/reactome/data-release-pipeline/pull/104/files#r339807277

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -219,4 +327,16 @@ public static void setAdaptor(MySQLAdaptor dbAdaptor)
 	{
 		dba = dbAdaptor;
 	}
+
+	public static void setHomologueMappingFile(Map<String, String[]> homologueMappingsCopy) { homologueMappings = homologueMappingsCopy; }
+
+	public static void setEligibleFilename(String eligibleFilename)
+	{
+		eligibleFilehandle = eligibleFilename;
+	}
```
jweiser: Similar to https://github.com/reactome/data-release-pipeline/pull/104/files#r340082898

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+	 * 	Each input, output, and catalyst is screened to verify that the Reaction will be inferred.
+	 * 	This prevents the majority of orphan PEs that were being created during orthoinference.
+	 * 	Only some EntitySets are still orphaned due to the complexity behind attempting to screen them
+	 * @param reactionInst -- GKInstance that will be screened
+	 * @return -- Boolean is returned that indicates if Reaction is fully inferrable or not
+	 * @throws Exception
+	 */
+	private static boolean reactionComponentsAreInferrable(GKInstance reactionInst) throws Exception {
+		// First gather all inputs, outputs and the PEs in catalyst activities
+		// Inputs/Outputs/CatalystPEs need to be stored in seperate collections. At time of writing, having it all stored in
```
jweiser: Type - separate?

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+		// This function finds the total number of distinct proteins associated with an instance, as well as the number that can be inferred.
+		// Total proteins are stored in reactionProteinCounts[0], inferrable proteins in [1], and the maximum number of homologues for any entity involved in index [2].
+		// Reactions with no proteins/EWAS (Total = 0) are not inferred.
+		List<Integer> reactionProteinCounts = ProteinCountUtility.getDistinctProteinCounts(reactionInst);
+		int reactionTotalProteinCounts = reactionProteinCounts.get(0);
+		if (reactionTotalProteinCounts > 0) {
+			logger.info("Total protein count for RlE: " + reactionTotalProteinCounts);
+			eligibleCount++;
+			String eligibleEventName = reactionInst.getAttributeValue(DB_ID).toString() + "\t" + reactionInst.getDisplayName() + "\n";
+			Files.write(Paths.get(eligibleFilehandle), eligibleEventName.getBytes(), StandardOpenOption.APPEND);
```
jweiser: Similar to https://github.com/reactome/data-release-pipeline/pull/104/files#r340084682

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+				return false;
+			}
+		}
+		// Screen catalyst PhysicalEntities
+		for (GKInstance reactionCatalystPE : reactionCatalystPEs) {
+			if (!componentIsInferrable(reactionCatalystPE)) {
+				return false;
+			}
+		}
+		return true;
```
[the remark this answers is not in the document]
jweiser: The only thing is the `componentIsInferrable` exception would need to be handled in that method rather than thrown.  If it's practical to handle the exception in the method it gets thrown rather than throwing it to the method caller it can help reduce the methods which throw exceptions.

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+	}
+
+	// This looks a lot like the code structure found in OrthologousEntityGenerator, just without the actual inference code or override functionality
+	// This will screen all instance types except for some EntitySets, which are complex to screen ahead of time.
+	private static boolean componentIsInferrable(GKInstance reactionComponent) throws Exception {
+		// This block doesn't do anything aside from prevent non-species-containing instances from going through other screening.
+		// During actual inference, a non-species-containing instance would be returned without any inference.
+		if (!SpeciesCheckUtility.checkForSpeciesAttribute(reactionComponent)) {
+//				return true;
```
jweiser: This could be a guard clause and still have the `return true;` commented in.  Empty blocks are best avoided.

orthoinference/src/main/java/org/reactome/orthoinference/ReactionInferrer.java
SolomonShorser-OICR: Yeah, this could confuse future developers, as Java does not really use file handles.

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+				return false;
+			}
+		} else if (reactionComponent.getSchemClass().isa(Complex) || reactionComponent.getSchemClass().isa(Polymer) || reactionComponent.getSchemClass().isa(EntitySet)) {
+			List<Integer> complexProteinCounts = ProteinCountUtility.getDistinctProteinCounts(reactionComponent);
+			int totalProteinCounts = complexProteinCounts.get(0);
+			int inferrableProteinCounts = complexProteinCounts.get(1);
+			if (reactionComponent.getSchemClass().isa(Complex) || reactionComponent.getSchemClass().isa(Polymer)) {
+				int percent = 0;
+				if (totalProteinCounts > 0) {
+					percent = (inferrableProteinCounts * 100) / totalProteinCounts;
```
SolomonShorser-OICR: Using an `int` for `percent` doesn't cause any loss-of-precision problems?

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
```diff
@@ -101,9 +107,111 @@ public static boolean checkIfInstanceShouldBeSkipped(GKInstance reactionInst) th
...
+	// This looks a lot like the code structure found in OrthologousEntityGenerator, just without the actual inference code or override functionality
+	// This will screen all instance types except for some EntitySets, which are complex to screen ahead of time.
+	private static boolean componentIsInferrable(GKInstance reactionComponent) throws Exception {
+		// This block doesn't do anything aside from prevent non-species-containing instances from going through other screening.
+		// During actual inference, a non-species-containing instance would be returned without any inference.
+		if (!SpeciesCheckUtility.checkForSpeciesAttribute(reactionComponent)) {
+//				return true;
+		} else if (reactionComponent.getSchemClass().isa(GenomeEncodedEntity))
+		{
+			if (reactionComponent.getSchemClass().toString().contains(EntityWithAccessionedSequence)) {
```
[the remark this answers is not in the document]
SolomonShorser-OICR: `isa` probably wouldn't make sense since this is comparing Strings. It's possible that the String-representation of the SchemaClass is something like `[EntityWithAccessionedSequence]` so `equals` would fail as well. Not sure...

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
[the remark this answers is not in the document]
jweiser: Yes, `allMatch` is a short-circuiting operation: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#allMatch-java.util.function.Predicate-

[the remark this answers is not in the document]
SolomonShorser-OICR: One way to simplify this might be to add all of the inputs, outputs, PEs into a big list and then just use one loop:
```
List<GKInstance> bigList = new ArrayList<>();
bigList.addAll(reactionInputs);
bigList.addAll(reactionCatalystPEs);
bigList.addAll(reactionOutputs);

for (GKInstance instance : bigList) {
    if (!componentIsInferrable(instance)) {
        return false;
    }
}
return true;
```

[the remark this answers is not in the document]
SolomonShorser-OICR: This still has `return` in the middle of the function, but only in two places, and the code is smaller so it's not as bad ;)

orthoinference/src/main/java/org/reactome/orthoinference/SkipInstanceChecker.java
jweiser: `double percent = 0` at line 199 would be good

PR #110, opened 2019-11-06
Feature/ncbi uniprot isoforms

jweiser:
This PR updates data-exporter to use UniProt isoform accessions (e.g. P12345-1) where they exist instead of the UniProt parent/canonical accession (e.g. P12345).  

This is to remove 404 results for exported Reactome content search links resulting from using the parent accession when the isoform accession was the participant in Reactome.


data-exporter/src/main/java/org/reactome/release/dataexport/NCBIEntry.java
```diff
@@ -68,8 +68,9 @@ public String getUniprotDisplayName() {
 				"MATCH (rgp:ReferenceGeneProduct)-[:referenceDatabase]->(rd:ReferenceDatabase)",
 				"MATCH (rgp)-[:referenceGene]->(rds:ReferenceDNASequence)",
 				"WHERE rd.displayName = 'UniProt' AND rds.databaseName = 'NCBI Gene'",
-				"RETURN rgp.dbId, rgp.displayName, rgp.identifier, rds.identifier",
-				"ORDER BY rgp.identifier;"
+				"RETURN rgp.dbId, rgp.displayName, coalesce(rgp.variantIdentifier, rgp.identifier) as rgp_accession,"
```
cookersjs: Based on what I found online, the coalesce method will return the first shared value between the submitted expression list? https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-coalesce

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -19,12 +20,18 @@
...
+		"([OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2})"
+	); // As defined at https://www.uniprot.org/help/accession_numbers (October 2019)
+
 	private static Map<Long, UniProtReactomeEntry> uniProtReactomeEntryMap = new HashMap<>();
 
-	private static Map<Session, Map<String, Set<ReactomeEvent>>> uniprotAccessionToTopLevelPathwaysCache =
-		new HashMap<>();
-	private static Map<Session, Map<String, Set<ReactomeEvent>>> uniprotAccessionToReactomeEventCache = new HashMap<>();
-	private static Map<Session, Map<String, Set<Long>>> uniprotAccessionToReactionLikeEventIdCache = new HashMap<>();
+	private static Map<Session, Map<UniProtReactomeEntry, Set<ReactomeEvent>>>
```
cookersjs: I might just not have got far enough in the PR, but where does UniProtReactomeEntry and ReactomeEvent data types come from? Are they a custom thing? What do they look like?

data-exporter/src/main/java/org/reactome/release/dataexport/NCBIEntry.java
jweiser: Yup, `coalesce` is used here to take the "variant identifier" value, which is the UniProt isoform accession, if it exists, but to use the "identifier" which is the parent/canonical UniProt accession otherwise for non-isoforms where "variant identifier" will be null.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -127,104 +160,140 @@ private UniProtReactomeEntry(long dbId, String accession, String displayName) {
...
+		if (uniprotReactomeEntryToReactomeEventCache.containsKey(graphDBSession)) {
+			return uniprotReactomeEntryToReactomeEventCache.get(graphDBSession);
 		}
 
 		logger.info("Computing UniProt to Reactome events");
 
 		Map<Long, ReactomeEvent> eventCache = ReactomeEvent.fetchReactomeEventMap(graphDBSession);
-		Map<String, Set<Long>> uniprotAccessionToReactionLikeEventId = fetchUniProtAccessionToRLEId(graphDBSession);
+		Map<UniProtReactomeEntry, Set<Long>> uniprotReactomeEntryToReactionLikeEventId =
+			fetchUniProtReactomeEntryToRLEId(graphDBSession);
```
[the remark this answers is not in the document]
jweiser: Oh, the competition is on now!

I always preferred "like" being capitalized for this class, so I am rebelling :).

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -127,104 +160,140 @@ private UniProtReactomeEntry(long dbId, String accession, String displayName) {
...
 
-		Map<String, Set<Long>> uniprotAccessionToReactionLikeEventId = new HashMap<>();
+		Map<UniProtReactomeEntry, Set<Long>> uniprotReactomeEntryToReactionLikeEventId = new HashMap<>();
 		while (statementResult.hasNext()) {
 			Record record = statementResult.next();
 
-			String uniprotAccession = record.get("rgp.identifier").asString();
+			long uniprotDbId = record.get("rgp.dbId").asLong();
+			String uniprotAccession = record.get("rgp_accession").asString();
+			String uniprotDisplayName = record.get("rgp.displayName").asString();
```
cookersjs: I've seen these strings ("rgp.displayName") appear in multiple places, and I also noticed for this PR that they were changed from an older format to a newer one. Since they seem to be mutable and also used in multiple places, should they be considered for a Constants class?

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -267,14 +336,15 @@ private void setDbId(long dbId) {
 	 * a 6 or 10 character String)
 	 */
 	private void setAccession(String accession) {
-		final List<Integer> ACCEPTED_UNIPROT_ACCESSION_LENGTHS = Arrays.asList(6, 10);
 		if (accession == null) {
 			throw new NullPointerException("UniProt Accession is null");
 		}
 
-		if (!ACCEPTED_UNIPROT_ACCESSION_LENGTHS.contains(accession.length())) {
+		if (!isValidCanonicalAccession(accession) && !isValidIsoformAccession(accession)) {
```
cookersjs: If this was ever changed to use ReactomeJavaConstants as an imported static class, this 'access' variable will become problematic. Just some food for thought!

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -303,6 +373,37 @@ private void setDisplayName(String displayName) {
...
+	 * Checks if a passed String is a valid canonical UniProt accession (i.e. 6 or 10 alphanumeric characters matching
+	 * the regular expression "[OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2}" as defined at
+	 * https://www.uniprot.org/help/accession_numbers as of October 2019).
+	 * E.g. P12345
+	 * @param accession UniProt accession to test for validity
+	 * @return true if the accession passed is a valid canonical (i.e. not an isoform) UniProt accession,
+	 * false otherwise
+	 */
+	private boolean isValidCanonicalAccession(String accession) {
+		final Pattern ANCHORED_PARENT_ACCESSION_REGEX = Pattern.compile("^" + PARENT_ACCESSION_REGEX.pattern() + "$");
```
cookersjs: I am confused about why it is `final`, and why it is placed here in this function instead of at the top of the class?

data-exporter/src/test/java/org/reactome/release/dataexport/NCBIEntryTest.java
```diff
@@ -64,7 +64,7 @@ public void retrievesNCBIEntryFromDummyGraphDB() {
 
 		List<NCBIEntry> ncbiEntries = NCBIEntry.getUniProtToNCBIGeneEntries(dummyGraphDBServer.getSession());
 
-		assertThat(ncbiEntries, contains(expectedNCBIEntry));
+		assertThat(ncbiEntries, hasItem(expectedNCBIEntry));
```
cookersjs: I believe that `contains` is the more stringent matcher. What was the reason for switching to `hasItem`?

data-exporter/src/test/java/org/reactome/release/dataexport/ReactomeEventTest.java
```diff
@@ -34,7 +33,7 @@ public void retrievesNCBIEntryFromDummyGraphDB() {
 		Map<Long, ReactomeEvent> reactomeEventCache =
 			ReactomeEvent.fetchReactomeEventMap(dummyGraphDBServer.getSession());
 
-		assertThat(reactomeEventCache, is(aMapWithSize(7)));
+		assertThat(reactomeEventCache, is(aMapWithSize(15)));
```
cookersjs: I guess this is one of the magic numbers you'd want to replace with a properly named variable?

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtDbIdGenerator.java
```diff
@@ -0,0 +1,9 @@
+package org.reactome.release.dataexport;
+
+public class UniProtDbIdGenerator {
+	private static long uniProtDbId = 1L;
+
+	public static long getNextUniProtDBID() {
```
cookersjs: I'm a little confused by what is going on here

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -127,104 +160,140 @@ private UniProtReactomeEntry(long dbId, String accession, String displayName) {
...
+		uniprotReactomeEntryToReactionLikeEventId.keySet().stream().forEach(uniprotAccession -> {
 			Set<Long> reactionLikeEventIds =
-				uniprotAccessionToReactionLikeEventId.computeIfAbsent(uniprotAccession, k -> new HashSet<>());
+				uniprotReactomeEntryToReactionLikeEventId.computeIfAbsent(uniprotAccession, k -> new HashSet<>());
 
 			// Add RLEs containing UniProt accession
-			uniprotAccessionToReactomeEvent
+			uniprotReactomeEntryToReactomeEvent
 				.computeIfAbsent(uniprotAccession, k -> new HashSet<>())
 				.addAll(reactionLikeEventIds.stream().map(eventCache::get).collect(Collectors.toSet()));
```
SolomonShorser-OICR: This expression (and others like it) is rather dense, could probably use a comment to explain what is going on here.

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtReactomeEntryTest.java
```diff
@@ -43,33 +54,63 @@ public void differentUniProtObjectsWithDifferentValuesNotEqual() {
...
+		UniProtReactomeEntry uniprot = UniProtReactomeEntry.get(
+			getNextUniProtDBID(), ISOFORM_ACCESSION, DUMMY_UNIPROT_DISPLAY_NAME
+		);
+
+		assertThat(uniprot.getAccession(), equalTo(ISOFORM_ACCESSION));
+	}
+
+	@Test
+	public void incorrectIsoformAccessionThrowsIllegalArgumentException() {
+		final String INCORRECT_ISOFORM_ACCESSION = "P123456-1"; // 7 character parent accession is illegal in UniProt
```
cookersjs: Semantics: Would it be better described as Invalid rather than Incorrect?

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: I _think_ the change from "rgp." to "rgp_" is because of changes to queries, such as this `coalesce(rgp.variantIdentifier, rgp.identifier) as rgp_accession`. Here, `AS` is used to alias a more complex expression to a new name, `rgp_accession` in this case. It should be noted that `rgp_accession` is NOT a name found in any structure within the database (at least I don't think it is). 

I suppose if the same aliases are repeated throughout several queries, creating a set of String constants _might_ be useful. But not necessary.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: @cookersjs Huh?

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: Some developers declare everything final, except for what they _know_ will need to be mutable. It's not necessary, it's just a habit.

data-exporter/src/main/java/org/reactome/release/dataexport/UCSC.java
```diff
@@ -209,14 +209,14 @@ private String getNoEventsErrorMessage(UniProtReactomeEntry uniProtReactomeEntry
 				"-[:referenceDatabase]->(rd:ReferenceDatabase)",
 				"WHERE ewas.speciesName IN ['Homo sapiens', 'Rattus norvegicus', 'Mus musculus'] AND rd.displayName =" +
```
SolomonShorser-OICR: Why only these three species? Should this list be configurable?

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
cookersjs: I'm curious about why you decided to implement custom objects? As always, my focus is on the future developer who will come to Reactome to look at the codebase. I do think this could be a clarifying feature, but I am just curious about the motivation

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -127,104 +160,140 @@ private UniProtReactomeEntry(long dbId, String accession, String displayName) {
...
+	private static Map<UniProtReactomeEntry, Set<Long>> fetchUniProtReactomeEntryToRLEId(Session graphDBSession) {
+		if (uniprotReactomeEntryToReactionLikeEventIdCache.containsKey(graphDBSession)) {
+			return uniprotReactomeEntryToReactionLikeEventIdCache.get(graphDBSession);
 		}
 
 		logger.info("Computing UniProt to RLE id");
 
 		StatementResult statementResult = graphDBSession.run(
 			String.join(System.lineSeparator(),
 				"MATCH (rgp:ReferenceGeneProduct)<-[:referenceEntity|:referenceSequence|:hasModifiedResidue]-" +
```
SolomonShorser-OICR: This query seems a bit complex. Should it have a brief comment above it to clarify what it's doing?

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: ...although I agree that in a two line function, declaring the _only_ variable as `final` seems unnecessary. But it doesn't do any harm.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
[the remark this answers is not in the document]
cookersjs: No one can challenge the Data Model's authority!

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
cookersjs: In orthoinference, we have throughout the code imported ReactomeJavaConstants statically, so that things like `ReactomeJavaConstants.Pathway` can actually be written as `Pathway`. Joel has been an advocate for that in the past (though not recently), and so if anyone was to ever implement that in this code, the variable `accession` might cause complications

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -303,6 +373,37 @@ private void setDisplayName(String displayName) {
...
+	/**
+	 * Checks if a passed String is a valid isoform UniProt accession (i.e. 6 or 10 alphanumeric characters matching
+	 * the regular expression "[OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9]([A-Z][A-Z0-9]{2}[0-9]){1,2}" as defined at
+	 * https://www.uniprot.org/help/accession_numbers as of October 2019, and then followed by a dash and by any number
+	 * of digits).  E.g. P12345-1
+	 * @param accession UniProt accession to test for validity
+	 * @return true if the accession passed is a valid UniProt accession for an isoform, false otherwise
+	 */
+	private boolean isValidIsoformAccession(String accession) {
+		Pattern isoformSuffixRegex = Pattern.compile("-[0-9]+");
```
SolomonShorser-OICR: There's a number of `Pattern`s that are used in this class that do not vary. Creating a new pattern that is always the same every time a function is called is not very efficient. If these patterns do not vary then they could be created as `static final` members of the enclosing class.

data-exporter/src/test/java/org/reactome/release/dataexport/EuropePMCTest.java
```diff
@@ -16,6 +16,94 @@
...
+		);
+	}
+
+	@Test
+	public void properEuropePMCProfileXML() {
+		final String REACTOME_DESCRIPTION =
+			"Reactome is a free, open-source, curated and peer-reviewed pathway database.";
+
+		final String EXPECTED_EUROPE_PMC_PROFILE_XML = String.join(System.lineSeparator(),
+			"<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>",
```
SolomonShorser-OICR: Is it necessary to have XML inline in the test method? Could it be moved to a separate file?

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtReactomeEntryTest.java
SolomonShorser-OICR: Not sure: http://sparrowflights.blogspot.com/2014/03/incorrect-vs-invalid-vs-illegal.html

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: I needed classes to hold and manipulate data for Reactome events (pathways and reaction like events) and Reference Gene Products for UniProt instances.  The Java API and data model has some of that functionality but it made sense to create these two classes to hold the specific data and methods needed for the project.  I may re-organize the classes in this project into more specific subpackages going forward so the project structure is a bit clearer.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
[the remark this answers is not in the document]
jweiser: I'm trying to free your mind, Justin.  I can only open the door - you're the one who has to walk through it...

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: Yeah, I was thinking of consolidating repeated parts of the queries through the PR.  I'll look into it.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: Fair point - if ReactomeJavaConstants were imported statically, they could cause confusion with a local method variable shadowing it.  I can make the variable name more specific like `uniProtAccession` to avoid that.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: I usually prefer to make my constants final even if it's scoped to a method.  @SolomonShorser-OICR is right - it's not necessary but just an indication I use to show it's meant to give a name to something constant and cryptic/magical.

data-exporter/src/test/java/org/reactome/release/dataexport/NCBIEntryTest.java
jweiser: That's true - `contains` is more stringent but it requires the `expectedNCBIEntry` to be the only entry in `ncbiEntries`.  Another entry was added in the mock data to include the case of a UniProt isoform, so it was necessary to change to `hasItem` so the test would succeed with only one of the two entries being checked for.

data-exporter/src/test/java/org/reactome/release/dataexport/ReactomeEventTest.java
jweiser: Yes, this was the one I was thinking about yesterday.  I will replace it with a named variable!

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtDbIdGenerator.java
jweiser: Before I was using hardcoded numbers to give mock db ids in the tests.  Since the UniProtReactomeEntry objects that get created are cached (so duplicates aren't created), this caused a headache in making sure the numbers didn't collide when running multiple tests.  I created this as a helper class to generate mock db ids starting at 1 and to give the next available number.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: That's fair - I'll work on making it more readable.

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtReactomeEntryTest.java
jweiser: Looking at the definitions, maybe "Illegal" is the best of the three since it violates UniProt's accession rules.

data-exporter/src/main/java/org/reactome/release/dataexport/UCSC.java
jweiser: I'm not sure -- this was the way it was in the Perl script which predates my starting with Reactome.  I can e-mail Peter and Robin about it to ask.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: Sure - I'll make it clearer.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
jweiser: I'll move them to `static final` members.

data-exporter/src/test/java/org/reactome/release/dataexport/EuropePMCTest.java
jweiser: I'm kind of inclined to have it remain in the test file just so the test is self-contained and readable at a glance without needing to look up the contents of another file.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: tbh, I wouldn't worry too much about that.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
SolomonShorser-OICR: Which is why I'm not a fan of static imports like that. ;)

data-exporter/src/main/java/org/reactome/release/dataexport/UCSC.java
SolomonShorser-OICR: It might be that UCSC only handles these three, in which case a comment stating that should be enough. Or a constant string named "UCSC_ACCEPTED_SPECIES"... either should be fine.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
cookersjs: Yea while the static imports did make the code look cleaner, I think that they did make the code a bit more esoteric. Maybe we should discuss this a bit so that we can decide if we want to have them in the release code? Right now, we've Frankensteined the codebase a bit with Orthoinference having static imports but other modules not

jweiser: It's aliiivve!

Yeah, I think the static importing suggestion I made for those variables was misguided.  Using the `ReactomeJavaConstants` prefix may be the better way to go.

data-exporter/src/test/java/org/reactome/release/dataexport/UniProtDbIdGenerator.java
cookersjs: Fair enough, could a brief comment describing the methods purpose be added?

jweiser: Yup, will do!

data-exporter/src/main/java/org/reactome/release/dataexport/UCSC.java
jweiser: The named constant sounds good.  I'll do that and verify the accepted species with Peter and Robin as well.

data-exporter/src/main/java/org/reactome/release/dataexport/UniProtReactomeEntry.java
```diff
@@ -205,46 +208,143 @@ private UniProtReactomeEntry(long dbId, String accession, String displayName) {
...
+	) {
+		Set<ReactomeEvent> pathwayReactomeEvents = new HashSet<>();
+
+		pathwayReactomeEvents.addAll(reactionLikeEventIds.stream().flatMap(
+			rleId -> {
+				Map<Long, Set<Long>> rleIdToPathwayIds = fetchRLEIdToPathwayId(graphDBSession);
+				Set<Long> pathwayIds = rleIdToPathwayIds.computeIfAbsent(rleId, k -> new HashSet<>());
+				return pathwayIds.stream();
+			}
+		).map(pathwayId -> convertDbIdToReactomeEvent(pathwayId, graphDBSession)).collect(Collectors.toList()));
```
SolomonShorser-OICR: So, you map Event IDs to... Pathway IDs. Then you map those to... Events? Could this get a little more documentation?

Also: why two separate calls to flatMap/map? Why not just have one mapping function?

jweiser: Okay, I reworked it so it should be clearer now.

PR #113, opened 2019-11-11
Feature/post step test other identifiers

cookersjs:
This is actually just a PR for the other identifiers work. Stable Id History post-step code will be in another PR.

Simple post-step check for Other Identifiers. For each species in the database, count the number of ReferenceGeneProduct instances that have their 'otherIdentifier' slot filled. Compare this count with the same count done on the previous database and report if there is less.

Please suggest other unit tests if you can think of any.

### DIFF
--- a/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
+++ b/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
@@ -1,0 +1,41 @@
+package org.reactome.release.otherIdentifiers;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.gk.persistence.MySQLAdaptor;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.sql.SQLException;
+import java.util.Properties;
+
+public class Main {
+
+    private static final Logger logger = LogManager.getLogger();
+    private static final String RESOURCES_DIR = Paths.get("src", "main", "resources").toString();
+
+    public static void main(String[] args) throws IOException, SQLException {
+
+        String pathToConfig = args.length > 0 ? args[0] : Paths.get(RESOURCES_DIR ,"config.properties").toString();
+
+        Properties props = new Properties();
+        props.load(new FileInputStream(pathToConfig));
+
+        //Set up DB adaptor
+        String username = props.getProperty("release.database.user");
+        String password = props.getProperty("release.database.password");
+        String host = props.getProperty("release.database.host");
+        int port = Integer.valueOf(props.getProperty("release.database.port"));
+
+        String releaseCurrent = props.getProperty("release_current.name");
+        String releasePrevious = props.getProperty("release_previous.name");
+
+        MySQLAdaptor dbAdaptor = new MySQLAdaptor(host, releaseCurrent, username, password, port);
+        MySQLAdaptor dbAdaptorPrev = new MySQLAdaptor(host, releasePrevious, username, password, port);
+
+        logger.info("Executing post-step checks");
+        PostStepChecks.compareOtherIdentifierCounts();
+
+    }
+}
--- a/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
+++ b/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
@@ -15,7 +15,7 @@ public class Main {
     private static final Logger logger = LogManager.getLogger();
     private static final String RESOURCES_DIR = Paths.get("src", "main", "resources").toString();
 
-    public static void main(String[] args) throws IOException, SQLException {
+    public static void main(String[] args) throws Exception {
 
         String pathToConfig = args.length > 0 ? args[0] : Paths.get(RESOURCES_DIR ,"config.properties").toString();
 
@@ -35,7 +35,7 @@ public class Main {
         MySQLAdaptor dbAdaptorPrev = new MySQLAdaptor(host, releasePrevious, username, password, port);
 
         logger.info("Executing post-step checks");
-        PostStepChecks.compareOtherIdentifierCounts();
+        PostStepChecks.compareOtherIdentifierCounts(dbAdaptor, dbAdaptorPrev);
 
     }
 }
--- a/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
+++ b/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
@@ -5,9 +5,7 @@ import org.apache.logging.log4j.Logger;
 import org.gk.persistence.MySQLAdaptor;
 
 import java.io.FileInputStream;
-import java.io.IOException;
 import java.nio.file.Paths;
-import java.sql.SQLException;
 import java.util.Properties;
 
 public class Main {
--- a/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
+++ b/other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
@@ -9,6 +9,13 @@ import org.gk.persistence.MySQLAdaptor;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ *  This is a QA step that will be run after finishing an OtherIdentifiers run. For each species,
+ *  the count of ReferenceGeneProduct instances that have their 'otherIdentifier' attribute filled is found
+ *  between the current and previous database. If the previous database has a higher count, a warning message
+ *  is output and should be looked into. If the count is still relatively close,  it can probably be ignored.
+ */
+
 public class PostStepChecks {
     private static final Logger logger = LogManager.getLogger();
 
@@ -17,20 +24,36 @@ public class PostStepChecks {
         Collection<GKInstance> speciesInstances = dba.fetchInstancesByClass(ReactomeJavaConstants.Species);
 
         for (GKInstance speciesInstance : speciesInstances) {
-            int currentOtherIdentifierCount = getOtherIdentifierCount(dba, speciesInstance);
-            int previousOtherIdentifierCount = getOtherIdentifierCount(dbaPrev, speciesInstance);
+            int currentOtherIdentifierCount = getRGPInstanceOtherIdentifierCount(dba, speciesInstance);
+            int previousOtherIdentifierCount = getRGPInstanceOtherIdentifierCount(dbaPrev, speciesInstance);
             if (currentOtherIdentifierCount < previousOtherIdentifierCount) {
                 logger.warn(speciesInstance.getDisplayName() + " has fewer ReferenceGeneProduct instances with OtherIdentifiers compared to previous release:  " +
                         "release_current - " + currentOtherIdentifierCount + " -- release_previous - " + previousOtherIdentifierCount);
             }
         }
-
-
     }
 
-    private static int getOtherIdentifierCount(MySQLAdaptor dba, GKInstance speciesInstance) throws Exception {
-        AtomicInteger currentOtherIdentifierCount = new AtomicInteger();
+    /**
+     * Wrapper method for determining other identifier count.
+     * @param dba MySQLAdaptor
+     * @param speciesInstance GKInstance pertaining to a specific species
+     * @return The other identifier count, as an int.
+     * @throws Exception
+     */
+    private static int getRGPInstanceOtherIdentifierCount(MySQLAdaptor dba, GKInstance speciesInstance) throws Exception {
         Collection<GKInstance> rgpInstances = dba.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceGeneProduct, ReactomeJavaConstants.species, "=", speciesInstance);
+
+        int currentOtherIdentifierCount = getOtherIdentifierCount(rgpInstances);
+        return currentOtherIdentifierCount;
+    }
+
+    /**
+     * Determines the number of ReferenceGeneProduct instances that have a filled otherIdentifier attribute.
+     * @param rgpInstances Collection of RGP GKInstances associated with a species.
+     * @return An int is returned, representing the otherIdentifier count.
+     */
+    public static int getOtherIdentifierCount(Collection<GKInstance> rgpInstances) {
+        AtomicInteger currentOtherIdentifierCount = new AtomicInteger();
         rgpInstances.forEach(rgpInstance -> {
             try {
                 if (rgpInstance.getAttributeValuesList(ReactomeJavaConstants.otherIdentifier).size() > 0) {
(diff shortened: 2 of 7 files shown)


other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+
+    /**
+     * Wrapper method for determining other identifier count.
+     * @param dba MySQLAdaptor
+     * @param speciesInstance GKInstance pertaining to a specific species
+     * @return The other identifier count, as an int.
+     * @throws Exception
```
jweiser: If possible, it would be good to include what causes an exception to be thrown.  Usually, I just temporarily remove the `throws Exception` part of the method signature and see what part of the code throws the exception and document that a failure of that logic to complete successfully is what causes it.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    private static int getRGPInstanceOtherIdentifierCount(MySQLAdaptor dba, GKInstance speciesInstance) throws Exception {
+        Collection<GKInstance> rgpInstances = dba.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceGeneProduct, ReactomeJavaConstants.species, "=", speciesInstance);
+
+        int currentOtherIdentifierCount = getOtherIdentifierCount(rgpInstances);
+        return currentOtherIdentifierCount;
```
jweiser: No need for a temporary variable here - you can return the result of `getOtherIdentifierCount(rgpInstances)` directly.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Determines the number of ReferenceGeneProduct instances that have a filled otherIdentifier attribute.
+     * @param rgpInstances Collection of RGP GKInstances associated with a species.
+     * @return An int is returned, representing the otherIdentifier count.
+     */
+    public static int getOtherIdentifierCount(Collection<GKInstance> rgpInstances) {
```
jweiser: The name `getOtherIdentifierCount` kind of sounds like a total count of all other identifiers in the passed collection of `rgpInstances`. Maybe its more precise to name the method something like `getInstanceWithOtherIdentifiersCount`?

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -0,0 +1,68 @@
...
+        AtomicInteger currentOtherIdentifierCount = new AtomicInteger();
+        rgpInstances.forEach(rgpInstance -> {
+            try {
+                if (rgpInstance.getAttributeValuesList(ReactomeJavaConstants.otherIdentifier).size() > 0) {
```
jweiser: You may need to check that the value you get back is not null before calling `.size()` to ensure there is no NullPointerException being thrown.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -0,0 +1,68 @@
...
+                    currentOtherIdentifierCount.getAndIncrement();
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        });
```
jweiser: In this case, a traditional `for` loop might be better than the `forEach` method called on the `rgpInstances` collection.  That way you can use an `int` primitive in place of an AtomicInteger and the Exception can be part of the method signature instead of handled with a try/catch in the lambda (it drives me crazy that you can't throw the exception inside the lambda to the method containing the lambda!).

Another possibility would be to do something like this:

```
return rgpInstances.stream().filter(rgp -> hasOtherIdentifiers(rgp))).count();
```
and then just create the `hasOtherIdentifiers` method.

```
private static boolean hasOtherIdentifiers(GKInstance rgpInstance) {
    try {
        List<String> otherIdentifiers = rgpInstance.getAttributeValuesList(ReactomeJavaConstants.otherIdentifier);

        return otherIdentifiers != null && !otherIdentifiers.isEmpty();
    } catch(Exception e) {
        String errorMessage = "Unable to retrieve other identifiers from RGP instance " + rgpInstance;
        
        logger.fatal(errorMessage, e);
        throw new RuntimeException(errorMessage, e);
    }
}
```

other-identifiers/src/main/resources/config.properties
```diff
@@ -0,0 +1,8 @@
+
+release.database.user=
+release.database.password=
+release.database.host=
+release.database.port=
```
jweiser: Would it make sense to have the expected 3306 port filled in here?

other-identifiers/src/test/java/org/reactome/release/otherIdentifiers/PostStepChecksTest.java
```diff
@@ -0,0 +1,45 @@
...
+        "javax.xml.*", "com.sun.org.apache.xerces.*", "org.xml.sax.*", "com.sun.xml.*", "org.w3c.dom.*", "org.mockito.*"})
+public class PostStepChecksTest {
+
+    private final int correctOtherIdentifierCount = 2;
+
+    @Mock
+    GKInstance mockGKInstance;
+
+    Collection<GKInstance> gkInstanceCollection = new ArrayList<>();
+    List<Object> otherIdentifierCollection = new ArrayList<>();
```
jweiser: Could this be `List<String>`?

other-identifiers/src/test/java/org/reactome/release/otherIdentifiers/PostStepChecksTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void getOtherIdentifierCountReturnsAtomicInt() throws Exception {
+
+        gkInstanceCollection.add(mockGKInstance);
+        gkInstanceCollection.add(mockGKInstance);
+        otherIdentifierCollection.add("Mock otherIdentifier");
+        Mockito.when(mockGKInstance.getAttributeValuesList(ReactomeJavaConstants.otherIdentifier)).thenReturn(otherIdentifierCollection);
+        int otherIdentifierCount = PostStepChecks.getOtherIdentifierCount(gkInstanceCollection);
+        assertThat(otherIdentifierCount, is(equalTo(correctOtherIdentifierCount)));
```
jweiser: It may be better to mock two separate GKInstance objects and give them separate other identifier collections.  You could do the setup in a method with the `@BeforeEach` annotation and then tests to check you get expected results for these kinds of situations:

- Passing a null value
- Passing an empty collection
- Passing a collection with one instance with one other identifier
- Passing a collection with one instance multiple identifiers (to ensure the instance is still only counted once despite having many identifiers)
- Passing a collection with multiple instances with a combination of no, one, or multiple identifiers each to ensure you get what you expect in each case

jweiser:
Will the post step test for the comparison of other identifier numbers broken by species be added in a separate PR?  

Also, a test for the overall number of other identifiers between the current and previous releases would be useful to complement this PR comparing the number of RGPs with other identifiers.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
cookersjs: Ah right, thats cause I split the method up but I didn't give the resulting change any kind of hard look

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
cookersjs: Since its `getAttributeValuesList` it will return an empty array here. I believe it would only return null if 'otherIdentifier' wasn't a valid attribute for the instance type

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static void main(String[] args) throws Exception {
+
+        String pathToConfig = args.length > 0 ? args[0] : Paths.get(RESOURCES_DIR ,"config.properties").toString();
+
+        Properties props = new Properties();
+        props.load(new FileInputStream(pathToConfig));
+
+        //Set up DB adaptor
+        String username = props.getProperty("release.database.user");
```
SolomonShorser-OICR: If these constants ("release.database.user", "release_current.name", etc...) get used across multiple steps (and I think they do), they should probably be moved out to release-common-lib as some sort of COMMON_PROPERTY_KEYS. Not today, perhaps, but one day.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
SolomonShorser-OICR: I hope that would throw an exception to break the application, since that really should not happen.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
cookersjs: I went with `getCountOfInstancesWithOtherIdentifiers`

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/Main.java
cookersjs: Yea thats a nice idea, it will make the configuration files much more manageable as well

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
cookersjs: Your correct, I checked afterwards and if it isn't a valid attribute then an exception is thrown

SolomonShorser-OICR: And it causes the application to terminate? The InvalidAttributeException (I think) doesn't get swallowed/suppressed?

jweiser:
> Will the post step test for the comparison of other identifier numbers broken by species be added in a separate PR?
> 
> Also, a test for the overall number of other identifiers between the current and previous releases would be useful to complement this PR comparing the number of RGPs with other identifiers.

Ah, disregard the first sentence/question - I didn't notice the check was already broken down by species.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
cookersjs: It doesn't cause it to terminate since the error is caught in the try/catch. In this case when I tried it it just caused continuous stacktraces to be output

SolomonShorser-OICR: Oh. I'm thinking that this type of exception _should_ break the application flow.

jweiser: The approach in this comment might help - https://github.com/reactome/data-release-pipeline/pull/113#discussion_r344835986.  If an exception happens, it probably does indicate something unusual we should investigate.

cookersjs: From the perspective of my code, I am explicitly pulling out `ReferenceGeneProduct` instances. Based on this, there isn't any concern about this exception being thrown unless the Data Model changes

jweiser: But I think the code requires the exception to be handled in order to compile?

SolomonShorser-OICR: It does. It gets caught, and the stack trace gets printed.

jweiser: True, but if the exception is being caught in the code anyway and if it were to occur, termination of the program is probably the best way to go.

other-identifiers/src/main/java/org/reactome/release/otherIdentifiers/PostStepChecks.java
```diff
@@ -38,31 +46,37 @@ public static void compareOtherIdentifierCounts(MySQLAdaptor dba, MySQLAdaptor d
      * @param dba MySQLAdaptor
      * @param speciesInstance GKInstance pertaining to a specific species
      * @return The other identifier count, as an int.
-     * @throws Exception
+     * @throws Exception can be caused by errors interacting with the database using the db adaptor.
```
jweiser: Could this be a little bit more specific to say the error would be in attempting to retrieve ReferenceGeneProduct instances for the passed species?

other-identifiers/src/test/java/org/reactome/release/otherIdentifiers/PostStepChecksTest.java
```diff
@@ -2,44 +2,100 @@
...
+    }
+
+    @Test
+    public void hasOtherIdentifiersNullTest() {
+        try {
+            PostStepChecks.hasOtherIdentifiers(null);
+        } catch(Exception e) {
+            e.printStackTrace();
+            assertTrue(e.getMessage().equals("Unable to retrieve other identifiers from RGP instance null"));
+        }
```
jweiser: It might be better to use `assertThrows` from Hamcrest to check an expected exception.  I have an example that I've used in my code here:

https://github.com/reactome/data-release-pipeline/blob/develop/data-exporter/src/test/java/org/reactome/release/dataexport/EuropePMCTest.java#L90

other-identifiers/src/test/java/org/reactome/release/otherIdentifiers/PostStepChecksTest.java
```diff
@@ -2,44 +2,100 @@
...
-        Mockito.when(mockGKInstance.getAttributeValuesList(ReactomeJavaConstants.otherIdentifier)).thenReturn(otherIdentifierCollection);
-        int otherIdentifierCount = PostStepChecks.getOtherIdentifierCount(gkInstanceCollection);
+    @Test
+    public void getCountOfInstancesWithOtherIdentifiersNullTest() {
+        try {
+            long otherIdentifierCount = PostStepChecks.getCountOfInstancesWithOtherIdentifiers(null);
+        } catch (Exception e) {
+            e.printStackTrace();
+            assertTrue(e.toString().contains("NullPointerException"));
+        }
```
jweiser: See https://github.com/reactome/data-release-pipeline/pull/113/files/17543e4a8453e4249537e52f7f9d641d77f2a2ed..1e2e5e834e292248a26352658eab776d42136a11#r344968165

PR #117, opened 2019-11-18
URL fix script

cookersjs:
Fix for the addlinks problems.

Two things:

1) It is in the orthoinference code since a 'datafix' folder already existed here. We should consider moving this folder to somewhere more appropriate
2) Keep in mind this is just a hotfix, so we just need to review the efficacy of the fix

I'll share the files I created for this in Slack

### DIFF
--- a/orthoinference/src/main/java/org/reactome/datafix/KEGGAddLinksMycobacteriumH37RvLinkFix.java
+++ b/orthoinference/src/main/java/org/reactome/datafix/KEGGAddLinksMycobacteriumH37RvLinkFix.java
@@ -12,6 +12,13 @@ import java.io.FileReader;
 import java.util.*;
 
 public class KEGGAddLinksMycobacteriumH37RvLinkFix {
+
+    // In release 71, some of the links for the 'KEGG Gene M. tuberculosis H37Rv' ReferenceDatabase were reported as broken during
+    // the link-check phase of 'Addlinks-Insertion'. The problem was that historically M. tuberculosis H37Rv identifiers had been
+    // prepended with 'mtv', but now some had the prefix 'mtu'. This was due to the fact that KEGG now has two H37Rv species,
+    // denoted by mtu or mtv. The temporary solution was that the base url for the referenceDatabase was modified to remove the default
+    // 'mtv' string and the appropriate prefix added to the identifier attribute of the DatabaseIdentifier instances.
+
     public static void main(String[] args) throws Exception {
 
         Properties props = new Properties();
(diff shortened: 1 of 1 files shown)


orthoinference/src/main/java/org/reactome/datafix/KEGGAddLinksMycobacteriumH37RvLinkFix.java
```diff
@@ -0,0 +1,103 @@
...
+                identifierToPrefixMap.put(identifier, prefixWithIdentifier);
+            } else {
+                System.out.println("LOOKOUT!");
+            }
+        }
+
+        //TODO: Should display name be updated? Technically the identifier includes the prefix, but from Reactome's perspective we've never included the prefix for any of the KEGG identifiers.
+
+        // Now with the map, I get the reference database of interest. Then I get all referrals of it, which corresponds to the 178 DatabaseIdentifiers.
+        GKInstance keggMycoDBInst = (GKInstance) dba.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceDatabase, ReactomeJavaConstants.name, "=", "KEGG Gene (Mycobacterium tuberculosis H37Rv)").iterator().next();
```
jweiser: There's only one KEGG Gene (Mycobacterium tuberculosis) reference database in release_current, right?

SolomonShorser-OICR:
Originally, 17 links were tested. 12 failed, 5 were ok. Does this process fix ALL identifiers such that all links will work correctly?

orthoinference/src/main/java/org/reactome/datafix/KEGGAddLinksMycobacteriumH37RvLinkFix.java
cookersjs: Yes there is only one for H37Rv, and another for another strain called CDC1551

jweiser: Ah okay and the affected instances are all for H37Rv?

cookersjs:
> Looks good to me as long as the test on a copy of the database worked. Did any of the print statement alerting you to anything unexpected come up during a test run?

@jweiser Nothing came up during the test run after I switched to using just the uniprot file outputs. When I initially parsed the KEGG download file it only had 113/178 identifiers, but uniprot had all 178

cookersjs:
> Originally, 17 links were tested. 12 failed, 5 were ok. Does this process fix ALL identifiers such that all links will work correctly?

@SolomonShorser-OICR Yes the number of identifiers found in the uniprot files is 178, which matches the number of databaseIdentifiers associated with the M. tuberculosis H37Rv reference database instance

orthoinference/src/main/java/org/reactome/datafix/KEGGAddLinksMycobacteriumH37RvLinkFix.java
cookersjs: Yes they are, sorry for belated answer. There was 178 DatabaseIdentifiers that are for H37Rv ReferenceDatabase (determined by solomon) and that was the exact number of referrals to the ReferenceDatabase that were updated

PR #121, opened 2019-12-04
Feature/release common lib refactoring

jweiser:
Refactors classes in release-common-lib package and adds JavaDocs to new methods (includes private methods for making their purpose clear and so it is present if they become public, either in their current class or if extracted to a helper class).

### DIFF
--- a/chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
+++ b/chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
@@ -15,15 +15,13 @@ import org.reactome.util.compare.DBObjectComparer;
 
 public class JavaVsPerlComparison
 {
+	private static final int MAX_COMPARISON_RECURSION_DEPTH = 1;
+	private static final boolean CHECK_REFERRERS = true;
 
 	public static void main(String[] args) throws Exception
 	{
-		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor(
-			"localhost", "perl_gk_central", "root", "root", 3306
-		);
-		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor(
-			"localhost", "java_gk_central", "root", "root", 3306
-		);
+		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "perl_gk_central", "root", "root", 3306);
+		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "java_gk_central", "root", "root", 3306);
 		int diffCount = 0;
 		int sameCount = 0;
 
@@ -74,7 +72,8 @@ public class JavaVsPerlComparison
 			};
 
 			int currentDiff = DBObjectComparer.compareInstances(
-				javaInstance, perlInstance, sb, 1, customAttributeNameFilter, true
+				javaInstance, perlInstance, sb, MAX_COMPARISON_RECURSION_DEPTH, customAttributeNameFilter,
+				CHECK_REFERRERS
 			);
 
 			if (currentDiff > 0)
--- a/chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
+++ b/chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
@@ -18,58 +18,82 @@ public class JavaVsPerlComparison
 
 	public static void main(String[] args) throws Exception
 	{
-		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "test_reactome_66_before_addlinks.dump.sql", "root", "root", 3309);
-		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "gk_central_R66_before_chebi_update.sql", "root", "root", 3308);
+		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor(
+			"localhost", "perl_gk_central", "root", "root", 3306
+		);
+		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor(
+			"localhost", "java_gk_central", "root", "root", 3306
+		);
 		int diffCount = 0;
 		int sameCount = 0;
 
 		@SuppressWarnings("unchecked")
-		String chebiRefDBID = (new ArrayList<GKInstance>( javaUpdatedDB.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceDatabase, ReactomeJavaConstants.name, "=", "ChEBI"))).get(0).getDBID().toString();
+		String chebiRefDBID = (new ArrayList<GKInstance>(
+			javaUpdatedDB.fetchInstanceByAttribute(
+				ReactomeJavaConstants.ReferenceDatabase, ReactomeJavaConstants.name, "=", "ChEBI")
+			)
+		).get(0).getDBID().toString();
 
 		@SuppressWarnings("unchecked")
-		Collection<GKInstance> javaRefMolecules = (Collection<GKInstance>) javaUpdatedDB.fetchInstanceByAttribute(ReactomeJavaConstants.ReferenceMolecule, ReactomeJavaConstants.referenceDatabase, "=", chebiRefDBID);
-
-//		@SuppressWarnings("unchecked")
-//		Collection<GKInstance> perlRefMolecules = (Collection<GKInstance>) perlUpdatedDB.fetchInstanceByAttribute("ReferenceMolecule", "referenceDatabase", "=", chebiRefDBID);
+		Collection<GKInstance> javaRefMolecules = (Collection<GKInstance>) javaUpdatedDB.fetchInstanceByAttribute(
+			ReactomeJavaConstants.ReferenceMolecule, ReactomeJavaConstants.referenceDatabase, "=", chebiRefDBID
+		);
 
 		for (GKInstance javaInstance : javaRefMolecules)
 		{
 			StringBuilder sb = new StringBuilder();
 			GKInstance perlInstance = perlUpdatedDB.fetchInstance(javaInstance.getDBID());
-			
+
 			// This is a filter, to filter instances out of the comparison, based on attributes.
 			//
-			// This filter excludes anything where the attribute is one of: name, displayName, referenceEntity, formula, identifier
-			// AND the object in question is a SimpleEntity and the attribute is referenceEntity.
+			// This filter excludes anything where the attribute is one of: name, displayName,
+			// referenceEntity, formula, identifier AND the object in question is a SimpleEntity and the attribute
+			// is referenceEntity.
 			Predicate<? super SchemaAttribute> customAttributeNameFilter = schemaAttribute -> {
 				// Is the attribute's name in the list?
-				boolean nameIsInList = (new HashSet<String>(Arrays.asList(ReactomeJavaConstants.name, ReactomeJavaConstants._displayName, ReactomeJavaConstants.referenceEntity, ReactomeJavaConstants.formula, ReactomeJavaConstants.identifier))).contains(schemaAttribute.getName());
-				
-				// This condition below is to avoid recursing from the SimpleEntity back the ReferenceMolecule via SimpleEntity's referenceEntity attribute.
-				// The maxRecursionDepth would prevent things from getting too out of hand, but you'll still get a lot of redundant diffs if you don't do this.
+				boolean nameIsInList = (new HashSet<>(
+					Arrays.asList(
+						ReactomeJavaConstants.name, ReactomeJavaConstants._displayName,
+						ReactomeJavaConstants.referenceEntity, ReactomeJavaConstants.formula,
+						ReactomeJavaConstants.identifier
+					)
+				)).contains(schemaAttribute.getName());
+
+				// This condition below is to avoid recursing from the SimpleEntity back the ReferenceMolecule via
+				// SimpleEntity's referenceEntity attribute.  The maxRecursionDepth would prevent things from getting
+				// too out of hand, but you'll still get a lot of redundant diffs if you don't do this.
 				// Is the attribute "referenceEntity" and is it on a SimpleEntity object?
 				@SuppressWarnings("unchecked")
-				boolean simpleEntityWithReferenceEntityAttribute = ( new ArrayList<SchemaClass>(schemaAttribute.getSchemaClass())).get(0).getName().equals(ReactomeJavaConstants.SimpleEntity)
-					&& schemaAttribute.getName().equals(ReactomeJavaConstants.referenceEntity);
-				
+				boolean simpleEntityWithReferenceEntityAttribute =
+					(new ArrayList<SchemaClass>(
+						schemaAttribute.getSchemaClass()
+					)).get(0).getName().equals(ReactomeJavaConstants.SimpleEntity) &&
+					schemaAttribute.getName().equals(ReactomeJavaConstants.referenceEntity);
+
 				return nameIsInList && !simpleEntityWithReferenceEntityAttribute;
 			};
-			
-			int currentDiff = DBObjectComparer.compareInstances(javaInstance, perlInstance, sb, 1, customAttributeNameFilter, true);
-			
+
+			int currentDiff = DBObjectComparer.compareInstances(
+				javaInstance, perlInstance, sb, 1, customAttributeNameFilter, true
+			);
+
 			if (currentDiff > 0)
 			{
-				System.out.println("===\nDiffs on instance: "+javaInstance.toString());
+				System.out.println("===\nDiffs on instance: " + javaInstance.toString());
 				System.out.println(sb.toString());
 			}
 			else
 			{
 				sameCount++;
 			}
-			
+
 			diffCount += currentDiff;
 		}
-		
-		System.out.println("Number of ReferenceMolecule diffs: "+diffCount + "; Number of sames: "+sameCount + "; Total number of ReferenceMolecules compared: "+javaRefMolecules.size());
+
+		System.out.println(
+			"Number of ReferenceMolecule diffs: " + diffCount + "; " +
+			"Number of sames: " + sameCount + "; " +
+			"Total number of ReferenceMolecules compared: " + javaRefMolecules.size()
+		);
 	}
 }
--- a/go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
+++ b/go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
@@ -1,12 +1,17 @@
 package org.reactome.release.goupdate;
 
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -21,34 +26,68 @@ public class JavaVsPerlComparison
 {
 	public static void main(String[] args) throws Exception
 	{
-		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "gk_central_2019_04_18", "root", "root", 3306);//new MySQLAdaptor("localhost", "gk_central_after_uniprot_update.dump.sql", "root", "root", 3308);
-		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "gk_central_2019_04_18", "root", "root", 3307);//new MySQLAdaptor("localhost", "gk_central_R66_before_chebi_update.sql", "root", "root", 3307);
+		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "perl_gk_central", "root", "root", 3306);
+		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "java_gk_central", "root", "root", 3306);
+
+		Map<String, Integer> diffNameToCount = new ConcurrentHashMap<>();
+		Map<String, Integer> sameNameToCount = new ConcurrentHashMap<>();
+
 		AtomicInteger diffCount = new AtomicInteger(0);
 		AtomicInteger sameCount = new AtomicInteger(0);
-		
+
 		AtomicInteger regulationDiffCount = new AtomicInteger(0);
 		AtomicInteger regulationSameCount = new AtomicInteger(0);
-		
+
 		AtomicInteger peDiffCount = new AtomicInteger(0);
 		AtomicInteger peSameCount = new AtomicInteger(0);
-		
+
 		AtomicInteger catActDiffCount = new AtomicInteger(0);
 		AtomicInteger catActSameCount = new AtomicInteger(0);
-		
+
 		AtomicInteger accessionsMissingFromPerl = new AtomicInteger(0);
-		
+
 		@SuppressWarnings("unchecked")
-		Set<GKInstance> javaUpdatedGOBiologicalProcesses = (Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_BiologicalProcess);
+		Set<GKInstance> javaUpdatedGOBiologicalProcesses =
+			(Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_BiologicalProcess);
 		@SuppressWarnings("unchecked")
-		Set<GKInstance> javaUpdatedGOMolecularFunctions = (Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_MolecularFunction);
+		Set<GKInstance> javaUpdatedGOMolecularFunctions =
+			(Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_MolecularFunction);
 		@SuppressWarnings("unchecked")
-		Set<GKInstance> javaUpdatedGOCellularComponents = (Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_CellularComponent);
-		
+		Set<GKInstance> javaUpdatedGOCellularComponents =
+			(Set<GKInstance>) javaUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_CellularComponent);
+
+		System.out.println("All Java GO instances obtained");
+
 		List<GKInstance> listOfAllJavaGOThings = new ArrayList<>();
 		listOfAllJavaGOThings.addAll(javaUpdatedGOCellularComponents);
 		listOfAllJavaGOThings.addAll(javaUpdatedGOMolecularFunctions);
 		listOfAllJavaGOThings.addAll(javaUpdatedGOBiologicalProcesses);
-		
+
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOBiologicalProcesses =
+			(Set<GKInstance>) perlUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_BiologicalProcess);
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOMolecularFunctions = (
+			Set<GKInstance>) perlUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_MolecularFunction);
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOCellularComponents =
+			(Set<GKInstance>) perlUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_CellularComponent);
+
+		System.out.println("All Perl GO instances obtained");
+
+		List<GKInstance> listOfAllPerlGOThings = new ArrayList<>();
+		listOfAllPerlGOThings.addAll(perlUpdatedGOCellularComponents);
+		listOfAllPerlGOThings.addAll(perlUpdatedGOMolecularFunctions);
+		listOfAllPerlGOThings.addAll(perlUpdatedGOBiologicalProcesses);
+
+		Map<String, List<GKInstance>> accessionToPerlInstances = new HashMap<>();
+		for (GKInstance perlGOInstance : listOfAllPerlGOThings) {
+			accessionToPerlInstances.computeIfAbsent(
+				getAccession(perlGOInstance), k -> new ArrayList<>()
+			).add(perlGOInstance);
+		}
+
+
 		Comparator<? super GKInstance> dbIdComparator = new Comparator<GKInstance>()
 		{
 			@Override
@@ -56,15 +95,18 @@ public class JavaVsPerlComparison
 			{
 				return o1.getDBID().compareTo(o2.getDBID());
 			}
-			
+
 		};
-		
+
 		listOfAllJavaGOThings.sort(dbIdComparator);
 		StringBuilder mainSB = new StringBuilder();
 		Map<String, MySQLAdaptor> javaAdaptorPool = new HashMap<>();
 		Map<String, MySQLAdaptor> perlAdaptorPool = new HashMap<>();
-		//for (GKInstance javaGoInst : listOfAllJavaGOThings)
-		listOfAllJavaGOThings.parallelStream().forEach( javaGoInst -> 
+
+		System.out.println("Number of Java GO instances is " + listOfAllJavaGOThings.size());
+
+		AtomicInteger processJavaGOInstances = new AtomicInteger(0);
+		listOfAllJavaGOThings.parallelStream().forEach( javaGoInst ->
 		{
 			try
 			{
@@ -72,7 +114,10 @@ public class JavaVsPerlComparison
 				MySQLAdaptor javaDBAdaptor ;
 				if (!javaAdaptorPool.containsKey(threadID))
 				{
-					javaDBAdaptor = new MySQLAdaptor(javaUpdatedDB.getDBHost(), javaUpdatedDB.getDBName(), javaUpdatedDB.getDBUser(), javaUpdatedDB.getDBPwd(), javaUpdatedDB.getDBPort());
+					javaDBAdaptor = new MySQLAdaptor(
+						javaUpdatedDB.getDBHost(), javaUpdatedDB.getDBName(), javaUpdatedDB.getDBUser(),
+						javaUpdatedDB.getDBPwd(), javaUpdatedDB.getDBPort()
+					);
 					javaAdaptorPool.put(threadID, javaDBAdaptor);
 				}
 				else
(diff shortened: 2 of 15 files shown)


release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -78,18 +93,26 @@ public static int compareInstances(GKInstance inst1, GKInstance inst2, StringBui
...
 	public static int compareInstances(
-		GKInstance inst1, GKInstance inst2, StringBuilder sb, int maxRecursionDepth, boolean checkReferrers
+		GKInstance instance1, GKInstance instance2, StringBuilder stringBuilder,
+		int maxRecursionDepth, boolean checkReferrers
 	)
 	{
-		return compareInstances(inst1, inst2, sb, 0, 0, maxRecursionDepth, null, checkReferrers);
+		final int diffCount = 0;
+		final int initialRecursionDepth = 0;
+		final Predicate<? super SchemaAttribute> customAttributeNameFilter = null;
```
SolomonShorser-OICR: Why is this object being created and immediately assigned null? I don't think it actually gets used. Why not just use `null` instead of `customAttributeNameFilter` whose value is null?

SolomonShorser-OICR: Same goes for `diffCount` and `initialRecursionDepth`, I guess.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -147,225 +176,428 @@ public static int compareInstances(
...
 
-			if (!class1.getName().equals("InstanceEdit") && !class2.getName().equals("InstanceEdit"))
-			{
-				String[] indentArray = new String[recursionDepth * 2];
-				Arrays.fill(indentArray, " ");
-				String indentString = String.join("", indentArray);
+			return 1;
+		}
 
-				if (!class1.getName().equals(class2.getName()))
```
SolomonShorser-OICR: Is the output the same if this condition is removed?

SolomonShorser-OICR:
There are _significant_ changes to DBObjectComparer and EnsemblServiceResponseProcessor. Have they been tested to ensure that the code still functions as expected?

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -22,27 +26,34 @@
...
 	 * If the instances have different schema classes, that will count as 1 diff.
 	 * If a multi-valued attribute has a different number of elements between the two instances,
 	 * that will count as 1 diff and the elements will NOT be compared.
 	 */
-	public static int compareInstances(GKInstance inst1, GKInstance inst2, StringBuilder sb)
+	public static int compareInstances(GKInstance instance1, GKInstance instance2, StringBuilder stringBuilder)
 	{
-		return compareInstances(inst1, inst2, sb, 5, false);
+		final int maxRecursionDepth = 5;
+		final boolean checkReferrers = false;
```
cookersjs: Should this be a global variable and capitalized? I don't personally have an issue with this way of doing it, but should it be that way for consistency across the project?

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -112,18 +135,24 @@ public static int compareInstances(
...
 	 */
 	public static int compareInstances(
-		GKInstance inst1, GKInstance inst2, StringBuilder sb, int maxRecursionDepth,
+		GKInstance instance1, GKInstance instance2, StringBuilder stringBuilder, int maxRecursionDepth,
 		Predicate<? super SchemaAttribute> customAttributeNameFilter, boolean checkReferrers
 	)
 	{
-		return compareInstances(inst1, inst2, sb, 0, 0, maxRecursionDepth, customAttributeNameFilter, checkReferrers);
+		final int diffCount = 0;
+		final int initialRecursionDepth = 0;
```
cookersjs: Similar to above, should this be capitalized and a global variable?

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
SolomonShorser-OICR: Usually, it is only moved to class-level scope and CAPITALIZED if it's going to be accessed in several different places.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -147,225 +176,428 @@ public static int compareInstances(
...
+	 * @param instance1 First instance to compare
+	 * @param instance2 Second instance to compare
+	 * @return <code>true</code> if the instances are of different types; <code>false</code> otherwise
+	 */
+	private static boolean differentInstanceTypes(GKInstance instance1, GKInstance instance2)
+	{
+		String instanceType1 = instance1.getSchemClass().getName();
+		String instanceType2 = instance2.getSchemClass().getName();
+
+		return !instanceType1.equals(instanceType2);
```
cookersjs: For _brevity's_ sake, it could be 
`return !instanceType1.getSchemClass().getName().equals(instance2.getSchemClass().getName())`,
though I can see that might not be the best for _clarity's_ sake

SolomonShorser-OICR: There's always a trade-off, and when that happens, you'll always get dissenting opinions. Fun times... ;)

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -147,225 +176,428 @@ public static int compareInstances(
...
+	 * "modified", and "created" are filtered out (i.e. ignored).
+	 * @return Predicate function used to filter out SchemaAttributes value by their names
+	 */
+	private static Predicate<? super SchemaAttribute> getDefaultAttributeNameFilter()
+	{
+		/*
+		 * Used for filtering out attributes that you don't want to compare because they will
+		 * probably create too much noise if you do.
+		 */
+		return a -> {
```
cookersjs: A more descriptive name for this variable would be nice

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -147,225 +176,428 @@ public static int compareInstances(
...
+	 * Returns <code>true</code> if the attribute passed is an "instance attribute" (i.e. its value(s) is/are
+	 * instance(s)); <code>false otherwise</code>
+	 * @param attribute Attribute to check for the type of value it holds
+	 * @return <code>true</code> if the attribute passed holds instances; <code>false</code> otherwise
+	 */
+	private static boolean isAttributeContainingInstances(SchemaAttribute attribute)
+	{
+		Class<?> attributeType = attribute.getType();
+
+		return attributeType.equals(GKInstance.class) || attributeType.equals(Instance.class);
```
cookersjs: Is there an actual difference between type `GKInstance` and type `Instance`? How are they employed differently

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
```diff
@@ -153,160 +150,249 @@ public EnsemblServiceResponseProcessor()
 	 * @param response HttpResponse from the EnsEMBL service
 	 * @param originalURI The URI of the EnsEMBL service queried (e.g rest.ensembl.org or rest.ensemblgenomes.org)
 	 * @return EnsemblServiceResult object containing the relevant information from the response
+	 * @deprecated use {@link #processResponse(HttpResponse) instead}
 	 */
+	@Deprecated
```
cookersjs: Why would we keep code we won't be using anymore? I can imagine having it for a little while after, but in time wouldn't it make sense to remove it so the class doesn't get bloated?

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
```diff
@@ -153,160 +150,249 @@ public EnsemblServiceResponseProcessor()
...
+	 * The counter for the number of times waited will be incremented each time this method is called when the
+	 * threshold has not been met.
+	 *
+	 * @return <code>true</code> if the number of times query has been told to wait is less than the threshold;
+	 * <code>false</code> otherwise.
+	 */
+	private boolean timesWaitedThresholdNotMet()
+	{
+		// If we get told to wait >= 5 times, let's just take the hint and stop trying.
+		final int MAX_TIMES_TO_WAIT = 5;
```
cookersjs: Similar as early on in my PR -- this one is capitalized but should it be globally available?

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
SolomonShorser-OICR: Usually if you're producing a public API, you mark old code as deprecated for a release or two before removing it. That way, users who see it's deprecated can switch to something else. We don't really have a lot of external users so it's probably safe to remove it immediately, or in the next release. But some `@Deprecated` things can linger for years. I think there are some in the core Java libraries. They aren't used or recommended, but they are left so as not to break anything that relies on them.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
cookersjs: I think for this particular case there was an additional  time where it was accessed, in the function immediately following it

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: I wanted to give those values, even though they are null and 0, names so it is clear what is being passed to the `compareInstances` method on the next line.  It is hard to understand at a glance what the parameter is when the literal value is passed.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: The condition was actually just relocated to line 188: https://github.com/reactome/data-release-pipeline/pull/121/files/600bdc5e31fbe03661507750fb03f8b195feabe1#diff-b4b21932944d31bdd5550b507e7e2195R188

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Yup, you're right.  I will move these to be class-scoped since they are re-used.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Yeah, I considered implementing it in one line, but I thought the extra lines were worth it for making it clear that the types of the instances were being obtained and compared.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Renamed to "attribute"!

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: It looks like `Instance` is the interface type and `GKInstance` is the class that implements the interface.  I'm not sure if there is another class that implements `Instance`.

@SolomonShorser-OICR I'm not too sure how `SomeType.class` works.  Is it possible for a Class object to hold the value of an interface and `attribute.getType()` would be equal to `Instance.class`?

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
jweiser: Yeah, I'm trying to follow semantic versioning for our releases.  If we were to remove a public method instead of deprecating it, it would require moving release-common-lib to v2.0.  Actually, since the public API has new but backwards-compatible changes, I should update the version to 1.2.0 instead of 1.1.1.

I don't mind going to v2.0, but I thought maybe we should give it a bit of time to see if there are any more changes or additions to release-common-lib before incrementing the major version.

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
jweiser: Done!

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Yup, done!

jweiser:
> There are _significant_ changes to DBObjectComparer and EnsemblServiceResponseProcessor. Have they been tested to ensure that the code still functions as expected?

That's a good point.  I'm working on testing now.  The EnsemblServiceResponseProcessor now has unit tests here: https://github.com/reactome/data-release-pipeline/pull/121/files#diff-75a6d81e82555296bca4e36a6cfd4dc3

I'm continuing to work on testing the classes as a whole to see the results before and after refactoring are the same.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -22,27 +26,37 @@
...
 public class DBObjectComparer
 {
+	private static Map<GKInstance, Map<SchemaAttribute, List<Object>>> instanceAttributeToValuesMap = new HashMap<>();
+	private static Map<GKInstance, List<SchemaAttribute>> instanceToRegularAttributesMap = new HashMap<>();
+	private static Map<GKInstance, List<SchemaAttribute>> instanceToReferrerAttributesMap = new HashMap<>();
+
+	private static final int MAX_RECURSION_DEPTH = 5;
+	private static final int INITIAL_INSTANCES_DIFFERENCES_COUNT = 0;
+	private static final int INITIAL_RECURSION_DEPTH = 0;
+
```
SolomonShorser-OICR: I would recommend prefixing these variable names with "DEFAULT_" since they can be overridden if users give their own values.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -147,225 +173,428 @@ public static int compareInstances(
...
+	 * used as input to the function.
+	 * @param instance Instance the function will use to retrieve attribute values
+	 * @return Function which accepts a SchemaAttribute and returns values for the instance passed during its
+	 * construction
+	 */
+	private static Function<SchemaAttribute, List<Object>> constructGetValuesFunctionForRegularAttributes(
+		GKInstance instance
+	)
+	{
+		return (attribute) ->
```
SolomonShorser-OICR: Why is there a lambda here? It looks like this would work fine without it. This seems like an extra layer of complexity for no clear reason, but maybe I'm missing something?

Also, same question as above but for `constructGetValuesFunctionForReferrerAttributes`...

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Good call - done!

SolomonShorser-OICR: On second thought, maybe "INITIAL" is not needed if "DEFAULT" is there? Not sure if that gets redundant or just too long.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Originally, the logic was captured in a function and applied later in the code.  You're right, though, that now I may be able to run the code directly instead of creating and then calling the function.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
SolomonShorser-OICR: But why no "DEFAULT_" for MAX_RECURSION_DEPTH?

jweiser: Yeah, I think you're right.  I will remove "INITIAL" and replace it with "DEFAULT".  I'll also add it for "MAX_RECURSION_DEPTH" - when I made the change I only considered the ones that began with "INITIAL" and overlooked that one.

jweiser: Done!

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: Done!

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
SolomonShorser-OICR: Hmmm... I think 1.2.0 is appropriate since no new major functionality has been introduced, and it _really_ should be compatible with the code that's already using it.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
SolomonShorser-OICR: I can't recall how `.class` works in the context of interfaces. It should be possible to write a very small piece of test code to see if `GKInstance.class == Instance.class` (or is `.equals()` more appropriate here?), or just print GKInstance.class and Instancee.class to the console and see what you get. 

It might also depend on how the underlying Reactome library populates `attribute`...

release-common-lib/src/main/java/org/reactome/util/ensembl/EnsemblServiceResponseProcessor.java
jweiser: Yeah, for this PR I think I'm going to go with v.1.2.0: https://github.com/reactome/data-release-pipeline/pull/121/commits/59e5bb48da93b4caff70260fb86a5f2963cedc4a

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: It looks like they are different returning the following for `GKInstance.class.toString()` and`Instance.class.toString()` respectively:

- class org.gk.model.GKInstance
- interface org.gk.model.Instance

I'm not sure what calls the setter corresponding to `getType()` method for `attribute` to fill its type, but I think the code should okay as is.  Checking both the `GKInstance` class and `Instance` interface may not be necessary but covers more bases in case `Instance.class` is used.

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -192,38 +195,228 @@ private static int compareInstances(
...
+	/**
+	 * Compares the value of the passed attribute between two instances and returns the number of differences.  For an
+	 * value which is a GKInstance, differences are checked for recursively (up to the passed maxRecursionDepth).
+	 * For the base case of "simple" value (i.e. Strings, numbers, etc..., arrays of Strings/numbers/etc...), a count
+	 * of 1 is returned for any difference found between the values compared.
+	 * @param attribute Attribute for which values are being compared
+	 * @param attributeRelationshipType Relationship between the attribute and the instances passed (i.e. 'regular' or
+	 * referrer attribute
+	 * @param value1 First value to compare
+	 * @param value2 Second value to compare
```
SolomonShorser-OICR: I'm a little confused by this: Why are values being passed in? This function already has instance objects and attributes. Why can't values be looked up within the function vs. passed in from outside?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -107,94 +156,127 @@ public int compare(GKInstance o1, GKInstance o2)
...
 								{
 									regulationSameCount.incrementAndGet();
 								}
 							}
 						}
-						else if (javaGoInst.getSchemClass().getName().equals(ReactomeJavaConstants.GO_CellularComponent))
+						else if (
+							javaGoInst.getSchemClass().getName().equals(ReactomeJavaConstants.GO_CellularComponent)
+						)
 						{
```
SolomonShorser-OICR: Formatting change to reduce line-width by 1 char, but increase line-count by 2? ;)

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -82,18 +127,22 @@ public int compare(GKInstance o1, GKInstance o2)
...
 				{
 					perlDBAdaptor = perlAdaptorPool.get(threadID);
 				}
-				
+
 				javaGoInst.setDbAdaptor(javaDBAdaptor);
 				String accession = (String) javaGoInst.getAttributeValue(ReactomeJavaConstants.accession);
 				@SuppressWarnings("unchecked")
-				Collection<GKInstance> perlGoInsts = (Collection<GKInstance>) perlDBAdaptor.fetchInstanceByAttribute(javaGoInst.getSchemClass().getName(), ReactomeJavaConstants.accession, "=", accession);
+				Collection<GKInstance> perlGoInsts = accessionToPerlInstances.get(accession);
```
SolomonShorser-OICR: Not sure I'm reading this correctly, but it seems like an awful lot of new code was added to populate `accessionToPerlInstances`. Does it have a different result than the old code?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -223,4 +328,31 @@ else if (javaGoInst.getSchemClass().getName().equals(ReactomeJavaConstants.GO_Mo
...
+			return (String) instance.getAttributeValue(ReactomeJavaConstants.accession);
+		} catch (Exception e) {
+			throw new RuntimeException("Can't find accession for instance " + instance.getDBID(), e);
+		}
+	}
+
+	private static String getNameValuesAsString(Set<String> values)
+	{
+		return values
+			.stream()
```
SolomonShorser-OICR: Are streams really necessary for this sort of operation?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: It allows the GO instances from the Perl database to be queried at once and stored in a map of String accession value to the corresponding GKInstance.  The old code was making one database call for each individual accession.  When I was refactoring, I noticed a slower running time and I made this change to try to address it.  I think the issue may have been somewhere else in the way I structured something, but I thought it would be good to keep this change to have a single bulk query to the database rather than individual ones.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Fair point - I replaced this with an inline String.join() operation: https://github.com/reactome/data-release-pipeline/pull/121/commits/cf7396e26c9ffe9bf9acc8024dbddc376adc01d9

I tried to do the same for https://github.com/reactome/data-release-pipeline/pull/121/files/9a9fd71d2fd9cc06e60023dbba7fe18787e88547..01f3cf3fc99e30e9a14e697af6e22683f989212c#diff-a1b0b46e16fc2dac4e77db0ea04b1bb7R350 but since it's a set of Map entries, it needs to be mapped to String values before being joined and a stream seems like a more readable way to do that over a for-loop.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Yeah, okay :).  

I'm trying to keep line count to a cutoff of 120 characters so the width of the code is uniform and there's no need to horizontal scroll when working in an IDE or GitHub.  I agree, though, this adds a lot of whitespace.

Since checking the schema class name is logic that is re-used a few times, I just created a helper method to consolidate it and address the line width: https://github.com/reactome/data-release-pipeline/pull/121/commits/e72f83d95c31b984837bb1e3e7912283cbb7db5b

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
SolomonShorser-OICR: Ok, just be aware that streams often incur some overhead (not always a trivial amount), and they are not always more readable.

chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
```diff
@@ -18,58 +18,82 @@
...
+					schemaAttribute.getName().equals(ReactomeJavaConstants.referenceEntity);
+
 				return nameIsInList && !simpleEntityWithReferenceEntityAttribute;
 			};
-			
-			int currentDiff = DBObjectComparer.compareInstances(javaInstance, perlInstance, sb, 1, customAttributeNameFilter, true);
-			
+
+			int currentDiff = DBObjectComparer.compareInstances(
+				javaInstance, perlInstance, sb, 1, customAttributeNameFilter, true
```
cookersjs: What does the `1` mean in `compareInstances`? If it is an internal method, is it worth having it as a named variable?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Understood!  I always try to make my code as easy to follow as possible as the first priority and then optimize any hot spots after that.  I've found the principle of "it's easier to optimize correct code rather than correct optimize code" is usually right.

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -21,58 +26,98 @@
 {
 	public static void main(String[] args) throws Exception
 	{
-		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "gk_central_2019_04_18", "root", "root", 3306);//new MySQLAdaptor("localhost", "gk_central_after_uniprot_update.dump.sql", "root", "root", 3308);
-		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "gk_central_2019_04_18", "root", "root", 3307);//new MySQLAdaptor("localhost", "gk_central_R66_before_chebi_update.sql", "root", "root", 3307);
+		MySQLAdaptor perlUpdatedDB = new MySQLAdaptor("localhost", "perl_gk_central", "root", "root", 3306);
+		MySQLAdaptor javaUpdatedDB = new MySQLAdaptor("localhost", "java_gk_central", "root", "root", 3306);
```
cookersjs: The formatting for created MySQLAdaptor objects differs from how ChebiUpdate does it: https://github.com/reactome/data-release-pipeline/pull/121/files/600bdc5e31fbe03661507750fb03f8b195feabe1..cf7396e26c9ffe9bf9acc8024dbddc376adc01d9#diff-3adf77f599ea429e20f4476c381fdaf0R21-R26

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
SolomonShorser-OICR: I'm not sure that "simply don't use the streams API unless necessary" is an optimization, so...

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -21,58 +26,98 @@
...
 		listOfAllJavaGOThings.addAll(javaUpdatedGOCellularComponents);
 		listOfAllJavaGOThings.addAll(javaUpdatedGOMolecularFunctions);
 		listOfAllJavaGOThings.addAll(javaUpdatedGOBiologicalProcesses);
-		
+
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOBiologicalProcesses =
+			(Set<GKInstance>) perlUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_BiologicalProcess);
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOMolecularFunctions = (
```
cookersjs: _Nitpick alert_: The formatting differs between line 67-74 when initializing the Sets. Lines 67 & 73 do not have the opening parentheses but line 70 does

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -21,58 +26,98 @@
...
+		@SuppressWarnings("unchecked")
+		Set<GKInstance> perlUpdatedGOCellularComponents =
+			(Set<GKInstance>) perlUpdatedDB.fetchInstancesByClass(ReactomeJavaConstants.GO_CellularComponent);
+
+		System.out.println("All Perl GO instances obtained");
+
+		List<GKInstance> listOfAllPerlGOThings = new ArrayList<>();
+		listOfAllPerlGOThings.addAll(perlUpdatedGOCellularComponents);
+		listOfAllPerlGOThings.addAll(perlUpdatedGOMolecularFunctions);
+		listOfAllPerlGOThings.addAll(perlUpdatedGOBiologicalProcesses);
```
cookersjs: Instead of `listOfAllPerlGoThings`, maybe it should become `listOfAllPerlGoInstances`?

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
```diff
@@ -107,94 +156,123 @@ public int compare(GKInstance o1, GKInstance o2)
...
 								perlList = perlList.stream().sorted(dbIdComparator).collect(Collectors.toList());
-	
+
 								GKInstance javaReferrer =  javaList.get(0);
 								GKInstance perlReferrer = perlList.get(0);
-	
-								int numDiffs = DBObjectComparer.compareInstances(javaReferrer, perlReferrer, sb1, 0, isNameAttribute, true);
+
+								int numDiffs = DBObjectComparer.compareInstances(
+									javaReferrer, perlReferrer, sb1, 0, isNameAttribute, true
```
cookersjs: Similar to https://github.com/reactome/data-release-pipeline/pull/121/files/600bdc5e31fbe03661507750fb03f8b195feabe1..cf7396e26c9ffe9bf9acc8024dbddc376adc01d9#r358270477 -- though this time it is a `0` instead of a `1`. Should we have named variables for this?

release-common-lib/src/main/java/org/reactome/util/compare/AttributeProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+	 * Returns the attributes associated, via the passed attribute relationship type, with the passed schema class.
+	 * @param schemaClass SchemaClass for which to retrieve attributes
+	 * @param attributeRelationshipType Type of attributes to retrieve from the passed schema class (i.e. 'regular' or
+	 * referrer attributes)
+	 * @return List of SchemaAttribute objects associated, via the passed attribute relationship type, with the passed
+	 * schema class
+	 */
+	@SuppressWarnings("unchecked")
+	static List<SchemaAttribute> getAttributes(
+		SchemaClass schemaClass, AttributeRelationshipType attributeRelationshipType
```
cookersjs: Is `AttributeRelationshipType` an external class or a Reactome one? I've never seen it before

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
jweiser: The instances were actually only passed in for using their names in an error message.  I refactored to remove passing in the instances unnecessarily and so the method is only for comparing the two values now: https://github.com/reactome/data-release-pipeline/pull/121/commits/cb4d0ba714ffa1142354eb90990b0c3a8da2a3c1

release-common-lib/src/main/java/org/reactome/util/compare/AttributeProcessor.java
SolomonShorser-OICR: I'm pretty sure that's an `enum` Joel created to make it easier to check relationship types. https://github.com/reactome/data-release-pipeline/pull/121/commits/6b9cecd404784e334488c24488846abab86f9e2f#diff-b4b21932944d31bdd5550b507e7e2195R723

chebi-update/src/test/java/org/reactome/release/chebiupdate/JavaVsPerlComparison.java
jweiser: Done: https://github.com/reactome/data-release-pipeline/pull/121/commits/19a51ca5c32d5b1f22091bcfdfdf0bcf58690153

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Changed ChEBIUpdate to match: https://github.com/reactome/data-release-pipeline/pull/121/commits/19a51ca5c32d5b1f22091bcfdfdf0bcf58690153

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Nitpick picked :): https://github.com/reactome/data-release-pipeline/pull/121/commits/ce8ce4883a21048b4e003618de3da7b8ba3ae538#diff-a1b0b46e16fc2dac4e77db0ea04b1bb7R74

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Changed: https://github.com/reactome/data-release-pipeline/pull/121/commits/ce8ce4883a21048b4e003618de3da7b8ba3ae538

go-update/src/test/java/org/reactome/release/goupdate/JavaVsPerlComparison.java
jweiser: Done!  Variables at the top and the changes to the method calls towards the bottom: https://github.com/reactome/data-release-pipeline/pull/121/commits/ce8ce4883a21048b4e003618de3da7b8ba3ae538#diff-a1b0b46e16fc2dac4e77db0ea04b1bb7R74

release-common-lib/src/main/java/org/reactome/util/general/CollectionUtils.java
```diff
@@ -29,13 +30,34 @@
...
 
+	/**
+	 * Returns a list with all the elements of the passed lists in the order the lists are received (i.e. flattens
+	 * multiple lists into one).
+	 * @param <E> Type of the elements for the passed lists and returned list
+	 * @param lists List objects of type E to be combined into a single list
+	 * @return Combined list of type E or an empty List of type E if the lists received are null or empty
+	 */
+	@SafeVarargs
+	public static <E> List<E> combineLists(List<E> ...lists) {
```
SolomonShorser-OICR: I'm a bit surprised to see a function with a for-loop instead of something using the streams API, like `lists.stream().flatMap(l -> l.stream()).collect(Collectors.toList());` ;)

...not that I'm suggesting it needs to be changed (this probably performs better and is easier to understand). ;)

jweiser: Yeah, in this case, a for-loop seemed clearer to me and also since I was filtering out any `null` list values.

_"I don't always use functional programming, but when I do I choose streams..." (The most interesting programmer in the world)_

release-common-lib/src/main/java/org/reactome/util/compare/DBObjectComparer.java
```diff
@@ -287,317 +474,125 @@ private static String getInstanceTypeMismatchMessage(GKInstance instance1, GKIns
...
-		 * Used for filtering out attributes that you don't want to compare because they will
-		 * probably create too much noise if you do.
-		 */
-		return a -> {
-			return !a.getName().equals(ReactomeJavaConstants.DB_ID)
-				&& !a.getName().equals(ReactomeJavaConstants.dateTime)
-				&& !a.getName().equals(ReactomeJavaConstants.modified)
-				&& !a.getName().equals(ReactomeJavaConstants.created);
-		};
+		return ((MySQLAdaptor) instance.getDbAdaptor()).getDBName();
```
cookersjs: Funny that you need to cast `MySQLAdaptor` for the method `getDbAdaptor()`

jweiser: Yeah, `getDbAdaptor()` returns the interface type `PersistenceAdaptor` which is implemented by `MySQLAdaptor` and a few other things, but when working in this context the actual object is always a `MySQLAdaptor`.

PR #123, opened 2019-12-17
Create LICENSE

jweiser:
Since we are publishing release-common-lib to Maven Central, I realized we should include the license referenced in its POM file in the repository.

The only change made to the license template is here: https://github.com/reactome/data-release-pipeline/compare/develop...add-license-1?quick_pull=1#diff-9879d6db96fd29134fc802214163b95aR632

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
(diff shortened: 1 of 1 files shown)


SolomonShorser-OICR:
I'm not sure it's my place to approve license changes. Was Robin OK with this license choice?

jweiser:
> I'm not sure it's my place to approve license changes. Was Robin OK with this license choice?

Yes, I think so.  It was the license we chose when discussing the creation of the reactome-base artifact.  It's also what is referenced in the release-common-lib's POM file.

SolomonShorser-OICR:
If this is the license that's supposed to be used, then I'll approve.

adamjohnwright:
@hawrobin mentioned in slack that for software the Apache2 license has been agreed upon. @jweiser this does not appear to be an Apache 2 license. Is there a reason for the choice of license?

